PR #39, opened 2018-04-01
Update cats to 1.1.0 (and dependencies)

trane:
Because of the update to cats 1.1.0, all of the dependencies needed to
be updated, too. There were also breaking api changes to both cats,
http4s, and fs2:

cats:
  1. Inject -> InjectK
  1. Coproduct -> EitherK
  1. EitherT.left now infers from F[A]
  1. No more need for sequenceU*, traverseU* due to Unapply updates

fs2:
  1. fs2.Strategy -> ExecutionContext
  1. fs2.Task -> cats.effect.IO

http4s:
  1. Most types now take a `F[_] : Effect`
  1. Request -> Request[F[_]]
  1. Entity*coder[A] -> Entity*coder[F[_], A]
  1. jsonEncoderOf[A] -> jsonEncoderOf[F[_], A]

I tested code generation with petstore.yaml and the resulting project
succeeded.

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/swagger/codegen/ClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/swagger/codegen/ClientGenerator.scala
@@ -23,7 +23,7 @@ object ClientGenerator {
       clientExtraImports <- getExtraImports(context.tracing)
       clients <- groupedRoutes.map({ case (pkg, routes) =>
         for {
-          clientCalls <- routes.map(generateClientOperation(pkg, context.tracing, protocolElems) _).sequenceU
+          clientCalls <- routes.map(generateClientOperation(pkg, context.tracing, protocolElems) _).sequence
           clientName = s"${pkg.lastOption.getOrElse("").capitalize}Client"
           tracingName = Option(pkg.mkString("-")).filterNot(_.isEmpty)
           ctorArgs <- clientClsArgs(tracingName, schemes, host, context.tracing)
@@ -38,7 +38,7 @@ object ClientGenerator {
 
           Client(pkg, clientName, stats.map(SwaggerUtil.escapeTree))
         }
-      }).sequenceU
+      }).sequence
     } yield Clients(clients)
   }
 }
--- a/modules/codegen/src/main/scala/com/twilio/swagger/codegen/ClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/swagger/codegen/ClientGenerator.scala
@@ -21,9 +21,9 @@ object ClientGenerator {
     for {
       clientImports <- getImports(context.tracing)
       clientExtraImports <- getExtraImports(context.tracing)
-      clients <- groupedRoutes.map({ case (pkg, routes) =>
+      clients <- groupedRoutes.traverse({ case (pkg, routes) =>
         for {
-          clientCalls <- routes.map(generateClientOperation(pkg, context.tracing, protocolElems) _).sequence
+          clientCalls <- routes.traverse(generateClientOperation(pkg, context.tracing, protocolElems) _)
           clientName = s"${pkg.lastOption.getOrElse("").capitalize}Client"
           tracingName = Option(pkg.mkString("-")).filterNot(_.isEmpty)
           ctorArgs <- clientClsArgs(tracingName, schemes, host, context.tracing)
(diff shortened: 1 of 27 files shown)


modules/codegen/src/main/scala/com/twilio/swagger/codegen/ClientGenerator.scala
```diff
@@ -23,7 +23,7 @@ object ClientGenerator {
       clientExtraImports <- getExtraImports(context.tracing)
       clients <- groupedRoutes.map({ case (pkg, routes) =>
         for {
-          clientCalls <- routes.map(generateClientOperation(pkg, context.tracing, protocolElems) _).sequenceU
+          clientCalls <- routes.map(generateClientOperation(pkg, context.tracing, protocolElems) _).sequence
```
tpolecat: Just FYI, `.map(f).sequence` can be written as `.traverse(f)`. It looks like you have several such cases here.

modules/codegen/src/main/scala/com/twilio/swagger/codegen/generators/AkkaHttpServerGenerator.scala
```diff
@@ -112,7 +113,7 @@ object AkkaHttpServerGenerator {
                 ServerRoute(pathStr, httpMethod, operation)
               }
             }
-          }).sequenceU.map(_.flatten)
+          }).sequence.map(_.flatten)
```
tpolecat: I believe this can be `.flatSequence`.

trane: I'm sure this is a resolvable error, but I'm getting:

```
Cannot prove that cats.data.EitherT[com.twilio.swagger.codegen.Logger,String,scala.collection.mutable.Iterable[com.twilio.swagger.codegen.ServerRoute]] <:< G[List[A]].
```
it looks like the `mutable.Iterable` is the problem and if I stick a `.toList` on line 115, then `flatSequence` works. Any thoughts on a better approach?

tpolecat: Yeah I didn't look at the types carefully. I was assuming that was a monadic join. If it's the weird stdlib `.flatten` then what you have now is the best you can do.

blast-hardcheese: I'm thinking
```
diff --git a/modules/codegen/src/main/scala/com/twilio/swagger/codegen/generators/AkkaHttpServerGenerator.scala b/modules/codegen/src/main/scala/com/twilio/swagger/codegen/generators/AkkaHttpServerGenerator.scala
index e053afd..a2c7e72 100644
--- a/modules/codegen/src/main/scala/com/twilio/swagger/codegen/generators/AkkaHttpServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/swagger/codegen/generators/AkkaHttpServerGenerator.scala
@@ -104,17 +104,17 @@ object AkkaHttpServerGenerator {
       case ExtractOperations(paths) =>
         for {
           _ <- Target.log.debug("AkkaHttpServerGenerator", "server")(s"extractOperations(${paths})")
-          routes <- paths.map({ case (pathStr, path) =>
+          routes <- paths.traverse({ case (pathStr, path) =>
             for {
               _ <- Target.log.info("AkkaHttpServerGenerator", "server", "extractOperations")(s"(${pathStr}, ${path})")
               operationMap <- Target.fromOption(Option(path.getOperationMap), "No operations defined")
             } yield {
-              operationMap.asScala.map { case (httpMethod, operation) =>
+              operationMap.asScala.toList.map { case (httpMethod, operation) =>
                 ServerRoute(pathStr, httpMethod, operation)
               }
             }
-          }).sequence.map(_.flatten)
-        } yield routes
+          })
+        } yield routes.flatten
 
       case GetClassName(operation) =>
         for {
```
might be clearer, as it surfaces where we actually flatten the paths we get (from `paths`) and the operations (`operationMap.toScala.toList`) into a single `List[ServerRoute]`, should we ever discover we need to do more work before this information is lost.

blast-hardcheese:
This is great, thanks!

Issue #40, opened 2018-04-02
Not all definitions end up in Swagger.definitions


andrew-domino:
While parsing out swagger where definitions are only referenced in `additionalProperties` the model is never created. This is due to https://github.com/swagger-api/swagger-parser/issues/342 and is fixed in swagger-parser 1.0.34.

Issue #36, opened 2018-03-26
Principled Java code generation


blast-hardcheese:
Jotting down some thoughts for the future. The core underlying concept of guardrail seems doable in Java; one challenge would be maintaining the flexibility of the various typeclasses without global registries. Explicit value passing and typechecking responses.

The biggest advantage I see for providing this would be to permit an easier migration path from existing Java services into Scala, by first following best practices isolating the business logic from the web framework and using generated models+encoders/decoders for all communication, then migrating the business logic piecemeal into Scala by using `scala-java8-compat` to adapt down to `toCompletableFuture`.

blast-hardcheese:
Some code to throw darts at:
```
interface GetUserResponse {
    int statusCode;
}

class GetUserResponseBuilder() {
    class GetUserResponseOk() extends GetUserResponse {
        int statusCode = 200;
    }

    class GetUserResponseBadRequest(String message) extends GetUserResponse {
        int statusCode = 400;
        String getMessage() {
            return message;
        }
    }

    GetUserResponse Ok() {
        return new GetUserResponseOk();
    }

    GetUserResponse BadRequest(String message) {
        return new GetUserResponseBadRequest(message);
    }
}

interface FooUser {
    CompletableFuture<GetUserResponse> getUser(GetUserResponseBuilder respond, String username);
}

new FooUser {
    CompletableFuture<GetUserResponse> getUser(GetUserResponseBuilder respond, String username) {
        CompletableFuture<GetUserResponse> result = new CompletableFuture<>();
        if (username == "foo") {
            result.complete(respond.Ok());
        } else {
            result.complete(respond.BadRequest("Unknown user: " + username));
        }
        return result;
    }
}
```

ktoso:
Yeap, this would be pretty great to grow adoption and to get people to be able to move/switch if they wish to do so...

I think it'd be amazing to have both languages covered, and all these tools expose Routes, since then one can literally combine "anything they want" into such routes...

For reference, this is the style that akka-grpc generates for Java: https://github.com/akka/akka-grpc/blob/master/plugin-tester-java/src/main/java/io/grpc/examples/helloworld/GreeterServiceImpl.java 
Similar enough btw... :-)

ktoso:
We discussed some more -- it would be preferable to avoid "string concating style" but rather work on ASTs... If you have a preference which APIs to use posting here would help @blast-hardcheese, thanks!

blast-hardcheese:
[...]

PR #41, opened 2018-04-02
Update swagger-parser to fix definition parse bug

trane:
While parsing swagger with additionalProperties, not all definitions are
resolved in swagger-parser < 1.0.34.
(https://github.com/swagger-api/swagger-parser/pull/602)
This brings us to the most recent version of swagger-parser on the 1.x
line.

Fixes: #40

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/swagger/codegen/SwaggerUtil.scala
+++ b/modules/codegen/src/main/scala/com/twilio/swagger/codegen/SwaggerUtil.scala
@@ -295,6 +295,8 @@ object SwaggerUtil {
         Target.pure(Resolved(t"Double", None, Default(d).extract[Double].map(Lit.Double(_))))
       case d: DecimalProperty =>
         Target.pure(Resolved(t"BigDecimal", None, None))
+      case u: UntypedProperty =>
+        Target.error("FIXME: Got an untyped property, currently not supported")
       case p: AbstractProperty if p.getType.toLowerCase == "integer" =>
         Target.pure(Resolved(t"BigInt", None, None))
       case p: AbstractProperty if p.getType.toLowerCase == "number" =>
--- a/modules/codegen/src/main/scala/com/twilio/swagger/codegen/SwaggerUtil.scala
+++ b/modules/codegen/src/main/scala/com/twilio/swagger/codegen/SwaggerUtil.scala
@@ -296,7 +296,7 @@ object SwaggerUtil {
       case d: DecimalProperty =>
         Target.pure(Resolved(t"BigDecimal", None, None))
       case u: UntypedProperty =>
-        Target.error("FIXME: Got an untyped property, currently not supported")
+        Target.pure(Resolved(t"io.circe.Json", None, None))
       case p: AbstractProperty if p.getType.toLowerCase == "integer" =>
         Target.pure(Resolved(t"BigInt", None, None))
       case p: AbstractProperty if p.getType.toLowerCase == "number" =>
--- a/build.sbt
+++ b/build.sbt
@@ -61,7 +61,7 @@ val codegenSettings = Seq(
   ),
   libraryDependencies ++= testDependencies ++ Seq(
     "org.scalameta" %% "scalameta" % "2.0.1"
-    , "io.swagger" % "swagger-parser" % "1.0.32"
+    , "io.swagger" % "swagger-parser" % "1.0.34"
     , "org.tpolecat" %% "atto-core"  % "0.6.1"
     , "org.typelevel" %% "cats-core" % catsVersion
     , "org.typelevel" %% "cats-kernel" % catsVersion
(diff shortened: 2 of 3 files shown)


trane:
Interesting new NPE introduced from this change. Investigating.

trane:
Looks like we aren't handling `UntypedProperty` within `SwaggerUtil.scala#propMeta`, which is being triggered by 
```yaml
property:
  default: "what"
```. 

According to the Swagger spec (2.0) definitions (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#definitionsObject) can have `null` type via the JSON schema spec: https://tools.ietf.org/html/draft-zyp-json-schema-04#section-3.5

What is the most reasonable thing to do here?

I can think of a few things:

1. Fail when `type` isn't present.
1. Set the type to some `null` or nullable-like value (not sure how we represent this in scala.meta
1. Ignore it
1. Recursively inspect the `default` value (which is itself required to be a JSON schema object) and use the resulting resolved type to be the type we use (though we still run into this nullable situation)

**Comment by trane** [comment]
For now, I'm failing property definitions that are untyped.

**Comment by blast-hardcheese** [comment]
@trane Thanks for the contribution and investigation! Everything looks good, save for the merge conflict -- would you please resolve this at your convenience?

**Comment by blast-hardcheese** [comment]
@trane One last time, my apologies

**Comment by trane** [comment]
@blast-hardcheese, I'm concerned about some of the behavior I'm seeing now with this code. Namely, I'm getting unhelpful errors now like:

```
Error: Unable to resolve type for null (null null null null)
```

Where before I would get
```
Error: No responses defined for getActivities
```

**Comment by blast-hardcheese** [comment]
@trane So, addressing https://github.com/twilio/guardrail/pull/41#issuecomment-377987766:

As for HTTP, a "`null`" value doesn't really make sense. `?Foo=` is still `foo=""`. JSON, however, could represent literal `null` values. For this reason, defaulting the type-less properties to `None.type` might make sense, but then what does it mean to have a `None.type` be `required`. This would be a great opportunity for static parameters, where an undefined parameter always gets encoded as a `None` and can't be overwritten.

**Comment by trane** [comment]
> This would be a great opportunity for static parameters, where an undefined parameter always gets encoded as a None and can't be overwritten.

@blast-hardcheese I'm not sure I follow what you mean by "static parameters". Are you saying that we generate a class constructor which has the `None.type` singleton type instead of an `Option[???]`?

```scala
case class MyParam(foo: String, bar: None.type)
```

trane:
I did a bit more digging into what swagger-codegen did with the addition of `UntypedProperty`
- https://github.com/swagger-api/swagger-core/issues/2507

For all the Scala code generators, they either returned `Any` (Akka, Scala, etc) or `Object` (finch). Though, how could we create an encoder/decoder for such an arbitrary type in any meaningful way?
It appears in the case of `additionalProperties` and `schema` etc, that these all just have to be some valid json, nothing more.

Perhaps the right answer here is to resolve as a `io.circe.Json`?

blast-hardcheese:
Exposing it as the underlying json library type makes sense. This would also be a stopgap for #43, although not a great one.

trane:
Alright, I have resolved the issue with the introduction of `UntypedProperty` from the new swagger-parser in the best way I can think of for now. Should be good to merge if you feel good about it, too, @blast-hardcheese

modules/codegen/src/main/scala/com/twilio/swagger/codegen/SwaggerUtil.scala
```diff
@@ -295,6 +295,8 @@ object SwaggerUtil {
         Target.pure(Resolved(t"Double", None, Default(d).extract[Double].map(Lit.Double(_))))
       case d: DecimalProperty =>
         Target.pure(Resolved(t"BigDecimal", None, None))
+      case u: UntypedProperty =>
+        Target.pure(Resolved(t"io.circe.Json", None, None)) // TODO: o.getProperties
```
blast-hardcheese: Does `getProperties` exist on `UntypedProperty`? [UntypedProperty.java](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/models/properties/UntypedProperty.java) suggests it doesn't (which would also make sense, as there's no way for it to know)

blast-hardcheese: Additionally, something to note here is that SwaggerUtil needs to eventually be collapsed down into the different ProtocolGenerators, otherwise `io.circe.Json` won't be valid for other JSON libraries (`json4s`, support for Java, etc). Not a concern right now, but this is something that'll need to happen for #36

trane: yeah, I figured that the `io.circe.Json` was good for now given that the protocol generator stuff isn't parameterized yet. I hope it's as simple as a find/replace of `io.circe.Json` to `${correctProtocolGenerator}` when that happens.

blast-hardcheese:
This looks good, thanks!

PR #60, opened 2018-05-30
Refactoring

stanislav-chetvertkov:
simplify string interpolation and redundant code constructs, align package names with their physical location, explicit types for implicit definitions


modules/codegen/src/main/scala/com/twilio/guardrail/extract/Extractable.scala
```diff
@@ -2,25 +2,20 @@ package com.twilio.guardrail.extract
...
   def extract(v: Any): Try[T]
 }
 
 object Extractable {
-  def build[T](f: PartialFunction[Any, T]): Extractable[T] = {
-    new Extractable[T] {
-      def extract(v: Any): Try[T] = Try(f(v))
-    }
-  }
+  def build[T](f: PartialFunction[Any, T]): Extractable[T] = (v: Any) => Try(f(v))
```
blast-hardcheese: I honestly have no idea how this works, how is `(v: Any) => Try(f(v))` an instance of `Extractable`?

src/test/scala/package.scala
```diff
@@ -3,9 +3,11 @@ package com.twilio
 import _root_.io.swagger.models._
 import _root_.io.swagger.parser.SwaggerParser
 import cats.arrow.FunctionK
+import cats.free.Free
```
blast-hardcheese: Is this necessary?

modules/codegen/src/main/scala/com/twilio/guardrail/extract/Extractable.scala
stanislav-chetvertkov: I think it uses SAM inference, something similar was available for a while since Java 8, intellij actually now suggests it (but you are not using it as far as I know)

blast-hardcheese: Yeah, looks like it! I saw all the SAM stuff when it first came out, but I didn't look into it very deeply; it seems this is exactly how that works. Very cool!

blast-hardcheese:
One more thought, I'm assuming these are primarily IntelliJ suggestions? If so, let's integrate https://github.com/lucidsoftware/neo-sbt-scalafmt to prevent future drift. I often just pick the formatting that makes the most sense to me at the time, but integrating a proper formatter with consistent rules would probably remove a lot of back-and-forth.

stanislav-chetvertkov:
Thats a great idea, in our team we are using scalariform but I heard that scalafmt is better. I will add it in the scope of this PR

blast-hardcheese:
Thanks!

Issue #65, opened 2018-06-08
Unable to process Multipart Form file uploads in akka-http


blast-hardcheese:
Current generated code looks like:
```
[error] .../target/scala-2.12/src_managed/main/.../Routes.scala:45:296: type mismatch;
[error]  found   : akka.http.scaladsl.common.NameReceptacle[akka.http.scaladsl.model.BodyPartEntity]
[error]  required: akka.http.scaladsl.server.directives.FormFieldDirectives.FieldMagnet
[error]     } ~ (put & path(...) & discardEntity & (formField(Symbol("File").as[BodyPartEntity]))) {
```

`.as[BodyPartEntity]` uses the `SFU` (`FromEntityUnmarshaller[StrictForm]`), any large file uploads would need to be made strict, consuming huge amounts of memory until processed.

Proposal: Switch to fully async stream processing, similar to https://github.com/akka/akka-http/blob/6ba9c4df3d8fea95d655fe40523a87f7d4db35a6/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala#L22 , as well as exposing `(FileInfo => File) => Unmarshal[BodyPartEntity, File]` to let the user choose where to save the files. An adapter from `BodyPartEntity`  to `String` can be derived, allowing any `FromStringUnmarshaller`'d types to be unpacked automatically.

blast-hardcheese:
Resolved by #66

PR #66, opened 2018-06-26
Async multipart form file upload

blast-hardcheese:
- Adding `GeneratorSettings` to provide customizable parameters
- Adding an non-strict path for `Multipart.FormData.BodyPart`-processing, with explicit handling for `File` uploads (#65)
- Fixing a bug where entity is discarded before any `formData` fields are processed

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
@@ -46,9 +46,10 @@ object AkkaHttpClientGenerator {
 
     def apply[T](term: ClientTerm[T]): Target[T] = term match {
       case GenerateClientOperation(className, RouteMeta(pathStr, httpMethod, operation), tracing, protocolElems) =>
-        def generateUrlWithParams(path: String,
-                                  pathArgs: List[ScalaParameter],
-                                  qsArgs: List[ScalaParameter]): Target[Term] = {
+        def generateUrlWithParams(
+            path: String,
+            pathArgs: List[ScalaParameter],
+            qsArgs: List[ScalaParameter]): Target[Term] = {
           for {
             _ <- Target.log.debug("generateClientOperation", "generateUrlWithParams")(
               s"Using $path and ${pathArgs.map(_.argName)}")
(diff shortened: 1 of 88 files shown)


blast-hardcheese:
I'd like to add a full file upload test to `RoundTrip` before releasing, to ensure there's nothing missing, even though it compiles.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
```diff
@@ -45,13 +44,11 @@ object AkkaHttpClientGenerator {
         .fold(param"host: String")(v => param"host: String = ${Lit.String(v)}")
 
     def apply[T](term: ClientTerm[T]): Target[T] = term match {
-      case GenerateClientOperation(className, RouteMeta(pathStr, httpMethod, operation), tracing, protocolElems) =>
-        def generateUrlWithParams(path: String,
-                                  pathArgs: List[ScalaParameter],
-                                  qsArgs: List[ScalaParameter]): Target[Term] = {
+      case GenerateClientOperation(className, RouteMeta(pathStr, httpMethod, operation), tracing, protocolElems, generatorSettings) =>
+        implicit val gs = generatorSettings
```
stanislav-chetvertkov: implicits need to have types defined

blast-hardcheese: I was trying to restructure things so `Target` would be a `ReaderT[F, GeneratorSettings, T]`, but I can't clear up the circular dependency; I'd much rather not have all these parameters everywhere

modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
```diff
@@ -443,44 +473,268 @@ object AkkaHttpServerGenerator {
...
-    def formToAkka: List[ScalaParameter] => Target[Option[Term]] = {
+    def formToAkka: List[ScalaParameter] => Target[Option[Term]] =
       directivesFromParams(
         arg => tpe => Target.pure(q"formField(Symbol(${arg}).as[${tpe}])"),
         arg => tpe => Target.pure(q"formField(Symbol(${arg}).as[${tpe}].*)"),
         arg => tpe => Target.pure(q"formField(Symbol(${arg}).as[${tpe}].*).map(Option.apply _)"),
         arg => tpe => Target.pure(q"formField(Symbol(${arg}).as[${tpe}].?)")
       ) _
+
+    def asyncFormToAkka(operationId: String): List[ScalaParameter] => Target[(Option[Term], List[Stat])] = { params =>
```
stanislav-chetvertkov: OMG 🤕

blast-hardcheese: this follows reasonably closely the strategy in akka, though widening to support more than one uploaded field: https://github.com/akka/akka-http/blob/master/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala#L82

blast-hardcheese: I expect that consumers will want some sort of override ability to stream to S3 or something like that, which is why there are so many override points

modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
blast-hardcheese: I was able to come up with a decent middleground using `ReaderT`. If you wouldn't mind taking another look, I think this approach reads much cleaner.

PR #75, opened 2018-07-13
Added implicit show for Java URL type

lyrahall14:
This compiles, but I'm pretty unfamiliar with how Guardrail works, so I'm not sure if there are any other changes I need to make along with it. Just submitting this so we can iterate.


modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaGenerator.scala
```diff
@@ -5,6 +5,7 @@ import cats.syntax.either._
 import cats.~>
 import com.twilio.guardrail.terms._
 import scala.meta._
+import java.net.URL
```
blast-hardcheese: This import isn't necessary, for two reasons:
1) you've specified the full type in `build[java.net.URL]`
2) the line you've added below is inside `source"""`, which is a scala.meta string interpolator; we use the parsed structure that comes out of that string interpolator to write files that will then be compiled later (code generation, to be compiled by future invocations of `scalac`)

PR #80, opened 2018-07-17
bump swagger parser library version

stanislav-chetvertkov:


stanislav-chetvertkov:
I guess it also needs a couple of examples with swagger 3 specs...

blast-hardcheese:
Unfortunately this seems as though it'll be quite a bit more involved than just bumping a version number -- initially I assumed that would be the case as well, but it seems as though the underlying model representing the swagger format has changed.

```diff
diff --git a/modules/codegen/src/main/scala/com/twilio/guardrail/ReadSwagger.scala b/modules/codegen/src/main/scala/com/twilio/guardrail/ReadSwagger.scala
index 30dbcd9..f552941 100644
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/ReadSwagger.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/ReadSwagger.scala
@@ -2,17 +2,19 @@ package com.twilio.guardrail
 
-import _root_.io.swagger.models.Swagger
-import _root_.io.swagger.parser._
+import io.swagger.v3.parser.OpenAPIV3Parser
+import io.swagger.v3.oas.models.OpenAPI
 import cats._
 import java.nio.file.Path
 import scala.io.AnsiColor
 import scala.util.Try
 
-case class ReadSwagger[T](path: Path, next: Swagger => T)
+case class ReadSwagger[T](path: Path, next: OpenAPI => T)
 object ReadSwagger {
   def unsafeReadSwagger[T: Monoid](rs: ReadSwagger[T]): T =
     (for {
       absolutePath <- Try(rs.path.toAbsolutePath.toString).toOption
-      swagger      <- Option(new SwaggerParser().read(absolutePath))
+      swagger      <- Option(new OpenAPIV3Parser().read(absolutePath))
     } yield rs.next(swagger))
       .getOrElse {
         println(s"${AnsiColor.RED}Spec file ${rs.path} is either incorrectly formatted or missing.${AnsiColor.RESET}")
```

this starts us down the rabbit hole. It seems as though there is [a Swagger/OpenAPI2.0 -> OpenAPI3.0 translation layer](https://github.com/swagger-api/swagger-parser/blob/v2.0.1/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java) included with the parser, so we'll still be able to support OpenAPI2.0 specifications, but I haven't given it more than a cursory glance.

Tangentially related, I'd previously done some work towards investigating whether it would be possible to derive a parser from json-schema itself; there have been some challenges with the hand-written parser, primarily around error reporting, which I believe could be solved by simply deriving a parser that was consistently expressive at all levels (I'm thinking errors reported as JSON Paths to exactly where the problem is, instead of [`Spec file ${rs.path} is either incorrectly formatted or missing.`](https://github.com/twilio/guardrail/blob/8cdf9092d28ac6af371f267c590680aceabaa530/modules/codegen/src/main/scala/com/twilio/guardrail/ReadSwagger.scala#L18).

It seems as though the OpenAPI 3.0 parser has resolved [the most frustrating issue](https://github.com/swagger-api/swagger-parser/blob/v1.0.36/modules/swagger-parser/src/main/java/io/swagger/parser/SwaggerParser.java#L135-L137), which suggests the underlying implementation has gotten similar treatment, but I still maintain that hand-written parsers will eventually be replaced with something derived from a specification.

PR #82, opened 2018-07-27
Polymorphism

stanislav-chetvertkov:
trying to add support for #43

PR #83, opened 2018-07-27
Scalameta bugfix

blast-hardcheese:
https://github.com/scalameta/scalameta/issues/1661 resolved the encoding bug that required manual backtick-escaping


blast-hardcheese:
@stanislav-chetvertkov This will conflict with #82, as it removes the `Escaping` object entirely

PR #98, opened 2018-09-22
Deep hierarchy

dpblh:
The current implementation mistakenly believes that inheritance can only be a one-level.
It's look like wrong decision. A class can inherit from another class while the parent class can also be inherited from some class.

for instance:
`A -> B -> C`

It is more correct to save the hierarchy of inheritance not in the parent but in the child.

In this implementation was made couple common changes
first:
`case class ADT(name: String, tpe: Type.Name, trt: Defn.Trait, companion: Defn.Object) extends StrictProtocolElems` - field child was removed
second:
`case class ClassDefinition(name: String, tpe: Type.Name, cls: Defn.Class, companion: Defn.Object, parents: List[SupperClass] = Nil)
    extends StrictProtocolElems` - field parents was added

This solves the problem deep-inheritance.

I hope. It'll help you.

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
@@ -167,7 +167,7 @@ object CirceProtocolGenerator {
         }
 
       case RenderDTOClass(clsName, selfTerms, parents) =>
-        val discriminator = parents.collectFirst { case SuperClass(_, _, _, Some(discriminator)) => discriminator }
+        val discriminator = parents.find(_.discriminator.isDefined).flatMap(_.discriminator) //collectFirst { case SuperClass(_, _, _, Some(discriminator)) => discriminator }
         val parentNameOpt = parents.headOption.map(_.clsName)
         val terms = (parents.flatMap(_.params.map(_.term)) ++ selfTerms).filterNot(
           param => discriminator.contains(param.name.value)
@@ -181,7 +181,7 @@ object CirceProtocolGenerator {
         Target.pure(code)
 
       case EncodeModel(clsName, needCamelSnakeConversion, selfParams, parents) =>
-        val discriminator = parents.collectFirst { case SuperClass(_, _, _, Some(discriminator)) => discriminator }
+        val discriminator = parents.find(_.discriminator.isDefined).flatMap(_.discriminator)
         val params = (parents.flatMap(_.params) ++ selfParams).filterNot(
           param => discriminator.contains(param.name)
         )
@@ -232,7 +232,7 @@ object CirceProtocolGenerator {
         """)
 
       case DecodeModel(clsName, needCamelSnakeConversion, selfParams, parents) =>
-        val discriminator = parents.collectFirst { case SuperClass(_, _, _, Some(discriminator)) => discriminator }
+        val discriminator = parents.find(_.discriminator.isDefined).flatMap(_.discriminator)
         val params = (parents.flatMap(_.params) ++ selfParams).filterNot(
           param => discriminator.contains(param.name)
         )
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
@@ -366,7 +366,7 @@ object CirceProtocolGenerator {
           }
 
         Target.pure(
-          parentNameOpt.fold(q"""sealed trait ${Type.Name(className)} {..${testTerms}}""")(
+          parentNameOpt.fold(q"""trait ${Type.Name(className)} {..${testTerms}}""")(
             parentName =>
               q"""trait ${Type.Name(className)} extends ${template"${init"${Type.Name(parentName)}(...$Nil)"}{..${testTerms}}"} """
           )
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
@@ -368,7 +368,7 @@ object CirceProtocolGenerator {
         Target.pure(
           parentNameOpt.fold(q"""sealed trait ${Type.Name(className)} {..${testTerms}}""")(
             parentName =>
-              q"""sealed trait ${Type.Name(className)} extends ${template"${init"${Type.Name(parentName)}(...$Nil)"}{..${testTerms}}"} """
+              q"""trait ${Type.Name(className)} extends ${template"${init"${Type.Name(parentName)}(...$Nil)"}{..${testTerms}}"} """
           )
         )
 
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
@@ -323,12 +323,9 @@ object CirceProtocolGenerator {
 
   object PolyProtocolTermInterp extends (PolyProtocolTerm ~> Target) {
     override def apply[A](fa: PolyProtocolTerm[A]): Target[A] = fa match {
-      case RenderADTCompanion(clsName, needCamelSnakeConversion, discriminator, encoder, decoder) =>
-        val normalizedDiscriminator =
-          if (needCamelSnakeConversion) q"io.circe.derivation.renaming.snakeCase(${Lit.String(discriminator)})"
-          else Lit.String(discriminator)
+      case RenderADTCompanion(clsName, discriminator, encoder, decoder) =>
         val code = q"""object ${Term.Name(clsName)} {
-             val discriminator:String = ${normalizedDiscriminator}
+             val discriminator:String = ${Lit.String(discriminator)}
             ..${List(encoder, decoder)}
           }
           """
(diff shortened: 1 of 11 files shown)


stanislav-chetvertkov:
I think that the approach in #82 is somewhat limited and prefer this one, don't know if it makes sense to merge it into #82 because had not been any changes. Probably makes sense to finish it in here and merge to master.

dpblh:
@blast-hardcheese Check it please

blast-hardcheese:
This looks good to me, other than it doesn't currently compile.

What still needs to be done:
- [x] rebase onto master
- [ ] Make #98 compile
- [ ] write end to end tests to verify the protocol
- [ ] resolve review comments from #82

blast-hardcheese:
@dpblh I think your idea of implementing encoders explicitly makes sense, additionally it follows how all other encoders are written. I support that direction.

blast-hardcheese:
@dpblh a64a04d is absolutely baffling. Phantom types?

blast-hardcheese:
Anyway, it looks like we just need `import cats.syntax.either._` to backport `flatMap` to 2.11.x's `Either`, as well as resolving:
```
[warn] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolElems.scala:41: Exhaustivity analysis reached max recursion depth, not all missing cases are reported.
[warn] (Please try with scalac -Ypatmat-exhaust-depth 40 or -Ypatmat-exhaust-depth off.)
[warn]                 .map {
[warn]                      ^
[warn] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolElems.scala:51: match may not be exhaustive.
[warn] It would fail on the following inputs: ADT(_, _, _, _), ClassDefinition((x: String forSome x not in name), _, _, _, _), EnumDefinition((x: String forSome x not in name), _, List(_), _, _), EnumDefinition((x: String forSome x not in name), _, Nil, _, _), EnumDefinition(name, _, List(_), _, _), EnumDefinition(name, _, Nil, _, _), RandomType((x: String forSome x not in name), _)
[warn]                 .map {
[warn]                      ^
[warn] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolElems.scala:61: match may not be exhaustive.
[warn] It would fail on the following inputs: ADT(_, _, _, _), ClassDefinition((x: String forSome x not in name), _, _, _, _), EnumDefinition((x: String forSome x not in name), _, List(_), _, _), EnumDefinition((x: String forSome x not in name), _, Nil, _, _), EnumDefinition(name, _, List(_), _, _), EnumDefinition(name, _, Nil, _, _), RandomType((x: String forSome x not in name), _)
[warn]                 .map {
[warn]                      ^
[warn] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/swagger/ResolvedType.scala:79: match may not be exhaustive.
[warn] It would fail on the following input: ADT(_, _, _, _)
[warn]           .map {
[warn]                ^
[warn] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/swagger/ResolvedType.scala:91: match may not be exhaustive.
[warn] It would fail on the following input: ADT(_, _, _, _)
[warn]           .map {
[warn]                ^
[warn] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/swagger/ResolvedType.scala:101: match may not be exhaustive.
[warn] It would fail on the following input: ADT(_, _, _, _)
[warn]           .map {
[warn]                ^
[warn] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/swagger/ResolvedType.scala:112: match may not be exhaustive.
[warn] It would fail on the following input: ADT(_, _, _, _)
[warn]           .map {
[warn]                ^
[warn] 7 warnings found
```
after that I think we can squash and merge this into @stanislav-chetvertkov's branch

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
```diff
@@ -313,45 +324,59 @@ object CirceProtocolGenerator {
...
-      case DecodeADT(clsName, needCamelSnakeConversion) =>
-        val code = if (needCamelSnakeConversion) {
-          q"""implicit val decoder: Decoder[${Type.Name(clsName)}] =
-                  deriveDecoder[${Type.Name(clsName)}](io.circe.derivation.renaming.snakeCase)"""
-        } else {
-          q"""implicit val decoder: Decoder[${Type.Name(clsName)}] = deriveDecoder[${Type.Name(clsName)}]"""
-        }
+      case DecodeADT(clsName, children) =>
+        val childrenCases = children.map(
+          child => p"case ${Lit.String(child)} => c.as[${Type.Name(child)}]"
```
blast-hardcheese: This requires the child types have the same name as the discriminator values? Is there any chance of these getting out of sync or conflicting in some way?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
```diff
@@ -313,45 +324,59 @@ object CirceProtocolGenerator {
...
           .map { t =>
-            q"""def ${Term.Name(t.name.value)} : ${t.decltpe.getOrElse(Type.Name("Any"))}"""
+            val tpe: Type = t.decltpe
+              .flatMap({
+                case tpe: Type => Some(tpe)
+                case x =>
+                  println(s"Unsure how to map ${x.structure}, please report this bug!")
+                  None
+              })
+              .get
```
blast-hardcheese: Instead of using `get`, this should be a `Target.error("<error message>")` instead of using `Target.pure` below

blast-hardcheese:
The tests look great, one other thing though: what happens if there are more than one `$ref` in the `allOf`?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
dpblh: Currently discriminator/mapping doesn't support. Of course it's make sense to support that feature, but could I make it later?)

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
```diff
@@ -169,15 +166,25 @@ object CirceProtocolGenerator {
           } yield ProtocolParameter(term, name, dep, readOnlyKey, emptyToNullKey)
         }
 
-      case RenderDTOClass(clsName, terms, parentNameOpt) =>
+      case RenderDTOClass(clsName, selfTerms, parents) =>
+        val discriminator = parents.find(_.discriminator.isDefined).flatMap(_.discriminator) //collectFirst { case SuperClass(_, _, _, Some(discriminator)) => discriminator }
```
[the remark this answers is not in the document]
dpblh: Probably it should be last in hierarchy, but I have no test. I'll do it.

dpblh:
> The tests look great, one other thing though: what happens if there are more than one `$ref` in the `allOf`?

At the moment I implemented only one $ref

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
dpblh: It's copy-past from DecodeModel.
How should I do it? I mean. We have list of properties and if any one property is incorrect. Should I return `Target.error("<error message>")` or might I just skip the property?

blast-hardcheese: Definitely use `Target.error`. Dropping a property is not adhering to the protocol specification. It's important that we raise this error to the user quickly instead of having them hunt through the build logs to find out why.

I have a change to the sbt and Maven plugins that fails compilation if there are any errors, as currently we just fail to emit any code and continue compiling.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
blast-hardcheese: I'm thinking more about using a path for the type instead of a bare identifier.

Consider if my discriminator is `Either` or `Int` or something like that.

Better to use a structure like `AbstractPet.Dog` to avoid this entirely, as it is impossible for user input to cause a failure.

dpblh:
pull request into master https://github.com/twilio/guardrail/pull/104

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
dpblh: I refer to the swagger's documentation. And there is no word about that(
https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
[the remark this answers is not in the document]
dpblh: I thought a lot. And tried to get some information about that. But OpenAPI documentation have no mention of so hard logic.  I suggest to take first discriminators in hierarchy.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
dpblh: None.get it's impossible situation. If we take a look at the part of code
```
(finalDeclType, finalDefaultValue) = Option(property.getRequired)
              .filterNot(_ == false)
              .fold[(Type, Option[Term])](
                (t"Option[${tpe}]", Some(defaultValue.fold[Term](q"None")(t => q"Option($t)")))
              )(Function.const((tpe, defaultValue)) _)
            term = param"${Term.Name(argName)}: ${finalDeclType}".copy(default = finalDefaultValue)
            dep  = rawDep.filterNot(_.value == clsName) // Filter out our own class name
```
we can to see `param"${Term.Name(argName)}: ${finalDeclType}".`. `Type` is necessary.

blast-hardcheese:
Closing this in favor of #104

Issue #68, opened 2018-07-04
http4s server generation


blast-hardcheese:
Related to #31, currently only akka-http server stubs can be generated.

- [ ] Collect samples of generic servers covering supported use cases (and edge cases)
- [ ] Figure out how to wire in JSON de/serialization
- [ ] Write round-trip tests for http4s, similar to how akka-http round-trip tests work

blast-hardcheese:
Resolved in #102

PR #102, opened 2018-09-30
http4s server generation #68

cponomaryov:


cponomaryov:
@blast-hardcheese could you please take a look?

blast-hardcheese:
@cponomaryov Definitely, though I can't do this immediately.

From the initial review, it looks solid, and the contribution is very much appreciated.

Are you able to get by for now either using the CLI flags with a prebuilt jar or via publishing to a local repository?

cponomaryov:
@blast-hardcheese we would like to use it in production, so it would be nice to have it in maven central maybe with some experimental version. If it's not possible, we will wait for your review.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -0,0 +1,561 @@
...
+          arg => _ => Target.pure(p"${Term.Name(s"${operationId.capitalize}${arg.paramName.value.capitalize}Matcher")}(${Pat.Var(arg.paramName)})")
+        )(params).map {
+          case Nil => Option.empty
+          case x :: xs =>
+            Some(xs.foldLeft[Pat](x) { case (a, n) => p"${a} +& ${n}" })
+      }
+
+    def formToHttp4s: List[ScalaParameter] => Target[List[Enumerator]] =
+      directivesFromParams(
+        arg => _ => Target.pure(enumerator"${Pat.Var(arg.paramName)} = urlForm.values.get(${arg.argName.toLit}).get.head"),
```
blast-hardcheese: If a parameter doesn't match in akka-http, the routing layer falls down to the next method. If no routes match, a 404 is generated. It seems as though using `.get` and `head` would throw exceptions, which would cause a server error instead of user error.

Without pushing form handling into the routing layer, I don't immediately see a better solution than what you have, though possibly introducing a mandatory `400 Bad Request` in a method's response and using that in the `fold` after doing all the argument parsing safely would be an acceptable strategy.

blast-hardcheese: Additionally, to work around not having literal `enum` support (and to support more complex domain types), it's currently acceptable to use `x-scala-type` like this: https://github.com/twilio/guardrail/blob/master/modules/sample/src/main/resources/formData.yaml#L34

I noticed that we didn't have any generator/compiler tests for that file, so I added the following to `build.sbt`:
```diff
diff --git a/build.sbt b/build.sbt
index 2cc58dd..2cd17a8 100644
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,5 @@
+import scala.sys.process._
+
 val projectName = "guardrail-root"
 name := projectName
 organization in ThisBuild := "com.twilio"
@@ -61,12 +63,16 @@ fullRunTask(
   --client --specPath modules/sample/src/main/resources/custom-header-type.yaml --outputPath modules/sample/src/main/scala --packageName tests.customTypes.customHeader.client.akkaHttp --framework akka-http
   --server --specPath modules/sample/src/main/resources/custom-header-type.yaml --outputPath modules/sample/src/main/scala --packageName tests.customTypes.customHeader.server.http4s --framework http4s
   --server --specPath modules/sample/src/main/resources/custom-header-type.yaml --outputPath modules/sample/src/main/scala --packageName tests.customTypes.customHeader.server.akkaHttp --framework akka-http
+  --client --specPath modules/sample/src/main/resources/formData.yaml --outputPath modules/sample/src/main/scala --packageName form.client.akkaHttp --tracing --framework akka-http
+  --client --specPath modules/sample/src/main/resources/formData.yaml --outputPath modules/sample/src/main/scala --packageName form.client.http4s --tracing --framework http4s
+  --server --specPath modules/sample/src/main/resources/formData.yaml --outputPath modules/sample/src/main/scala --packageName form.server.akkaHttp --tracing --framework akka-http
+  --server --specPath modules/sample/src/main/resources/formData.yaml --outputPath modules/sample/src/main/scala --packageName form.server.http4s --tracing --framework http4s
 """.replaceAllLiterally("\n", " ").split(' ').filter(_.nonEmpty): _*
 )
 
 artifact in (Compile, assembly) := {
   val art = (artifact in (Compile, assembly)).value
-  art.copy(`classifier` = Some("assembly"))
+  art.withClassifier(Some("assembly"))
 }
 
 addArtifact(artifact in (Compile, assembly), assembly)
diff --git a/project/build.properties b/project/build.properties
index 133a8f1..0cd8b07 100644
--- a/project/build.properties
+++ b/project/build.properties
@@ -1 +1 @@
-sbt.version=0.13.17
+sbt.version=1.2.3
```
frustratingly, I hit some sort of bug in sbt's macro compilation for `runExample`, but upgrading to the latest sbt seems to have fixed it. (though all the same plugin versions were crosspublished to 1.x, so that was fortunate)

modules/sample/src/test/scala/core/Http4s/Http4sRoundTripTest.scala
```diff
@@ -0,0 +1,269 @@
...
+          petId: PositiveLong,
+          additionalMetadata: Option[String] = None,
+          file: Option[fs2.Stream[IO, Byte]] = None,
+          file2: fs2.Stream[IO, Byte],
+          file3: fs2.Stream[IO, Byte],
+          longValue: Long,
+          customValue: PositiveLong,
+          customOptionalValue: Option[PositiveLong] = None) = {
+        //val f1Length = file.flatMap({ case (f, _, _) => if (f.exists) { Some(f.length) } else None })
+        //val f2Length = if (file2._1.exists) { Some(file2._1.length) } else None
```
blast-hardcheese: Is there something useful that can be done to consume the provided streams?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
cponomaryov: Introduced a `400 Bad Request` if a parameter doesn't match. Added support for more complex domain types. Added tests for `formData.yaml` file.

modules/sample/src/test/scala/core/Http4s/Http4sRoundTripTest.scala
cponomaryov: Added consumption of the provided streams.

modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
```diff
@@ -494,6 +494,40 @@ object SwaggerUtil {
...
+                  case t"BigDecimal" => Right(p"BigDecimalVar(${Pat.Var(paramName)})")
+                  case t"Int"        => Right(p"IntVar(${Pat.Var(paramName)})")
+                  case t"Long"       => Right(p"LongVar(${Pat.Var(paramName)})")
+                  case t"BigInt"     => Right(p"BigIntVar(${Pat.Var(paramName)})")
+                  case tpe @ Type.Name(_) =>
+                    Right(p"${Term.Name(s"${tpe}Var")}(${Pat.Var(paramName)})")
+                }
+              } { _ =>
+                //todo add support for regex segment
+                Left("Unsupported feature")
```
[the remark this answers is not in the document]
cponomaryov: It makes sense, thanks

cponomaryov:
@blast-hardcheese I made changes according to your comments. Could you please take a look when you have time. Thanks.

cponomaryov:
@blast-hardcheese thanks, when we can expect a new version of `guardrail-maven-plugin` with these changes?

blast-hardcheese:
@cponomaryov I'm waiting for maven central to sync

cponomaryov:
@blast-hardcheese thanks

blast-hardcheese:
@cponomaryov OK, it should be available in five to ten minutes

PR #104, opened 2018-10-03
Deep hierarchy

dpblh:
original https://github.com/twilio/guardrail/pull/98

I couldn't merge it twice :(


dpblh:
@blast-hardcheese Could you check it again?

dpblh:
@blast-hardcheese approve and merge it soon. please.

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
```diff
@@ -142,48 +243,121 @@ object ProtocolGenerator {
...
-      protoImports      <- protocolImports
-      pkgImports        <- packageObjectImports
-      pkgObjectContents <- packageObjectContents
-      strictElems = ProtocolElems.resolve(elems).right.get
-    } yield ProtocolDefinitions(strictElems, protoImports, pkgImports, pkgObjectContents)
+      protoImports      <- protocolImports()
+      pkgImports        <- packageObjectImports()
+      pkgObjectContents <- packageObjectContents()
+
+      polyADTElems                           = ProtocolElems.resolve(polyADTs).right.get
```
blast-hardcheese: Why can't `.right.get` be handled in a way that doesn't throw exceptions?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
```diff
@@ -284,76 +308,76 @@ object CirceProtocolGenerator {
...
+          .filter(_.name.value != discriminator)
+          .map { t =>
+            val tpe: Type = t.decltpe
+              .flatMap({
+                case tpe: Type => Some(tpe)
+                case x =>
+                  println(s"Unsure how to map ${x.structure}, please report this bug!")
+                  None
+              })
+              .get
```
blast-hardcheese: Please `traverse` in order to use `Target.error` instead of throwing an exception

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/ProtocolSupportTermInterp.scala
```diff
@@ -0,0 +1,91 @@
...
+  private def extractConcreteTypes[T](definitions: List[(String, Model)]): EitherT[Settings, String, List[PropMeta]] =
+    Target.getGeneratorSettings.flatMap { implicit gs: GeneratorSettings =>
+      for {
+        entries <- definitions.traverse {
+          case (clsName, impl: ModelImpl) if Option(impl.getProperties).isDefined || Option(impl.getEnum).isDefined =>
+            val resolvedType: SwaggerUtil.ResolvedType = SwaggerUtil.Resolved(Type.Name(clsName), None, None)
+            Target.pure((clsName, resolvedType))
+
+          case (clsName, comp: ComposedModel) =>
+            //TODO start off here!!!!!
```
blast-hardcheese: What is this for?

modules/codegen/src/main/scala/com/twilio/guardrail/swagger/Escape.scala
```diff
@@ -0,0 +1,82 @@
+package com.twilio.guardrail.swagger
+
+import scala.meta.{ Init, Term, Tree, Type }
+import scala.util.matching.Regex
+
+object Escape {
```
blast-hardcheese: All the manual tree escaping was removed, this logic is all handled by scalameta properly now.

modules/codegen/src/main/scala/com/twilio/guardrail/swagger/ResolvedType.scala
```diff
@@ -0,0 +1,129 @@
...
+      F.foldMap(value) { a =>
+        f(a) match {
+          case Left(b)  => (A.pure(b), A.empty[C])
+          case Right(c) => (A.empty[B], A.pure(c))
+        }
+      }
+    }
+  }
+
+//  def resolve_(values: List[(String, ResolvedType)]): Target[List[(String, Resolved)]] = {
```
blast-hardcheese: What's this for?

modules/codegen/src/main/scala/com/twilio/guardrail/swagger/SwaggerUtil.scala
```diff
@@ -1,27 +1,46 @@
...
+
+  sealed trait LazyResolvedType extends ResolvedType
+
+  case class Deferred(value: String) extends LazyResolvedType
+
+  case class DeferredArray(value: String) extends LazyResolvedType
+
+  case class DeferredMap(value: String) extends LazyResolvedType
+
+  //fixme
```
blast-hardcheese: What's left to do?

modules/codegen/src/main/scala/com/twilio/guardrail/swagger/SwaggerUtil.scala
```diff
@@ -38,18 +57,20 @@ object SwaggerUtil {
...
+        .map(_._2.tpe)
+        .map(x => (clsName, Resolved(f(x), None, None)))
+
     def resolve_(values: List[(String, ResolvedType)]): Target[List[(String, Resolved)]] = {
       val (lazyTypes, resolvedTypes) = Foldable[List].partitionEither(values) {
         case (clsName, x: Resolved)         => Right((clsName, x))
         case (clsName, x: LazyResolvedType) => Left((clsName, x))
       }
 
-      def lookupTypeName(clsName: String, tpeName: String, resolvedTypes: List[(String, Resolved)])(f: Type => Type): Option[(String, Resolved)] =
```
blast-hardcheese: Why did the scoping of this function get expanded? It's not used anywhere else currently

modules/codegen/src/main/scala/com/twilio/guardrail/swagger/SwaggerUtil.scala
```diff
@@ -131,96 +154,98 @@ object SwaggerUtil {
             }
           } yield res
         case impl: ModelImpl =>
+          println(s"${impl.getDescription} (${impl.getEnum} ${impl.getName} ${impl.getType} ${impl.getFormat})")
```
blast-hardcheese: No new `println`s please

modules/codegen/src/main/scala/com/twilio/guardrail/swagger/SwaggerUtil.scala
```diff
@@ -131,96 +154,98 @@ object SwaggerUtil {
...
+              s"Unable to resolve type for ${impl.getDescription} (${impl.getEnum} ${impl.getName} ${impl.getType} ${impl.getFormat})"
             )
-          } yield Resolved(typeName(tpeName, Option(impl.getFormat()), ScalaType(impl)), None, None)
+          } yield Resolved(typeName(tpeName, Option(impl.getFormat), ScalaType(impl)), None, None)
       }
     }
 
   case class ParamMeta(tpe: Type, defaultValue: Option[Term])
-  def paramMeta[T <: Parameter](param: T): Target[ParamMeta] = {
-    def getDefault[U <: AbstractSerializableParameter[U]: Default.GetDefault](p: U): Option[Term] =
```
blast-hardcheese: Why did the scoping for `getDefault` get expanded?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/circe/protocol/CirceProtocolGenerator.scala
blast-hardcheese: (Previous discussion: https://github.com/twilio/guardrail/pull/98#discussion_r221834486)
Despite `None` being impossible, there's still a `println` there. As `traverse` would make using `Target.error` very easy, I don't see a good reason to use `.get` even if it is impossible to fail.

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
```diff
@@ -142,48 +243,121 @@ object ProtocolGenerator {
...
+    }
+
+    def classHierarchy(cls: String, model: Model): ClassHierarchy = ClassHierarchy(
+      cls,
+      model,
+      children(cls, model),
+      model match {
+        case m: ModelImpl     => Option(m.getDiscriminator)
+        case c: ComposedModel => firstInHierarchy(c).map(_.getDiscriminator)
+        case _                => None
```
blast-hardcheese: What is the use of a non-model as the child of a hierarchy? Would it be better for `classHierarchy` to return an `Option[ClassHierarchy]` or even a `Target.error` and make `discriminator` mandatory to avoid the `discriminator.get` down below?

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
```diff
@@ -95,37 +192,41 @@ object ProtocolGenerator {
...
-  def modelTypeAlias[F[_]](clsName: String, model: ModelImpl)(
+  def modelTypeAlias[F[_]](clsName: String, abstractModel: Model)(
       implicit A: AliasProtocolTerms[F],
       F: FrameworkTerms[F]
   ): Free[F, ProtocolElems] = {
     import F._
+    val model = abstractModel match {
+      case m: ModelImpl => Some(m)
+      case m: ComposedModel =>
+        m.getAllOf.asScala.toList.get(1).flatMap {
```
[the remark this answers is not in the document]
dpblh: Technically `getAllOf` can not be `null`. Because `ComposedModel` should contains at least himself.

dpblh:
@blast-hardcheese Would you like to check & approve it?)

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
blast-hardcheese: this comment hasn't been resolved

blast-hardcheese: this comment hasn't been resolved

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
blast-hardcheese: this comment hasn't been resolved

blast-hardcheese:
Thank you for your work on this; I'll review, but there were still some points from the last review

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
dpblh: this comment has been resolved. look at the code more attentively :)

blast-hardcheese: I hadn't looked at the code yet, as I'm not at my computer; just going off what comments hadn't been resolved by changes. I'll look.

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
blast-hardcheese: I'll look

modules/codegen/src/main/scala/com/twilio/guardrail/Common.scala
```diff
@@ -12,6 +12,9 @@ import com.twilio.guardrail.terms.{ CoreTerm, CoreTerms, ScalaTerms, SwaggerTerm
 import com.twilio.guardrail.terms.framework.FrameworkTerms
 import com.twilio.guardrail.generators.GeneratorSettings
 import java.nio.file.{ Path, Paths }
+
+import com.twilio.guardrail.protocol.terms.protocol.PolyProtocolTerms
```
blast-hardcheese: consolidate imports

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
```diff
@@ -95,37 +182,41 @@ object ProtocolGenerator {
       * so essentially we have to return false if:
       *   - there are any uppercase characters
       */
-    def couldBeSnakeCase(s: String): Boolean = s.toLowerCase(Locale.US) == s
```
blast-hardcheese: This function was moved without its comment

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -90,7 +90,13 @@ object CirceProtocolGenerator {
   object ModelProtocolTermInterp extends (ModelProtocolTerm ~> Target) {
     def apply[T](term: ModelProtocolTerm[T]): Target[T] = term match {
       case ExtractProperties(swagger) =>
-        Target.pure(Either.fromOption(Option(swagger.getProperties()).map(_.asScala.toList), "Model has no properties"))
+        Target.pure(
+          (swagger match {
+            case m: ModelImpl        => Option(m.getProperties)
+            case comp: ComposedModel => comp.getAllOf().asScala.toList.get(1).flatMap(prop => Option(prop.getProperties))
```
blast-hardcheese: What's the `get(1)` referring to? By writing things this way, we're only permitting `allOf` with two members, where the first member is the superclass, and the second is a model whose properties get appended to the first member?

If this is the case, is it possible to `.traverse(prop => Option(prop.getProperties))` to accumulate all union members? (I didn't see any tests for combining more than two models together; If I'm misreading and this is actually already possible, my apologies)

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -158,12 +164,24 @@ object CirceProtocolGenerator {
...
-        """)
+      case RenderDTOClass(clsName, selfTerms, parents) =>
+        val discriminator = parents.find(_.discriminator.isDefined).flatMap(_.discriminator) //collectFirst { case SuperClass(_, _, _, Some(discriminator)) => discriminator }
+        val parentNameOpt = parents.headOption.map(_.clsName)
+        val terms = (parents.reverse.flatMap(_.params.map(_.term)) ++ selfTerms).filterNot(
+          param => discriminator.contains(param.name.value)
+        )
+        val code = parentNameOpt
+          .fold(q"""case class ${Type.Name(clsName)}(..${terms})""")(
+            parentName => q"""case class ${Type.Name(clsName)}(..${terms}) extends ${template"${init"${Type.Name(parentName)}(...$Nil)"}"}"""
```
blast-hardcheese: Is it necessary to do this `${template"${init"${Type.Name` stuff? It seems as though `extends ${Type.Name("foo")}(...${Nil})` actually works:
```
scala> q"case class Foo(..${List(param"blop: Blop")}) extends ${Type.Name("blep")}(...${List(List(Term.Name("blop")))})"
res12: meta.Defn.Class = case class Foo(blop: Blop) extends blep(blop)
```

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -356,4 +383,62 @@ object CirceProtocolGenerator {
...
+          .filter(_.name.value != discriminator)
+          .map { t =>
+            val tpe: Type = t.decltpe
+              .flatMap({
+                case tpe: Type => Some(tpe)
+                case x =>
+                  println(s"Unsure how to map ${x.structure}, please report this bug!")
+                  None
+              })
+              .get
```
```scala
      case RenderSealedTrait(className, terms, discriminator, parents) =>
        for {
          testTerms <- (
            terms
              .filter(_.name.value != discriminator)
              .traverse { t =>
                for {
                  tpe <- Target.fromOption(
                    t.decltpe
                      .flatMap({
                        case tpe: Type => Some(tpe)
                        case x => None
                      })
                    , s"Unsure how to map ${x.structure}, please report this bug!")
                } yield q"""def ${Term.Name(t.name.value)}: ${tpe}"""
              }
          )
        } yield {
          parents
            .headOption
            .fold(q"""trait ${Type.Name(className)} {..${testTerms}}""")(
              parent => q"""trait ${Type.Name(className)} extends ${template"${init"${Type.Name(parent.clsName)}(...$Nil)"}{..${testTerms}}"} """
            )
        }
```
please

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
```diff
@@ -151,28 +242,105 @@ object ProtocolGenerator {
...
+      P: PolyProtocolTerms[F]
+  ): Free[F, ProtocolDefinitions] = {
     import S._
     import F._
+    import P._
 
     val definitions = Option(swagger.getDefinitions).toList.flatMap(_.asScala)
+    val hierarchies = groupHierarchies(definitions)
+
+    // todo without trait
```
blast-hardcheese: What's this referring to?

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
blast-hardcheese: I can't see what this initially referred to, resolving this conversation

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
dpblh: Sorry. I've wrote it before. `allOf` contains as first element own model and as second one superclass. In the swagger documentation no word about multiple inheritance.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
dpblh: Thanks for it.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
dpblh: Thank you so much. I've only changed one line.
`
- s"Unsure how to map ${x.structure}, please report this bug!")
+ t.decltpe.fold("Nothing to map")(x => s"Unsure how to map ${x.structure}, please report this bug!"))
`

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
dpblh: Nothing. It was removed.

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
dpblh: I really didn't want it. Couldn't we skip it?)

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
blast-hardcheese: Ah, yes; we talked about it in one of the other PRs. Thank you for explaining again.

Still, I'm wondering if something like:
```yaml
	  Cat:
    description: A representation of a cat
    allOf:
    - $ref: '#/definitions/Pet'
    - $ref: '#/definitions/Mammal'
    - type: object
      properties:
        huntingSkill:
          type: string
          description: The measured skill for hunting
          default: lazy
          enum:
          - clueless
          - lazy
          - adventurous
          - aggressive
```
would continue to add to this list. Could a dynamic approach be taken instead of hard-coding `1` here?

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
blast-hardcheese: Without moving the comment, now it just applies to nothing. Do you think there's a better place for this comment to live?

dpblh: Oops! I'm really sorry. I lost it.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
dpblh: I can't give you quickly answer. I need time to research.

blast-hardcheese:
This is wildly cool. I'm going to do one more full review and some local testing before merging.

blast-hardcheese:
🎉

Issue #33, opened 2018-02-01
Custom header types


blast-hardcheese:
`x-scala-type` doesn't seem to be fully respected for header parameters.

Suggested course of action:
Write a test with a spec with the parameter `x-scala-type: T` (for some `T`) and verify that the parameter is of type `T` and is handled via some `Marshaller[T, String]` before being supplied to the header constructor.

PR #124, opened 2018-10-31
Add test to check structure & client for custom header type

at1as:
As a ramp up to something more substantial, I'd like to start by tackling: https://github.com/twilio/guardrail/issues/33

A few questions about Issue 33 (to make sure I'm on the right track before continuing):

- is the assertion that header type parameters are not being respected for the client, the server, or both?

- is something like the attached approach answering the correct question? I'm setting `x-scala-type` to be something custom for the body and the header, and checking the generated code to see if the header is treated is like the body (the custom type is part of the definition)

### DIFF
--- a/modules/codegen/src/test/scala/core/issues/Issue33.scala
+++ b/modules/codegen/src/test/scala/core/issues/Issue33.scala
@@ -1,0 +1,95 @@
+package tests.core.issues
+
+import _root_.io.swagger.parser.SwaggerParser
+import cats.instances.all._
+import com.twilio.swagger._
+import com.twilio.guardrail.generators.AkkaHttp
+import com.twilio.guardrail._
+import com.twilio.guardrail.EnumDefinition
+import com.twilio.guardrail.tests._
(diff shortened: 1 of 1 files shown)


modules/codegen/src/test/scala/core/issues/Issue33.scala
```diff
@@ -0,0 +1,95 @@
...
+    |  /foo:
+    |    get:
+    |      operationId: getFoo
+    |      parameters:
+    |      - name: CustomHeader
+    |        in: header
+    |        type: string
+    |        x-scala-type: Bar
+    |        required: true
+    |      - name: CustomFormData
```
blast-hardcheese: I'd just remove the form parameters entirely, just having a single header parameter. In this example, `Bar` being an enumeration is really good, as this isn't supported by OpenAPI 2.0 currently.

blast-hardcheese:
This can be reopened when you're able to put more effort towards it, @at1as

PR #125, opened 2018-10-31
Sort routes to ensure deterministic output

easel:
I've been getting increasingly frustrated with my play project continually recompiling my guardrail routes and finally dug in to find out what was happening. In short, there are a couple of places in the route generation pipeline for akka http server that use groupBy and never sort the results, producing non-determinstic ordering in the Routes.scala files, causing the Scala compiler to have to compile and slow everything down.

I've hacked in sorting on the places I've found, and it resolves my issue. The failing tests are due to hardcoded output in a different order. I'm not sure how you're getting ahold of the generated .scala files to put in the tests so I thought I'd throw this up for comment.

Fixes https://github.com/twilio/sbt-guardrail/issues/8

### DIFF
--- a/modules/codegen/src/test/scala/tests/generators/akkaHttp/StaticParametersTest.scala
+++ b/modules/codegen/src/test/scala/tests/generators/akkaHttp/StaticParametersTest.scala
@@ -37,8 +37,8 @@ class StaticParametersTest extends FunSuite with Matchers with SwaggerSpecRunner
 
     val handler = q"""
       trait Handler {
-        def getFoo1(respond: Resource.getFoo1Response.type)(): scala.concurrent.Future[Resource.getFoo1Response]
         def getFoo2(respond: Resource.getFoo2Response.type)(): scala.concurrent.Future[Resource.getFoo2Response]
+        def getFoo1(respond: Resource.getFoo1Response.type)(): scala.concurrent.Future[Resource.getFoo1Response]
       }
     """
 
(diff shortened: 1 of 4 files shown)


blast-hardcheese:
This looks good, though it's surprising to me that I haven't run into this problem myself.

If you wouldn't mind fixing the failing test, that would be appreciated

easel:
@blast-hardcheese any hints where I can grab the generated .scala files from during the tests? The failures have a bunch of ScalaMeta AST spam that's going to be very difficult to convert back to the source code quoted in the tests.

blast-hardcheese:
I just add a println before the `===` to get the raw source, then paste that over the `q"""` blocks. The tests don't hit the filesystem.

easel:
Aha! Tests incoming shortly =p

easel:
Ok. Tests should pass CI now, they pass locally.

modules/codegen/src/test/scala/tests/generators/akkaHttp/StaticParametersTest.scala
```diff
@@ -37,8 +37,8 @@ class StaticParametersTest extends FunSuite with Matchers with SwaggerSpecRunner
 
     val handler = q"""
       trait Handler {
-        def getFoo1(respond: Resource.getFoo1Response.type)(): scala.concurrent.Future[Resource.getFoo1Response]
         def getFoo2(respond: Resource.getFoo2Response.type)(): scala.concurrent.Future[Resource.getFoo2Response]
+        def getFoo1(respond: Resource.getFoo1Response.type)(): scala.concurrent.Future[Resource.getFoo1Response]
```
blast-hardcheese: What causes `getFoo1` to sort below `getFoo2`? That's counterintuitive

blast-hardcheese: Oh, `sortBy(path, method)`; that makes sense.

easel:
Thanks for the merge!

PR #132, opened 2018-11-08
Bump http4s to 0.19.0

cponomaryov:


modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -369,8 +370,8 @@ object CirceProtocolGenerator {
           val decodeLong = implicitly[Decoder[Long]]
 
           implicit def decodeInstant: Decoder[Instant] = j8time.decodeInstant.or(decodeLong.map(Instant.ofEpochMilli))
-          implicit def decodeLocalDate: Decoder[LocalDate] = j8time.decodeLocalDateDefault.or(decodeInstant.map(_.atZone(ZoneOffset.UTC).toLocalDate))
-          implicit def decodeOffsetDateTime: Decoder[OffsetDateTime] = j8time.decodeOffsetDateTimeDefault.or(decodeInstant.map(_.atZone(ZoneOffset.UTC).toOffsetDateTime))
+          implicit def decodeLocalDate: Decoder[LocalDate] = j8time.decodeLocalDateWithFormatter(ISO_LOCAL_TIME).or(decodeInstant.map(_.atZone(ZoneOffset.UTC).toLocalDate))
```
```scala
def mirrorCirceTimeImplicit[T]: T = {
  import io.circe.java8.time._
  implicityly[T]
}
```

then

```scala
mirrorCirceTimeImplicit[Decoder[LocalDate]]
```
at least for a few releases of circe, to ease the transition between versions.

cponomaryov: Fixed in similar way

cponomaryov: Why do we need this?

blast-hardcheese: I want to avoid forcing too many dependency changes at the same time.

The plugins don't do any dependency analysis to determine which API compatibility level guardrail generates code for, so the consumer is just greeted with arbitrary compiler failures.

Also, if a consumer is locked into using one version of a downstream library (circe is auxillary, http4s is primary) than they can't upgrade.

As this is not a critical change in circe, I want to avoid penalizing the user.

blast-hardcheese:
Merged in #135

blast-hardcheese:
Just finished publishing v0.41.1, thank you for your patience

cponomaryov:
@blast-hardcheese thank you so much

PR #133, opened 2018-11-08
Language ambivalence

blast-hardcheese:
Starting to address #36 

The work involved in providing support for different languages is effectively the following:

- [x] Encode existing Free algebras with knowledge of which language they are targeting
- [ ] Remove all references to scala.meta from everything but the `ScalaLanguage` specification class and the generators
- [ ] Add a `language` parameter to the argument parser/context
- [ ] Add a java generators for akka-http clients and servers, using javaparser
- [ ] Add a Jackson generator, wire it in to the java-akka stuff [according to the docs](https://doc.akka.io/docs/akka-http/10.0.6/java/http/common/json-support.html)


blast-hardcheese:
I definitely don't expect to tick all those boxes in this PR, but this is the general direction this work is going in.

I'm also interested in feedback on the language encoding.

I was pleased that this worked. Initially I was trying to use path dependent types, and that turned out to be a very bad idea, since there was no proof that the instances for the values at the different paths were related at all, even if they were the same object just passed around.

PR #138, opened 2018-12-02
Scala 2.12.8

sullis:


stanislav-chetvertkov:
@blast-hardcheese Is the compiler bug issue https://github.com/twilio/guardrail/commit/cc20e726a89d17b1779a220114f6a05c3c9de2a4 still relevant?

blast-hardcheese:
Yes, I'll check to see if the issue is fixed in this new version

blast-hardcheese:
Ah, yes, even 2.12.8 snapshot had this issue, and there's been no movement on https://github.com/scala/bug/issues/11243

tomasherman:
any reason this isn't merged? it would help with metals integration ;)

tomasherman:
is the bug a big deal? It's a bit sad this can't be merged as it blocks `metals` from working properly :)

blast-hardcheese:
This bug was causing errors in completely unrelated code; it took an hour to resolve a relatively simple bug because the error was at the end of a very large function. This is very similar to

```scala
case Foo(a,b,c) =>
  def foo = ???
  foo
```
being completely fine for months, then something small changes and 
```scala
def foo = ???

case Foo(a,b,c) =>
  foo
```
is required to get it to compile again (the current assumption is due to some unsurfaced path dependence on return types)

The motivation of getting metals working with guardrail is definitely understandable. If I can get a repro working in scalac directly, I may be able to bisect to figure out what caused the bug and fix it.

Linking to https://github.com/scala/bug/issues/11243 for visibility

PR #149, opened 2018-12-24
Coreterm LanguageAbstraction

blast-hardcheese:
Breaking out `ScalaLanguage` from the last little bits of `CoreTerm`/`CodegenApplication`

This pushes language evaluation to the very edge of the world for the CLI interpreter by prepending the language desired for processing all arguments to the first CLI argument (ie: `cli.sh scala --client ...`)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -23,10 +23,9 @@ val scalatestVersion  = "3.0.5"
 
 mainClass in assembly := Some("com.twilio.guardrail.CLI")
 
-lazy val runExample: TaskKey[Unit] = taskKey[Unit]("Run with example args")
-// TODO: akka.NotUsed should exist in all generated sources, but there are no import verifying tests yet.
+lazy val runScalaExample: TaskKey[Unit] = taskKey[Unit]("Run scala generator with example args")
 fullRunTask(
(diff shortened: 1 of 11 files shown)


blast-hardcheese:
I haven't figured out a way to have the language parameter maintain safety without reflection from inside `CoreTerm[L, ?] ~> CoreTarget`.

I was thinking of `String => LanguageConcretization` combined with `LanguageConcretization ~> (({ type F[L] = CoreTerm[L, ?] })#F ~> CoreTarget)`, but the types don't meet at the end. I'd love a solution to this problem, but I think any additional work I do in this direction at this moment will be beyond me.

I opted to duplicate the entire interpreter, parameterized by some language-specific stuff. This seems to be a fairly straightforward approach, which can hopefully be cleaned up if it gets to be a problem. I expect the build tool plugins will have a far easier time with this than the CLI parser.

kelnos:
Given how small that interpreter is, I don't think it's a big deal to make a copy of it per language...

blast-hardcheese:
At least for now, I agree. If it becomes too burdensome I'm sure something else will present itself.

Issue #148, opened 2018-12-20
Document how to see validation issues while in development


sledorze:
Is there already a solution to this?

sledorze:
Sorry, I mean the server validation of the Json models the endpoint is receiving.

If I send a payload which does not correspond to the Swagger model, I would like to know what's wrong.
May I lack some insight on how to retrieve that from the server implementation (using akka-http) but wondering if something implementation agnostic exists.

blast-hardcheese:
Ah, this currently doesn't exist; I'm thinking it may be how we're doing the content validation exception, failing over to other available unmarshallers.

What kind of message would you expect to see? I'm assuming a Circe path would not be the best user experience; possibly a JSON path?

sledorze:
@blast-hardcheese anything that actually provide an hint *and* supports custom validation messages.
I'm not sure about what looks like a circe-path or JSON path yet.. (I've left Scala since 2012 and am just back to it, landing in the ecosystem - looking at it ATM)

sledorze:
@blast-hardcheese ok, I understand the problem now with circe-path.
I'm not sure but I guess the JSON Path would be extracted from the circe-path so this means we can have some way to handle directly a circe-path and possibly use a transformer to get a JSON path.
Exposing the circe DecodingFailure would be a good and the minimal, necessary first step.

blast-hardcheese:
@sledorze I agree; I've made these changes in the akka-http generator for now.

PR #150, opened 2019-01-06
Exposing helpful error messages

blast-hardcheese:
- Added error messages to polymorphic discriminators (previous error: `scala.MatchError: foo (of class java.lang.String)`)
- Exposed underlying circe decoding errors
- Removed incorrect usage of `NoContentException`

Initial work towards #148, actually exposing validation errors in clients and servers. Customization will be more tricky, still need to think about how that will work.


sledorze:
@blast-hardcheese thanks a lot!

blast-hardcheese:
@sledorze You're welcome! 0.41.4 is now available

PR #152, opened 2019-01-06
Client multi responses

### LINKED ISSUE #151: Support multiple client response codes, currently the "best" success code is used
This has been a significant difference between http4s and akka-http clients for a while.

An initial simplifying assumption was that most routes would only have one "successful" status code, and failure responses could be handled by the consumer.

This is undesirable due to the following points:
- Error handling is potentially dangerous, if consumers forget to `discardEntityBytes()`
- Mixing well typed responses with untyped responses does not give a good user experience

blast-hardcheese:
Depends on #150 
Resolves #151

This extends the akka-http clients to behave similarly to the http4s clients, generating a unique sealed abstract class for each operation and status code, permitting mapping and deserializing all known status codes.


blast-hardcheese:
Two major issues:
first, `jsonEntityUnmarshaller` has been marked deprecated, but it is still used in the multipart form-data unmarshaller.
second, I'd like to improve the user experience of consuming the response types.

Currently, a user needs to import `getFooBarResponse` as a sibling to the `Client`, then `match` on the result of calling `getFooBar`, matching on the different possible status codes.

What would be preferable would be either a `.fold[A]`, representing each status as a partial function from each decoded type to `A`, or some other way to make it easier for consumers to work with potentially unwieldy names.

Additionally, to ease the transition to the new behavior, `.best: Either[getFooBarResponse, FooBar]` could return the "best" status code, based on the current heuristic.

blast-hardcheese:
After playing with the `fold` implementation for a bit, I still feel that it's a compelling direction:
- Discoverability, as the type only offers one top-level method
- Implementation details hidden, reducing burden on the consumer

The tradeoffs are:
- Named parameters are almost a necessity for clarity
- As all parameters are peers, explicitly specifying the result type is likely necessary to prevent divergence

```scala
  import akka.http.scaladsl.model.HttpResponse
  def handleError: Either[Throwable, HttpResponse] => Future[String] = _.fold(ex => Future.successful(ex.getMessage()), { resp =>
    import scala.concurrent.duration._
    resp.entity.toStrict(5.seconds).map { entity =>
      s"${resp.status}: ${entity}"
    }
  })
  client.getPet("foo")
    .leftSemiflatMap(handleError)
    .flatMap(
      _.fold[EitherT[Future, String, IndexedSeq[_root_.client.definitions.Pet]]](
        handleOK = pet => EitherT.pure(IndexedSeq(pet)),     // Pet => A
        handleCreated = EitherT.pure(_),                     // IndexedSeq[Pet] => A
        handleAccepted = EitherT.pure(IndexedSeq.empty),     // => A
        handleNonAuthoritativeInformation = EitherT.pure(_)  // IndexedSeq[Pet] => A
      )
    ).value.onComplete(println)
```

kelnos:
I like the `fold` implementation, and I think sorta "forcing" use of named parameters is a strength, not a tradeoff.  However, what if an endpoint only has a single success status?  What would the return type (after handling errors) look like?

blast-hardcheese:
@kelnos https://github.com/twilio/guardrail/pull/152/files#diff-c99036f0b5ed1567f3f308983e33aa9dR196

This is a single response type that has no response body:
```scala
          def fold[A](handleOk: => A): A = this match {
            case GetBarResponse.Ok => handleOk
          }
```

If the response body had been specified, it'd look like this:
```scala
          def fold[A](handleOk: Foo => A): A = this match {
            case GetBarResponse.Ok(x) => handleOk(x)
          }
```

kelnos:
Having `fold` with a single 'branch' feels odd, but I guess is fine.

PR #158, opened 2019-01-10
Abstract companion objects

kelnos:
Companion objects are a feature somewhat unique to Scala, so we need to abstract them out.  The per-language generators can do what they need to do from there.  For example, the Java generator will turn them into `static` declarations on the associated class (if any).

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolElems.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolElems.scala
@@ -8,6 +8,12 @@ import com.twilio.guardrail.languages.LA
 import com.twilio.guardrail.terms.{ ScalaTerms, SwaggerTerms }
 import com.twilio.guardrail.protocol.terms.protocol.ProtocolSupportTerms
 
+case class StaticDefns[L <: LA](className: String,
+                                extraImports: List[L#Import],
+                                members: List[L#ObjectDefinition],
+                                definitions: List[L#Definition],
+                                values: List[L#ValueDefinition])
+
 sealed trait ProtocolElems[L <: LA]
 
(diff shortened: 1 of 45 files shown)


blast-hardcheese:
@kelnos As we're still dual-shipping 2.11 and 2.12, we can't use trailing commas

kelnos:
@blast-hardcheese ah damn, forgot.  fixed, tests are passing again.

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolElems.scala
```diff
@@ -8,6 +8,12 @@ import com.twilio.guardrail.languages.LA
 import com.twilio.guardrail.terms.{ ScalaTerms, SwaggerTerms }
 import com.twilio.guardrail.protocol.terms.protocol.ProtocolSupportTerms
 
+case class StaticDefns[L <: LA](className: String,
```
blast-hardcheese: I think this may end up being cumbersome in Java, as you'll need to check to make sure the static members are being added to the same class; at some point in the future, we may need `cats.data.Ior[Defn.Class, StaticDefinitions[L]]` to keep them associated. I think most of the time when we're emitting classes, there's only one in scope, so this isn't a problem at the moment.

kelnos: Yeah, I'm not 100% happy with this, but at the same time I do like the fact that these things are separated.  If we mash class+object generation together, then it becomes annoying for Scala.  Maybe that's ok?  I dunno; I'm feeling ok with this incremental change now... what do you think?

kelnos: Oh, you already hit the approve button, so I guess you're ok with it too ;)

PR #159, opened 2019-01-11
Making akka-http FormData return 413 instead of 500, delete failed files

### LINKED ISSUE #143: File uploads that exceed akka-http's limit are terminated without cleanup
The culprit is [`handleExceptions` being called inside the success block](https://github.com/twilio/guardrail/blob/20fb0bcfb136b8825dc027d4d28abbbf9c4a1d39/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala#L474-L486).

The fix seems to be moving this outside [`entity(as[Multipart.FormData]).flatMap`](https://github.com/twilio/guardrail/blob/20fb0bcfb136b8825dc027d4d28abbbf9c4a1d39/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala#L450), but a test for this issue should be written with a custom (ideally very small) file upload limit pointing at a temporary directory, then ensuring the directory is empty after the test runs.

blast-hardcheese:
Resolves #143 

This moves the exception handler up before the route -- the exception handler was previously not used due to being in the wrong place.

Additionally, calling `dest.delete()` _immediately_ upon any exception, to avoid the condition where an exception happens inside `UnmarshalToFile`, preventing the file handle from being added to the `AtomicReference[List[File]]`s to clean up.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -69,6 +69,7 @@ fullRunTask(
   --server --specPath modules/sample/src/main/resources/formData.yaml --outputPath modules/sample/src/main/scala --packageName form.server.http4s --framework http4s
   --server --specPath modules/sample/src/main/resources/formData.yaml --outputPath modules/sample/src/main/scala --packageName form.server.akkaHttp --framework akka-http
   --server --specPath modules/sample/src/main/resources/issues/issue127.yaml --outputPath modules/sample/src/main/scala --packageName issues.issue127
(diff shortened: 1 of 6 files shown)


blast-hardcheese:
Needed to work around https://github.com/akka/akka-http/issues/2381 in the test

PR #165, opened 2019-01-16
fix double trailing slash for http4s server

OleksandrSamsonov:
Fix double trailing slash for http4s server generator.

The issue:
When we have empty path for method like in swagger file below:
```yaml
swagger: 2.0
info:
  description: Some description
schemes:
  - http
paths:
  /:
    post:
      x-scala-package: somePackage
      operationId: someOperationId
      consumes:
        - application/json
      produces:
        - application/json
      responses:
        200:
```
generated server consumes only requests for this path with double slash like so:
http://localhost//


OleksandrSamsonov:
> This looks good, though please also ensure that this does not break trailing-slash matching, otherwise I'll merge and release within the week.
Thanks for good comment.

I have tested this on both cases like:
 http://localhost/, http://localhost
and it works as expect.

According to doc: https://http4s.org/v0.18/dsl/

> A request to the root of the service is matched with the Root extractor. Root consumes the leading slash of the path info. The following matches requests to GET /:

cponomaryov:
@blast-hardcheese could you please merge and release?

blast-hardcheese:
Yes, while reviewing this PR I found a few other things that should change to improve usability and error messages, thank you for the contribution!

PR #173, opened 2019-01-22
Endpoints xhr client

### LINKED ISSUE #169: Add a generator for endpoints to support scalajs clients
After exploring https://github.com/julienrf/endpoints/, I believe adding support for generating endpoints algebras with interpreters down to specific implementations would be beneficial (as a possible resolution to #140, as well as adding support for scalajs)

- [x] Generate endpoints algebras
- [x] Generate endpoints clients based on XMLHttpRequest
  - [x] Basic client generation with empty bodies
  - [x] Static query-string parameters
  - [x] Response status code mapping
  - [x] Request bodies
    - [x] text/plain
    - [x] `multipart/form-data`
      - [x] file uploads
    - [x] `application/x-www-form-urlencoded`
- [ ] End-to-end tests

blast-hardcheese:
Resolves #169 

There are currently no functional tests, other than the fact that the clients compile. I've still got to investigate how to do round-trip tests with a real `XMLHttpRequest` client, but it _does_ correctly handle decoding response bodies for different status codes.

Keeping this PR open for the time being. Those who are interested in trying it out, run the following via either:

### sbt
```
sbt:guardrail-root> cli --client --specPath .../path/to/spec.yaml --outputPath ...path/to/project/src/main/scala --packageName clients --framework endpoints
```

### shell
```
user$ ./cli.sh --client --specPath .../path/to/spec.yaml --outputPath ...path/to/project/src/main/scala --packageName clients --framework endpoints
```


modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
```diff
@@ -0,0 +1,462 @@
...
+
+object EndpointsClientGenerator {
+  object ClientTermInterp extends FunctionK[ClientTerm[ScalaLanguage, ?], Target] {
+
+    private[this] def toDashedCase(s: String): String = {
+      val lowercased =
+        "^([A-Z])".r.replaceAllIn(s, m => m.group(1).toLowerCase(Locale.US))
+      "([A-Z])".r
+        .replaceAllIn(lowercased, m => '-' +: m.group(1).toLowerCase(Locale.US))
+    }
```
kelnos: Tangential, but at some point it would be nice to put all the case converters in some common string-utils object.  Right now they're just all over the place.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
```diff
@@ -0,0 +1,462 @@
...
+            case (a, ScalaParameter(_, param, paramName, argName, _)) =>
+              val lifter: (Term.Name, RawParameterName) => Term = param match {
+                case param"$_: Option[$_]"      => liftOptionTerm _
+                case param"$_: Option[$_] = $_" => liftOptionTerm _
+                case _                          => liftTerm _
+              }
+              a :+ lifter(paramName, argName)
+          }
+          q"scala.collection.immutable.Seq[Option[String]](..$args).flatten"
+        }
```
kelnos: A lot of the form data and header param generation code looks _really_ similar to the scala/akka code.  Is this worth parameterizing?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
```diff
@@ -0,0 +1,462 @@
...
+                }, NonEmptyList(y, rest))
+            }
+
+            def hackyFoldLimitedTupleTree: NonEmptyList[Term] => NonEmptyList[Term] = {
+              case NonEmptyList(x, xs) =>
+                // FIXME: There's a hard limit in endpoints for a maximum Tupler size of 3 elements.
+                // This results in ((((1, 2, 3), 4, 5), 6, 7)...)
+                // As a result, we've got a rather complex structure to replicate this,
+                // bounded by ${limit}. If this limit is ever raised, it'll be a breaking change, unless
+                // this value becomes customizable.
```
kelnos: That's... unfortunate.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsGenerator.scala
```diff
@@ -0,0 +1,195 @@
...
+          case "506" => Target.pure((506, q"VariantAlsoNegotiates"))
+          case "507" => Target.pure((507, q"InsufficientStorage"))
+          case "508" => Target.pure((508, q"LoopDetected"))
+          case "509" => Target.pure((509, q"BandwidthLimitExceeded"))
+          case "510" => Target.pure((510, q"NotExtended"))
+          case "511" => Target.pure((511, q"NetworkAuthenticationRequired"))
+          case "598" => Target.pure((598, q"NetworkReadTimeout"))
+          case "599" => Target.pure((599, q"NetworkConnectTimeout"))
+          case _     => Target.raiseError(s"Unknown HTTP type: ${key}")
+        }
```
kelnos: Isn't this the third copy-paste of this code?  Can we move it somewhere common?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
blast-hardcheese: Possibly. I'm trying to avoid creating more spaghetti, especially as new frameworks and languages come into the mix.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
blast-hardcheese: yes.
[Twitter: image](https://twitter.com/blast_hardchese/status/1087088858368360448)

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsGenerator.scala
blast-hardcheese: Exceedingly unfortunately, not directly, as these are the underlying names of the status code objects in each library. akka-http uses `OK`, for example, where http4s uses `Ok`.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
kelnos: Oh man...

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsGenerator.scala
kelnos: So close, and yet so far...

PR #177, opened 2019-01-26
Initial OpenAPI v3 support

### LINKED ISSUE #54: support for openapi 3.0
Is there support for openapi 3.0.x specification?

blast-hardcheese:
Resolves #54 

Continuing work from @stanislav-chetvertkov on supporting the OpenAPI v3 specification.

- [x] Compiles with some shims, reflection, and casting
- [x] Tests all pass
- [x] Reducing reliance on casting
- [ ] Adding dense OpenAPI v3 specification test

No API changes are expected to the consumer API, meaning that while OpenAPI 3.x specifications are able to be used to generate clients and servers, full and correct adherence to the specification is not a goal of this PR.

Check the OpenAPI 3.x milestone for full support.


modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaParameter.scala
```diff
@@ -140,7 +140,7 @@ object ScalaParameter {
       resolved <- SwaggerUtil.ResolvedType.resolve[L, F](meta, protocolElems)
       SwaggerUtil.Resolved(paramType, _, baseDefaultValue) = resolved
 
-      required = parameter.getRequired()
+      required = Option[java.lang.Boolean](parameter.getRequired()).fold(false)(identity)
```
stanislav-chetvertkov: I still prefer having extension methods defined for swagger model entities to reduce boilerplate for such occasions

modules/codegen/src/test/scala/core/issues/Issue166.scala
```diff
@@ -43,6 +43,8 @@ class Issue166 extends FunSuite with Matchers with SwaggerSpecRunner {
   test("Handle generation of models") {
     val opts = new ParseOptions()
     opts.setResolve(true)
+    opts.setResolveFully(true)
+    opts.setResolveCombinators(true)
```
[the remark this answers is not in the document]
blast-hardcheese: I couldn't find any documentation either, but from my experience, it walks the model after parsing and links `$ref`s -- at least for now, this is advantageous, as the heuristic of `get$ref.split("/").last` will fail when `$ref: #/structures/api1/Pet` and `$ref: #/structures/api2/Pet`. There needs to be a better way to disambiguate.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaParameter.scala
blast-hardcheese: I'm in favor as well. My goal after this is to push all `OpenAPI` knowledge to its own algebra, so future changes can be done from the outside without exposing so much knowledge (`Option`, exceptions, `$ref`, tree-walking) into the core generators. I didn't know this was absolutely necessary before starting this work, but now I'm leaning very heavily in favor of that strategy. It's too much noise right now.

blast-hardcheese:
- [ ] Rebased off master
- [ ] Ensured that the source diffs of the `sample` project are identical between master and this branch

There are some known omissions from this initial attempt at OpenAPI v3. While I'm sure there are others, this is the initial list:

##### `#/components/requestBodies` are not considered at all

This will require a complete overhaul of how `Content-Type`s are handled. Instead of hacking things together, I'd much rather have a more clear strategy of supporting different content types natively in generated clients and servers. This may coincide with work to unify identical definitions between different runs as well.

##### Multiple `Content-Type`s for individual routes
Currently, to the best of our abilities, the new structure is shimmed into the old structure. This is likely not going to work for anything but the most trivial of OpenAPI v3 schemas. v2 should continue to work unchanged, as the v2 -> v3 converter seems to do a good job maintaining compatibility.

PR #200, opened 2019-03-21
Make DW server response classes simpler for the resources to use

kelnos:
This is _mostly_ user-invisible.  The base response class now holds the entity body (if any), and is parameterized on the entity body's type.  Now the resource class method doesn't need to do `instanceof` checks for each response type to attempt to extract entity bodies and can instead just pull out an `Optional<T>` from the abstract base class.

Depends on https://github.com/twilio/guardrail/pull/199

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
@@ -8,11 +8,11 @@ import cats.syntax.traverse._
 import com.github.javaparser.JavaParser
 import com.github.javaparser.ast.Modifier._
 import com.github.javaparser.ast.{ Modifier, NodeList }
-import com.github.javaparser.ast.`type`.{ ClassOrInterfaceType, PrimitiveType, Type, VoidType }
+import com.github.javaparser.ast.`type`.{ ClassOrInterfaceType, PrimitiveType, Type, VoidType, WildcardType }
 import com.github.javaparser.ast.body._
 import com.github.javaparser.ast.expr._
 import com.github.javaparser.ast.stmt._
-import com.twilio.guardrail.generators.{ Response, ScalaParameter, ScalaParameters }
+import com.twilio.guardrail.generators.{ Response, ScalaParameters }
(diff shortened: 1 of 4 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
```diff
@@ -463,7 +455,7 @@ object DropwizardServerGenerator {
                   )
                 )
 
-                val futureResponseType = completionStageType(responseType)
+                val futureResponseType = completionStageType(responseType.clone().setTypeArguments(new WildcardType))
```
blast-hardcheese: So now to implement this, I have to have the `<?>` in my method types?

```diff
-    CompletionStage<AddPetResponse> addPet(final Pet body);
+    CompletionStage<AddPetResponse<?>> addPet(final Pet body);
```

kelnos: Unfortunately yes.  It's a little visually ugly, but in practical terms isn't a big deal.  Alternatively we could do:
```
<T> CompletionStage<AddPetResponse<T>> addPet(final Pet body);
```
... but not sure that's a big improvement.

blast-hardcheese: What's the motivation behind this change? The previous behaviour matches the Scala strategy, since `isInstanceOf` and sealed trait matching are roughly equivalent. The only thing this doesn't do is exhaustivity checking. Could we implement these members in terms of `enum` to get exclusivity matching, without exposing this `<?>` stuff to the user?

kelnos: The main motivation is to make the generated code more compact (and nicer to read), and also to avoid some copy/paste between different server generator implementations.  Java of course doesn't have the exhaustivity checking, so any bugs around emitting all the if/else branches would need specific tests.

It's also nicer from a testing perspective; you can write tests that don't need to enumerate every single response type when building or testing mocks.  (This is a benefit our users would get, not just a benefit for guardrail itself.)

PR #201, opened 2019-03-21
Generate @TRACE and @PATCH annotations for Dropwizard

kelnos:
Turns out you're not supposed to use `@HttpMethod` on resource methods directly, but instead create your own annotations that are themselves annotated with `@HttpMethod`.

**Contributing to Twilio**


kelnos:
Yeah, DW/Jersey/JAX-RX doesn't support PATCH or TRACE natively.  I think version 2.1 or 3.0 of JAX-RS does, but DW/Jersey haven't moved to it yet.

PR #202, opened 2019-03-21
Move generated samples under target/ and refactor build.sbt a bit

kelnos:
This avoids a weird situation where a source root is a subdirectory of another source root, and makes it so the sample code generation gets cleaned up with `sbt clean`.  The gitignore rules become simpler, and IDEs don't get confused about where the generated code lives.

**Contributing to Twilio**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -47,35 +47,22 @@ val exampleCases: List[(java.io.File, String, Boolean, List[String])] = List(
   (sampleResource("pathological-parameters.yaml"), "pathological", false, List.empty)
 )
 
-val exampleJavaArgs: List[List[String]] = exampleCases
-  .foldLeft(List[List[String]](List("java")))({
-    case (acc, (path, prefix, tracing, extra)) =>
-      acc ++ (for {
-        kind <- List("client", "server")
-        frameworkPair <- List(
-          ("dropwizard", "dropwizard"),
-        )
-        (frameworkName, frameworkPackage) = frameworkPair
-        tracingFlag                       = if (tracing) Option("--tracing") else Option.empty[String]
-      } yield
-        (
-          List(s"--${kind}") ++
-            List("--specPath", path.toString()) ++
-            List("--outputPath", s"modules/sample-${frameworkPackage}/src/main/java/generated") ++
-            List("--packageName", s"${prefix}.${kind}.${frameworkPackage}") ++
-            List("--framework", frameworkName)
-        ) ++ tracingFlag ++ extra)
-  })
+val exampleFrameworkSuites = Map(
+  "scala" -> List(
+    ("akka-http", "akkaHttp", List("client", "server")),
+    ("endpoints", "endpoints", List("client")),
+    ("http4s", "http4s", List("client", "server"))
+  ),
+  "java" -> List(
+    ("dropwizard", "dropwizard", List("client", "server"))
+  )
+)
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -47,43 +47,30 @@ val exampleCases: List[(java.io.File, String, Boolean, List[String])] = List(
...
-      } yield
-        (
-          List(s"--${kind}") ++
-            List("--specPath", path.toString()) ++
-            List("--outputPath", s"modules/sample-${frameworkPackage}/src/main/java/generated") ++
-            List("--packageName", s"${prefix}.${kind}.${frameworkPackage}") ++
-            List("--framework", frameworkName)
-        ) ++ tracingFlag ++ extra)
-  })
+val exampleFrameworkSuites = Map(
```
blast-hardcheese: Why use a map? Why not just
```scala
object FrameworkSuites {
  val java = List(...)
  val scala = List(...)
}
```
?

build.sbt
```diff
@@ -128,7 +115,7 @@ addCommandAlias("example", "runtimeSuite")
 addCommandAlias("cli", "runMain com.twilio.guardrail.CLI")
 addCommandAlias("runtimeScalaSuite", "; resetSample ; runScalaExample ; " + scalaFrameworks.map(x => s"${x}Sample/test").mkString("; "))
 addCommandAlias("runtimeJavaSuite", "; resetSample ; runJavaExample ; " + javaFrameworks.map(x => s"${x}Sample/test").mkString("; "))
-addCommandAlias("runtimeSuite", "runtimeScalaSuite ; runtimeJavaSuite")
+addCommandAlias("runtimeSuite", "; runtimeScalaSuite ; runtimeJavaSuite")
```
blast-hardcheese: 😉

build.sbt
```diff
@@ -113,13 +100,13 @@ artifact in (Compile, assembly) := {
...
-val scalaFrameworks = List("akkaHttp", "endpoints", "http4s")
-val javaFrameworks = List("dropwizard")
+val scalaFrameworks = exampleFrameworkSuites("scala").map(_._2)
+val javaFrameworks = exampleFrameworkSuites("java").map(_._2)
 
 resetSample := {
   import scala.sys.process._
   (List("sample") ++ (scalaFrameworks ++ javaFrameworks).map(x => s"sample-${x}"))
-    .foreach(sampleName => s"git clean -fdx modules/${sampleName}/src modules/${sampleName}/target" !)
+    .foreach(sampleName => s"rm -rf modules/${sampleName}/target/generated" !)
```
blast-hardcheese: Still would prefer to stick to `git clean` here, as it will refuse to eat all your data if things go wrong. Operating with the bounds of the repo as the limit to potential damage is 💯

build.sbt
kelnos: Cuz then I'm passing a `List[(String, String, List[String])]` to `exampleArgs()`, which is fine, but then it wants a string to represent the language to pass as an argument, and... it's gone now.

build.sbt
kelnos: I like uniformity ;)

build.sbt
kelnos: Yeah, good call.

build.sbt
blast-hardcheese: Ah, that makes sense.

PR #203, opened 2019-03-22
Java code formatting

kelnos:
* Use [google-java-format](https://github.com/google/google-java-format) to format the generated Java code.  I don't love everything about Google's code style guidelines, but this output is _way_ better than javaparser's not-so-pretty printer.
* Sort class body declarations in a sane way so things look nice.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
@@ -2,7 +2,15 @@ package com.twilio.guardrail.generators.syntax
 
 import com.github.javaparser.JavaParser
 import com.github.javaparser.ast.`type`.{ ClassOrInterfaceType, Type }
-import com.github.javaparser.ast.body.{ ClassOrInterfaceDeclaration, Parameter }
+import com.github.javaparser.ast.body.{
+  BodyDeclaration,
+  ClassOrInterfaceDeclaration,
+  ConstructorDeclaration,
+  FieldDeclaration,
+  InitializerDeclaration,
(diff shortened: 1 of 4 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
```diff
@@ -177,6 +173,38 @@ object Java {
       }
   }
 
+  def sortDefinitions(defns: List[BodyDeclaration[_]]): List[BodyDeclaration[_]] = {
+    import com.github.javaparser.ast.Modifier._
+    def sortKeyFor(x: BodyDeclaration[_]): Int = x match {
+      case cd: ClassOrInterfaceDeclaration if cd.getModifiers.contains(PUBLIC)                              => 0
+      case cd: ClassOrInterfaceDeclaration if !cd.getModifiers.contains(PRIVATE)                            => 10
```
blast-hardcheese: `if (a) {} else if (!a) {} else {}`?

kelnos: This comment does make sense for a few of the cases below, but not this one.  Will clean up the other ones before merging.

kelnos: (Remember, the order in Java for access is public > protected > (no modifier) > private.)

blast-hardcheese: Honestly misread this to be `PUBLIC` `!PUBLIC`

PR #205, opened 2019-03-25
Unify build tool codepaths

blast-hardcheese:
- Combine `Target` and `CoreTarget`, at least in structure initially
- Unify the three different kinds of log handling
- Consolidate common build idioms from different build tooling into a set of reusable functions
- Using `NonEmptyList` to represent argument lists for different languages

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/Error.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/Error.scala
@@ -8,6 +8,8 @@ case class UnparseableArgument(name: String, message: String) extends Error
 case object NoArgsSpecified                                   extends Error
 case object NoFramework                                       extends Error
 case object PrintHelp                                         extends Error
+case class RuntimeFailure(message: String)                    extends Error
+case class UserError(message: String)                         extends Error
 object Error {
   case class ArgName(value: String) extends AnyVal
 }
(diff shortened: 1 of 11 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/Error.scala
```diff
@@ -8,6 +8,8 @@ case class UnparseableArgument(name: String, message: String) extends Error
 case object NoArgsSpecified                                   extends Error
 case object NoFramework                                       extends Error
 case object PrintHelp                                         extends Error
+case class RuntimeException(message: String)                  extends Error
```
kelnos: This naming is maybe kinda confusing, no?

modules/sample/src/main/resources/petstore.json
```diff
@@ -514,7 +514,7 @@
           {
             "name": "custom-optional-value",
             "in": "formData",
-            "type": "number",
+            "type": "integer",
```
kelnos: Heh, I noticed the runtime warning about this, thought it was intentional ;)

kelnos:
Odd build failure?

modules/codegen/src/main/scala/com/twilio/guardrail/Error.scala
blast-hardcheese: Yeah, maybe `RuntimeFailure` or something

modules/sample/src/main/resources/petstore.json
blast-hardcheese: lol, no; it never happened in Scala because there was an `x-scala-type`

blast-hardcheese:
bintray went down yesterday

PR #207, opened 2019-03-26
DW responses back in handler

kelnos:
Moves DW response definitions back into the handler interface, as inner classes.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/ServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/ServerGenerator.scala
@@ -66,7 +66,7 @@ object ServerGenerator {
             }
             (responseDefinitions, serverOperations) = responseServerPair.unzip
             renderedRoutes   <- generateRoutes(context.tracing, resourceName, basePath, serverOperations, protocolElems)
-            handlerSrc       <- renderHandler(handlerName, renderedRoutes.methodSigs, renderedRoutes.handlerDefinitions)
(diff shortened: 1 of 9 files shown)


blast-hardcheese:
The only question I've got is around whether these response classes belong in the Handler or the Resource. In scala, they're in the resource, as the `Handler` is a trait that the user has complete control over. Is there some advantage for doing things this way? If not, would you mind putting these in the Resource to avoid cluttering the handler?

kelnos:
My rationale here was that the response classes are used by the implementor of the handler interface, so it might make sense to keep them in the same place.  If the user goes and opens up the handler to look at the methods they need to implement, they also see the definitions for the responses right there.  In theory they don't even ever need to open up the resource classes unless they're curious.  But of course for people that use an IDE (probably 99% of Java developers), the source for the definitions is just one ctrl-click (or whatever) away, so maybe that's not a big deal.

I don't really feel strongly about this either way, so if you'd rather we put them in the resource class, I'm fine with that.  Just let me know.

blast-hardcheese:
Actually, considering this further (with some in-person discussion), it seems that the considerations at play here are:
- If the `Response` inner classes are part of the `Handler`, they can be tab-completed without contention, whereas `Resource` may have some other noise preventing more ergonomic consumption
- One reason `Response` lives in `Resource` in Scala's generators is because the object is passed _in_ to the handler methods by way of the first parameter list, instead of created in `Handler` and returned to `Resource`.
- Another reason `Response` lives in `Resource` is because Scala doesn't have static members to traits; putting these definitions into the trait would create path-dependent classes, and splitting the trait into a companion object would increase noise around the `Handler` definition

For these reasons, I retract my initial concerns. I think what you've got here is solid. Thanks for the back and forth!

PR #213, opened 2019-03-27
Make GetFrameworkDefinitions aware if tracing is enabled or not

kelnos:
**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/Common.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/Common.scala
@@ -119,7 +119,7 @@ object Common {
       frameworkImports   <- getFrameworkImports(context.tracing)
       frameworkImplicits <- getFrameworkImplicits()
       frameworkImplicitName = frameworkImplicits.map(_._1)
-      frameworkDefinitions <- getFrameworkDefinitions()
+      frameworkDefinitions <- getFrameworkDefinitions(context.tracing)
 
       files <- (clients.flatTraverse(writeClient(pkgPath, pkgName, customImports, frameworkImplicitName, dtoComponents, _)),
                 servers.flatTraverse(writeServer(pkgPath, pkgName, customImports, frameworkImplicitName, dtoComponents, _))).mapN(_ ++ _)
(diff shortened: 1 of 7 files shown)


blast-hardcheese:
This is a huge issue -- If multiple clients or servers are generated with the same package with _different_ `tracing` values, this will cause inscrutable compilation failures -- this is why tracing is only possible in non-shared structures.

kelnos:
For this particular use case, I want to generate a `RequestContext` interface that's guaranteed to be the same for client and server.  The idea is that guardrail's DW server generation emits code that builds a concrete `RequestContext` instance on every request that gets passed to the handler, and then the handler can pass that instance to client calls where tracing is enabled.  For this _particular_ use case, there's no conflict: if tracing is enabled in one (but not both) of the client or server generations, the `RequestContext` interface ends up there, but there's no harm to the generated code that doesn't use or know about it.

I do get that, if someone wasn't careful, they could write some framework generation code that ends up clobbering things if the tracing setting isn't enabled consistently.  Which sucks.  But also consider that there's nothing preventing that happening from the client and server generators: you can easily write code that emits classes of the same name with different contents in the client and server generators that ends up clobbering each other (and this is independent of the `tracing` setting; you can totally do this even if tracing is always disabled or always enabled).  The only solution _there_ is "just don't do that", so is it actually worse to pass through the tracing setting to the framework as well?  At least in this case there's only one place to keep things consistent between the client and server.

The other option I see is writing a `generateRequestContextIface()` method in a helper class, and then calling it from both the client and server generators when tracing is enabled (or not if it's not enabled).  The class will get clobbered, but it'll always get clobbered with the same contents, so it doesn't matter.  I'm ok with that if you're really against this.

blast-hardcheese:
The issue I'm trying to work around is the case of ordering.

If two clients or servers are configured in a build.sbt or pom.xml, the first without tracing and the second with, everything compiles fine, as you say. Swap them, and it doesn't compile anymore.

Relying on the order of generation is a level of complexity that I don't want to subject users to. The error is inscrutable (it would appear that guardrail's generation is broken, leading to spurious tickets), and even if the user is interested in figuring it out themselves, it would take a significant number of steps back to see where the actual issue is.

kelnos:
Not sure what you mean.  The directory doesn't get cleaned out between runs.  The order doesn't matter, at least for the specific case I'm dealing with.

kelnos:
It's already possible, though.  I can generate two clients in the same directory, one with tracing on, and one with tracing off.  If the client generator is not written with this in mind, it could generate the same support class with different contents depending on the tracing setting, breaking in exactly the same way.  This adds the ability to do the same with frameworks support classes, sure, but it's not materially different.

PR #217, opened 2019-03-29
Java misc fixes

kelnos:
Nothing major here, just fixing some type parameter constraints and moving some code around so it's more accessible to extensions.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
@@ -91,10 +91,13 @@ object Java {
   def safeParseRawImport(s: String): Target[ImportDeclaration]       = safeParse("safeParseRawImport")(JavaParser.parseImport, s"import ${s};")
   def safeParseRawStaticImport(s: String): Target[ImportDeclaration] = safeParse("safeParseStaticImport")(JavaParser.parseImport, s"import static ${s};")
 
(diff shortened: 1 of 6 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
```diff
@@ -65,6 +66,10 @@ object Java {
     def toNodeList: NodeList[T] = new NodeList[T](l: _*)
   }
 
+  implicit class RichNodeList[T <: Node](val nl: NodeList[T]) extends AnyVal {
+    def toList(implicit cls: ClassTag[T]): List[T] = List(nl.iterator.asScala.toArray: _*)
```
blast-hardcheese: `toList` instead of `toArray`?

kelnos: Is that better?  My feeling is that since varargs is (under the hood) just an array, why convert to List just so the compiler/JVM can convert to Array?

blast-hardcheese:
```scala
    def toList(implicit cls: ClassTag[T]): List[T] = List(nl.iterator.asScala.toArray: _*)
```
==
```scala
    def toList(implicit cls: ClassTag[T]): List[T] = nl.iterator.asScala.toList
```
no?

kelnos: wow.  did not see that in the autocomplete list.  i's dumb.

PR #224, opened 2019-04-02
Use cats effect Async instead of Effect

hanny24:
Both http4s generated server and client require implicit instance of `Effect[F]`. This seems as way too powerful constraint that limits certain use cases. One example would be `type F[T] = ReaderT[IO, Context, T]`. It does not have instance of `Effect`. If we were to change the requirement from `Effect` to `Async`, it would remove this limitation.

**Contributing to Twilio**


hanny24:
cc @tomasherman

hanny24:
Yes, existing codebases continue to work just fine as every `Effect` is also `Async`.

blast-hardcheese:
Very cool, thanks for the clarification.

PR #226, opened 2019-04-03
Load Java class definitions from resources in a safer manner

kelnos:
Makes the source-file-in-resources loading generic (so I can use it easily from other places), and safer (wrap with `Try` and return `Target` instead), and returns a `SupportDefinition` so we can pull import lists from the source file itself.

**Contributing to Twilio**


modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
```diff
@@ -214,13 +215,26 @@ object Java {
...
+                SupportDefinition[JavaLanguage](
+                  new Name(className),
+                  cu.getImports.toList,
+                  clsDef
+                )
+            )
+          )
+    }
+
+  lazy val SHOWER_SUPPORT_DEF: Target[SupportDefinition[JavaLanguage]] = loadSupportDefinitionFromResources("Shower")
```
blast-hardcheese: Is there any reason to have a singleton here? Why not `def`?

kelnos: It's intended to be used as an immutable object.  No reason aside from that.

blast-hardcheese: understood

kelnos:
@blast-hardcheese any objections to merging?

PR #232, opened 2019-04-10
Go back to non-parameterized resource classes for DW servers

kelnos:
Arguably the old way made the generated code a little nicer to read, but the aesthetics for the user were ugly with the type param, and especially the wildcard type param in the handler signatures.

This effectively reverts acd9f592478867c8ac54337c8deaa06d9ce55dfa

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
@@ -124,41 +124,30 @@ object DropwizardServerGenerator {
   }
 
   def generateResponseSuperClass(name: String): Target[ClassOrInterfaceDeclaration] = {
-    val genericTypeT   = JavaParser.parseClassOrInterfaceType("T")
-    val entityBodyType = optionalType(genericTypeT)
-
-    val cls = new ClassOrInterfaceDeclaration(util.EnumSet.of(PUBLIC, ABSTRACT), false, name)
-      .addTypeParameter("T")
+    val cls = new ClassOrInterfaceDeclaration(util.EnumSet.of(ABSTRACT), false, name)
     cls.addField(PrimitiveType.intType, "statusCode", PRIVATE, FINAL)
-    cls.addField(entityBodyType, "entityBody", PRIVATE, FINAL)
 
     cls
-      .addConstructor(PROTECTED)
+      .addConstructor()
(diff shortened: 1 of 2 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
```diff
@@ -353,11 +355,38 @@ object DropwizardServerGenerator {
                 val responseName = s"${handlerName}.${operationId.capitalize}Response"
                 val responseType = JavaParser.parseClassOrInterfaceType(responseName)
 
+                val entitySetterIfTree = NonEmptyList
+                  .fromList(responses.value.collect({
+                    case Response(statusCodeName, Some(_)) => statusCodeName
+                  }))
+                  .map(_.reverse.foldLeft[IfStmt](null)({
```
kelnos: Before you scream at me for using `null` as a `foldLeft` initializer, let me explain.  This actually does make sense and is weirdly elegant.  The constructor signature for `IfStmt` is
```
public IfStmt(Expression condition, Statement thenStmt, Statement elseStmt)
```
So for the last item in the list, `null` has to be passed for `elseStmt`.  In the old old old similar code, I mapped the response-with-entity list into a `List[IfStmt]` and passed `null` for `elseStmt` each time, and then did a side-effect-ful `reduceLeft` to just call `prevStmt.setElse(curStmt)` to mutate the existing statements.  It worked, but I considered that ugly, as well as requiring an extra pass over the list.

This feels more elegant to me: reverse the list of stuff, pass `null` as the initializer, and then pass that as `elseStmt` in the lambda body.  One iteration over the list, and no side-effect-ful mutation later.

blast-hardcheese: I definitely agree with you, especially considering the use of `NonEmptyList`. I wish that were able to be captured in the types, but all I've got is
```scala
def elif(elseClause: IfStmt, statusCodeName: String): IfStmt = ???
...map(_.reverse match { case NonEmptyList(last, rest) => rest.foldLeft(elif(null, last))(elif _)
```
which requires another definition. The only danger of what you have here is if `NonEmptyList` is relaxed, we leak null

kelnos: Relaxed in the sense that they allow a NEL to be empty??

Issue #222, opened 2019-03-31
Case-to-Case inehritance generated


tomasherman:
Hey,

i came across this case of invalid scala code generated - basically if you have a definition which contains allOf with just one other definition, case class extension hierarchy is generated.

The swagger.json:

```json
{
  "swagger" : "2.0",
  "info" : {
    "title" : "someapp",
    "description" : "someapp",
    "version" : "1"
  },
  "basePath" : "/v1",
  "schemes" : [ "http", "https" ],
  "produces" : [ "application/json" ],
  "paths": {},
  "definitions": {
    "PairedDeviceInfo" : {
      "description" : "Data about a device's pairing.",
      "allOf" : [ {
        "$ref" : "#/definitions/PairedDeviceFields"
      } ]
    },
    "PairedDeviceFields" : {
      "description" : "Fields describing the pairing of a managed / paired device.",
      "type" : "object",
      "properties" : {
        "status" : {
          "type": "integer"
        }
      },
      "required" : [ "status", "deviceState" ]
    }
  }
}
```

scala code: 
```scala
// PairedDeviceFields.scala
case class PairedDeviceFields(status: BigInt)
object PairedDeviceFields {
  implicit val encodePairedDeviceFields = {
    val readOnlyKeys = Set[String]()
    Encoder.forProduct1("status") { (o: PairedDeviceFields) => o.status }.mapJsonObject(_.filterKeys(key => !(readOnlyKeys contains key)))
  }
  implicit val decodePairedDeviceFields = Decoder.forProduct1("status")(PairedDeviceFields.apply _)
}

// PairedDeviceInfo.scala
case class PairedDeviceInfo(status: BigInt) extends PairedDeviceFields
object PairedDeviceInfo {
  implicit val encodePairedDeviceInfo = {
    val readOnlyKeys = Set[String]()
    Encoder.forProduct1("status") { (o: PairedDeviceInfo) => o.status }.mapJsonObject(_.filterKeys(key => !(readOnlyKeys contains key)))
  }
  implicit val decodePairedDeviceInfo = Decoder.forProduct1("status")(PairedDeviceInfo.apply _)
}
```

I would expect either a type alias or some traits to be generated ...

replicable example can be found here: https://github.com/tomasherman/guardrail-playground

tomasherman:
@blast-hardcheese just tagging you to make sure this doesn't get lost ;)

blast-hardcheese:
@tomasherman Unnecessary, fyi

[...]

PR #234, opened 2019-04-12
issue #222

tomasherman:
attempt at #222


blast-hardcheese:
So, in this case, I'd add an explicit test, similar to `modules/codegen/src/test/scala/core/issues/Issue126.scala`, that encapsulates what you hope to see out of your change. Diffing the output of current master against this branch yields the following:
```diff
 package issues.issue222.server.http4s.definitions
 import io.circe._
 import io.circe.syntax._
 import io.circe.generic.semiauto._
-case class Request(state: Option[BigInt] = None, id: Option[String] = None) extends RequestFields
+case class Request(id: Option[String] = None)
 object Request {
   implicit val encodeRequest = {
     val readOnlyKeys = Set[String]()
-    Encoder.forProduct2("state", "id") { (o: Request) => (o.state, o.id) }.mapJsonObject(_.filterKeys(key => !(readOnlyKeys contains key)))
+    Encoder.forProduct1("id") { (o: Request) => o.id }.mapJsonObject(_.filterKeys(key => !(readOnlyKeys contains key)))
   }
-  implicit val decodeRequest = Decoder.forProduct2("state", "id")(Request.apply _)
+  implicit val decodeRequest = Decoder.forProduct1("id")(Request.apply _)
 }
```
which I assume is not your goal. I presume your intent would be to preserve all the keys, just to get rid of the object inheritance, which your PR does partially, just needing a bit more massaging to get to where it needs to be.

Some considerations:
- What happens if there are multiple `$ref`s?
- What happens if there are multiple object literals?
- What happens if the `allOf` contains an object literal that itself has an `allOf`?

Not all of these need to be solved right now, I'm very much in favor of accomplishing primarily what is necessary to unblock consumers, but without consideration, it's very easy to implement something that will not be forward source-compatible.

Again, sorry for having blocked you; I'm interested in helping unblock this feature request, so feel free to ask any follow-up questions here.

tomasherman:
this is complex :( I added the test and you are correct that behaviour is not correct ... i will keep trying :/

tomasherman:
so i made some progress, but dropwizzard is still failing

tomasherman:
WOOOO! I can't believe it I made the tests pass 🎉

modules/codegen/src/test/scala/core/issues/Issue222.scala
```diff
@@ -0,0 +1,91 @@
...
+      """
+
+    compare(requestFields.tpe, expectedFieldsTpe)
+    compare(requestFields.cls, expectedFieldsCls)
+    compare(fieldsEncoder, expectedFieldsEncoder)
+    compare(fieldsDecoder, expectedFieldsDecoder) 
+  }
+  
+  private def compare(t1: Tree, t2: Tree): Assertion = {
+    println(s"$t1 | $t2")
```
blast-hardcheese: No `println` please

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
```diff
@@ -399,7 +400,9 @@ object ProtocolGenerator {
                 parents <- extractParents(comp, definitions, concreteTypes)
                 model   <- fromModel(clsName, comp, parents, concreteTypes)
                 alias   <- modelTypeAlias(clsName, comp)
-              } yield model.getOrElse(alias)
+              } yield {
+                model.getOrElse(alias)
```
blast-hardcheese: Unnecessary change

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
```diff
@@ -246,7 +246,7 @@ object ProtocolGenerator {
           val isRequired = requiredFields.contains(name)
           SwaggerUtil.propMeta[L, F](prop).flatMap(transformProperty(clsName, needCamelSnakeConversion, concreteTypes)(name, prop, _, isRequired))
       })
-      defn <- renderDTOClass(clsName, params, parents)
+      defn <- renderDTOClass(clsName, params, parents) //check for discriminator, see issue222
```
blast-hardcheese: Unnecessary comment -- The test infrastructure prevents regression, trying to work around the comments is undesirable.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -171,11 +171,13 @@ object CirceProtocolGenerator {
...
         val discriminators = parents.flatMap(_.discriminators)
-        val parenOpt       = parents.headOption
+        val parentOpt      = parents.headOption
         val terms = (parents.reverse.flatMap(_.params.map(_.term)) ++ selfParams.map(_.term)).filterNot(
           param => discriminators.contains(param.name.value)
         )
-        val code = parenOpt
+
+        val code = parentOpt
+          .filter(p => parents.exists(s => s.discriminators.nonEmpty)) // part of issue222
```
blast-hardcheese: This is not the right place to do this filter -- `parentOpt` should not even be able to be constructed. See the suggestion above.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -171,11 +171,13 @@ object CirceProtocolGenerator {
 
       case RenderDTOClass(clsName, selfParams, parents) =>
         val discriminators = parents.flatMap(_.discriminators)
-        val parenOpt       = parents.headOption
+        val parentOpt      = parents.headOption
```
```suggestion
        val parentOpt      = if (parents.exists(s => s.discriminators.nonEmpty)) { parents.headOption } else { None }
```

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -56,9 +56,11 @@ object JacksonGenerator {
     parentOpt.foreach({ parent =>
       val directParent = JavaParser.parseClassOrInterfaceType(parent.clsName)
       val otherParents = parent.interfaces.map(JavaParser.parseClassOrInterfaceType)
-      cls.setExtendedTypes(
-        new NodeList((directParent +: otherParents): _*)
-      )
+      if (parent.discriminators.nonEmpty) {
```
blast-hardcheese: One caveat here is that this only checks to see if the discriminators are `nonEmpty` on the _first_ parent, hence `parentOpt`, `val parentOpt = parents.headOption`.

This deviates from the behaviour in the Circe generator, which checks _all_ `parents.exists(s => s.discriminators.nonEmpty)`

Recommendation is to consider the implications of this difference, then pick the one that will benefit users the most in the future.

modules/codegen/src/test/scala/core/issues/Issue43.scala
```diff
@@ -444,7 +444,7 @@ class Issue43 extends FunSpec with Matchers with SwaggerSpecRunner {
           :: ADT(namePet, tpePet, trtPet, staticDefnsPet) :: ADT(nameCat, tpeCat, trtCat, staticDefnsCat) :: Nil,
         _,
         _,
-        _
+        _3
```
blast-hardcheese: Unnecessary change(?)

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
tomasherman: the idea was to add some messages for future generations trying to understand the code

tomasherman:
I think i addressed all questions / comments for this PR ... are there any changes needed before this can be merged?

modules/codegen/src/test/scala/core/issues/Issue222.scala
```diff
@@ -0,0 +1,90 @@
...
+    
+    val expectedFieldsTpe = t"""RequestFields"""
+    val expectedFieldsCls = q"""case class RequestFields(state: Option[BigInt] = None)"""
+    
+    val List(fieldsEncoder, fieldsDecoder) = requestFields.staticDefns.definitions
+
+    val expectedFieldsEncoder =  q"""
+         implicit val encodeRequestFields = {
+           val readOnlyKeys = Set[String]()
+           Encoder.forProduct1("state")((o: RequestFields) => o.state).mapJsonObject(_.filterKeys(key => !(readOnlyKeys contains key)))
```
blast-hardcheese: Isn't `"id"` missing from the encoder?

modules/sample/src/main/resources/issues/issue222.yaml
```diff
@@ -0,0 +1,26 @@
...
+paths: {}
+definitions:
+  Request:
+    description: Request fields with id
+    allOf:
+      - "$ref": "#/definitions/RequestFields"
+      - type: object
+        properties:
+          id:
+            type: string
```
blast-hardcheese: ### This doesn't render both `id` and `id2`
```suggestion
            type: string
      - type: object
        properties:
          id2:
            type: string
```
### This doesn't render parameters from _any_ object literals, so `id` is missing
```suggestion
            type: string
      - "$ref": "#/definitions/RequestFields2"
```

modules/codegen/src/test/scala/core/issues/Issue222.scala
tomasherman: what do you mean? there is no `id` in request fields, it's only in request itself ... unless im missing something?

blast-hardcheese: Ah, my apologies -- I encountered the error I mentioned below, and I was hasty in looking for an example of the undesired behaviour in what you already had here.

tomasherman:
Right ... i will try to fix the issues you found tomorrow ... thanks for
the heads up

tomasherman:
ok so this one was a doozy ... anyways, i added two more test cases, hopefully fixing your concerns ...

I fixed it but im not really sure why the original logic was the way it was (basically taking just last object from composed schema ... perheps bug and not intentional?

I also added some error handling - guardrail will now fail with error message in case some composed object references definition that does not exist (i wasted like 30 minutes debugging this case when fixing one of those tests i added so i added this validation/error reporting)

Let me know what you think!

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -104,9 +104,12 @@ object CirceProtocolGenerator {
         (swagger match {
           case m: ObjectSchema => Target.pure(Option(m.getProperties))
           case comp: ComposedSchema =>
-            Target.pure(Option(comp.getAllOf()).toList.flatMap(_.asScala.toList).lastOption.flatMap(prop => Option(prop.getProperties)))
+            val extractedProps =
+              Option(comp.getAllOf()).toList.flatMap(_.asScala.toList).map(e => Option(e.getProperties).map(_.asScala.toMap)).collect { case Some(e) => e }
```
```suggestion
              Option(comp.getAllOf()).toList.flatMap(_.asScala.toList).flatMap(e => Option(e.getProperties).map(_.asScala.toMap))
```

tomasherman:
@kelnos Brain, i know this is not your issue but the changes you merged yesterday broke this PR...

You seem like the guy that's mostly working on jackson stuff, how do you feel issue #222 should be compiled for Java? Basically it's this hierarchy: https://github.com/twilio/guardrail/blob/20ad1f605006e24f2a5a42c075e672b6b81408cf/modules/sample/src/main/resources/issues/issue222.yaml

kelnos:
@tomasherman I know OO is somewhat controversial, but my opinion here is to implement these as class hierarchies and not flatten them out, for Java at least.

A lack of discriminator just means that the spec author doesn't need polymorphic deserialization, but that doesn't say anything about how the user wants to use the generated classes.  Perhaps they've created a hierarchy intentionally and intend to exploit use of the base class so they can operate on all the classes generically?  I'd rather not take that possibility away.  For people who don't care, the interface to create and interact with the objects is the same either way, so there's no loss to usability by maintaining the hierarchy.

I think this is less an issue in Scala for some hand-wavy ill-defined reasons around programming philosophy, but I'm pretty happy with the behavior I've arrive at with Java/Jackson.

tomasherman:
Fair enough, however the generated code for the scenario i submitted above won't compile

The problem is that both Request and RequestFields have static method builder() ... and in Request the types don't match (you can see the output of travis job) ... do you have an opinion how to solve this?

kelnos:
@tomasherman sorry, just a sec -- speaking offline with @blast-hardcheese & realizing I was missing context around why your change is being made.  Agree that we need to flatten things out in the case where the swagger spec specifies multiple inheritance.

Unfortunately it's not super simple.  Reverting my Jackson PR probably won't help, as the old code was garbage at dealing with polymorphism, so I'd expect your new test cases to fail with that code as well.  I'm going to attempt to get the Jackson code generating flat objects for the mutiple-inheritance case over the next couple hours.  If I can get it done, we'll merge that PR plus yours, and then cut a release.  Otherwise we'll cut a release with what's in master now, and I'll continue working so we can get these changes out in the next release.

Really sorry that I hadn't kept tabs on your work well enough to realize we'd be affecting each other and should have been coordinating.

kelnos:
Ok, above PR referenced should make Jackson behave properly.  I'm going to try to merge our branches together to see if we actually end up with something that works for everything.

kelnos:
So my test spec doesn't actually build for scala with your branch merged in.  Not sure if it's worth fixing, but for reference:
```
swagger: 2.0
info:
  version: 1.0.0
paths: {}
definitions:
  A:
    type: object
    properties:
      a:
        type: string
  B:
    allOf:
      - $ref: "#/components/schemas/A"
      - type: object
        properties:
          b:
            type: string
  C:
    type: object
    properties:
      c:
        type: string
  D:
    type: object
    properties:
      d:
        type: string
  E:
    allOf:
      - $ref: "#/components/schemas/C"
      - $ref: "#/components/schemas/D"
      - type: object
        properties:
          e:
            type: string
  F:
    type: object
    discriminator: f
    required:
      - f
    properties:
      f:
        type: string
  G:
    type: object
    properties:
      g:
        type: string
  H:
    allOf:
      - $ref: "#/components/schemas/F"
      - $ref: "#/components/schemas/G"
      - type: object
        properties:
          h:
            type: string
```
And the errors I'm getting:
```
[error] /home/brian/src/twilio/guardrail/modules/sample-akkaHttp/target/generated/polymorphismMultipleInheritance/client/akkaHttp/definitions/H.scala:9:81: class G needs to be a trait to be mixed in
[error] case class H(g: Option[String] = None, h: Option[String] = None) extends F with G
[error]                                                                                 ^
[error] /home/brian/src/twilio/guardrail/modules/sample-akkaHttp/target/generated/polymorphismMultipleInheritance/client/akkaHttp/definitions/H.scala:9:12: case class H has case ancestor polymorphismMultipleInheritance.client.akkaHttp.definitions.G, but case-to-case inheritance is prohibited. To overcome this limitation, use extractors to pattern match on non-leaf nodes.
[error] case class H(g: Option[String] = None, h: Option[String] = None) extends F with G
[error]            ^
[error] /home/brian/src/twilio/guardrail/modules/sample-akkaHttp/target/generated/polymorphismMultipleInheritance/server/akkaHttp/definitions/H.scala:9:81: class G needs to be a trait to be mixed in
[error] case class H(g: Option[String] = None, h: Option[String] = None) extends F with G
[error]                                                                                 ^
[error] /home/brian/src/twilio/guardrail/modules/sample-akkaHttp/target/generated/polymorphismMultipleInheritance/server/akkaHttp/definitions/H.scala:9:12: case class H has case ancestor polymorphismMultipleInheritance.server.akkaHttp.definitions.G, but case-to-case inheritance is prohibited. To overcome this limitation, use extractors to pattern match on non-leaf nodes.
[error] case class H(g: Option[String] = None, h: Option[String] = None) extends F with G
[error]            ^
```

tomasherman:
Thanks for the effort! I think it's definitely worth fixing ... i will take look at this over the weekend and try to implement this properly, but if im not able to, can we agree to merge what i've done here? It's strictly better than what is curently implemented in guardrail and it will allow us (avast) to use it in our usecases and we can contribute further improvements along in the future

blast-hardcheese:
Merging this now, further bugfixes for https://github.com/twilio/guardrail/pull/234#issuecomment-488861156 can be in future releases. Better to have the new functionality and iterate on what we have for now.

Again, thank you for your diligent work on this PR!

tomasherman:
Sweet! Thank you guys! Or next contribution will probably be attempt at
validations

PR #235, opened 2019-04-12
DW/Jersey java8 time support

kelnos:
Adds a Jersey binder to allow using the java8/jsr310 datetime classes (`OffsetDateTime`, `LocalDate`, etc.) in DW server method parameters.

Unfortunately this requires the user to add code to their server initialization to register the binder.  Will add docs around that when I write the full Java docs.

Also fixes a small bug where if an operation was defined to consume multipart/form-data, but there were no form parameters of type `File`, the server generator would incorrectly use the `@FormParam` annotation for parameters rather than `@FormDataParam`.

**Contributing to Twilio**


andrewryno:
Not sure if I'm missing something, but isn't this covered in `dropwizard-jersey`? https://github.com/dropwizard/dropwizard/tree/master/dropwizard-jersey/src/main/java/io/dropwizard/jersey/jsr310

PR #243, opened 2019-04-17
Adding support for too many arguments

### LINKED ISSUE #45: GET method with more than 22 parameters
Probably this is not such a frequent case (or, let me phrase it differently: I *hope* it isn't), but I have a project with an API definition where one of the routes has 25 parameters. In this case, the generated code does not compile.

For the record, see [this conversation](https://discuss.lightbend.com/t/akka-http-get-method-with-more-than-22-parameters/585) with a workaround.

blast-hardcheese:
Resolves #45 

Finally got around to it, @manuelbernhardt!

Important notes
===

### Parse order
This changes the parse order of akka-http directives. Previously, we attempted to parse the entity before parsing the headers. Now, we parse everything we get in the base HTTP message before starting to parse the entity. This permits route failover before parsing the entity.

### Restrictions
~The only caveat is that this does not support more than 22 parameters in a single group (path, query string, headers, form fields). This may need to change, but the MVP here is to just split the individual groups; in the future, we'll need to divide the individual matchers into 22 parameter groups, but this should at least give a little bit of breathing room.~ Now it does!

### http4s
Turns out http4s also had a 22-parameter limit for the union between form field and header parameters, as they were added together then matched as a giant tuple. Now, the tuple is chunked into 21 parameter chunks, chaining the next tuple on the 22nd element.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
@@ -603,7 +603,7 @@ object SwaggerUtil {
             throw new UnsupportedOperationException
         )
 
-    def generateUrlAkkaPathExtractors(path: String, pathArgs: List[ScalaParameter[ScalaLanguage]]): Target[Term] = {
+    def generateUrlAkkaPathExtractors(path: String, pathArgs: List[ScalaParameter[ScalaLanguage]]): Target[(Term, List[Term.Name])] = {
       import akkaExtractor._
       for {
         partsQS <- runParse(path, pathArgs)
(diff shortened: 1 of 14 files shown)


manuelbernhardt:
@blast-hardcheese awesome, thanks! :)

modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
```diff
@@ -603,28 +603,34 @@ object SwaggerUtil {
...
           }
-        } else directive
+        } else allPairs
         result = queryParams.fold(trailingSlashed) { qs =>
-          q"${trailingSlashed} & ${qs}"
+          val NonEmptyList((directives, bindings), xs) = trailingSlashed
+          NonEmptyList((q"${directives} & ${qs}", bindings), xs)
         }
-      } yield result
+      } yield result.reverse
```
blast-hardcheese: I'm not super excited about having `reverse` in a bunch of different places in this PR, but `NonEmptyList`'s `unapply` is left-biased, so building it up backwards, then reversing at the end seems to be the most clean way to implement it.

Issue #249, opened 2019-04-25
in:query and type:array results in "missing parameter type"


blast-hardcheese:
After #243, it seems as though repeated optional query parameters don't compile anymore due to `ParamMagnet` confusion.

```diff
diff --git a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
index 193cb8733..875626f55 100644
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
@@ -618,10 +618,10 @@ object AkkaHttpServerGenerator {
                   params match {
                     case List() =>
                       next =>
-                        acc(Term.Apply(directive, List(next)))
+                        acc(Term.Apply(Term.Select(directive, Term.Name("apply")), List(next)))
                     case xs =>
                       next =>
-                        acc(Term.Apply(directive, List(Term.Function(xs.map(x => Term.Param(List.empty, x, None, None)), next))))
+                        acc(Term.Apply(Term.Select(directive, Term.Name("apply")), List(Term.Function(xs.map(x => Term.Param(List.empty, x, None, None)), next))))
                   }
               }))
           val methodMatcher  = bindParams(List((akkaMethod, List.empty)))
```

Suggested change:
```scala
get(
  path("foo")(
//  parameter(Symbol("bar").as[String].*).map(Option.apply _).apply( bar => // This works
    parameter(Symbol("bar").as[String].*).map(Option.apply _)( bar =>       // This causes "missing parameter type"
      discardEntity(
        complete(
          handler.getBaz(getBazResponse)(bar)))))))
```

PR #250, opened 2019-04-25
Disambiguate akka-http when optional parameter lists are specified

blast-hardcheese:
Resolves: #249 

- Resolves syntactic ambiguity for akka-http parameter matching
- Additionally, for optional `type: array` parameters, adds a `filter(_.nonEmpty)`, removing the possibility of `Some(Nil)`
- Also fixing optional list parameter matching compiler error in http4s that was exposed when adding the tests for akka-http.

**Contributing to Twilio**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,6 +41,7 @@ val exampleCases: List[(java.io.File, String, Boolean, List[String])] = List(
   (sampleResource("issues/issue164.yaml"), "issues.issue164", false, List.empty),
   (sampleResource("issues/issue215.yaml"), "issues.issue215", false, List.empty),
   (sampleResource("issues/issue218.yaml"), "issues.issue218", false, List.empty),
+  (sampleResource("issues/issue249.yaml"), "issues.issue249", false, List.empty),
   (sampleResource("multipart-form-data.yaml"), "multipartFormData", false, List.empty),
   (sampleResource("petstore.json"), "examples", false, List("--import", "support.PositiveLong")),
   (sampleResource("plain.json"), "tests.dtos", false, List.empty),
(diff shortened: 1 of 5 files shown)


kelnos:
Is there no better way to do this? All the `.apply` makes the generated code harder to read and pretty ugly, IMO :(

blast-hardcheese:
I'm working against https://github.com/scalameta/scalameta/issues/1634, unfortunately.
```scala
path("foo") {
  complete(200)
}
```
gets translated into
```scala
Term.Apply(
  Term.Apply(Term.Name("path"), Lit.String("foo")),
  Term.Block(List(Term.Apply(Term.Name("complete"), Lit.Int(200))))
)
```
when generated, but
```scala
Term.Apply(
  Term.Apply(Term.Name("path"), Lit.String("foo")),
  Term.Apply(Term.Name("complete"), Lit.Int(200))
)
```
when parsed. I tried a bunch of different ways to avoid this, but was ultimately unsuccessful. Now, however, scalac treats `parameter(Symbol("bar").as[String].*).map(Option.apply _)( bar => ... )` in such a way that we lose type information on the `Option.apply _`.

An alternate strategy here could be fixing `${tpe}` in `Option.apply[???] _` which runs the risk of the types slipping between those two points (for instsance if `${tpe}` isn't the base type, but the wrapped type, we'd end up with `tpe = Option[List[Long]]`, `Option[$tpe]`, ending up with `Option[Option[List[Long]]]`).

Let me try that instead.

blast-hardcheese:
@kelnos OK, I was able to constrain it to _just_ the offending
```scala
parameter(Symbol(${arg}).as[${tpe}].*).map(xs => Option[Iterable[${tpe}]](xs)).apply
```

Issue #242, opened 2019-04-17
Add alternate constructor for Jackson POJO Builder class that uses instance of the POJO as a template


kelnos:
So basically:
```java
public class Foo {
    public class Builder {
        public Builder(/* ... list of required properties ... */) {
            // existing constructor
        }

        public Builder(final Foo template) {
            // initialize builder fields with template's fields
        }

        // ...
    }

    // ...
}
```

PR #253, opened 2019-04-26
Add a copy constructor for Jackson POJOs

kelnos:
Depends on https://github.com/twilio/guardrail/pull/252
Addresses https://github.com/twilio/guardrail/issues/242

**Contributing to Twilio**


blast-hardcheese:
Why doesn't `builder()` automatically include all the properties defined in the DTO by default? `.builder().withFoo(123).build()` to change one property, for instance? It does the required ones, but not the optional ones, which I think is somewhat counter-intuitive

kelnos:
`.builder()` is a static method that doesn't have an instance to copy from.

I agree that, functionally, having a `.builder()` instance method that pre-fills a new builder with existing state makes sense, however, the pattern I've implemented is the one that's common and Java devs will be familiar with.

Issue #247, opened 2019-04-24
In Java builders, initialize optional List<> properties/parameters without Optional


kelnos:
Currently it looks like we do (in POJO and client call builders)
```java
private Optional<List<String>> foos = Optional.empty();

public Builder withFoos(final List<String> foos) {
    this.foos = Optional.of(foos);
}
```
Really we should just do
```java
private List<String> foos = Collections.emptyList();

public Builder withFoos(final List<String> foos) {
    this.foos = requireNonNull(foos);
}
```

blast-hardcheese:
This is to enforce required parameters in the builders? 👍

kelnos:
This ends up being a Jackson issue (though arguably it's behaving correctly/consistently).  If you have an `Optional<List<>>` property in a POJO that is `Optional.empty()`, you end up with a `null` for that property in the serialized JSON, which is probably never what you want.  The better approach is to just have a 'bare' `List<>` and initialize it to empty, which will end up as `[]` in the JSON.

From the perspective of the builder, the behavior and interface is the same.

gisripa:
`.objectMapper(new ObjectMapper().setSerializationInclusion(JsonInclude.Include.NON_ABSENT))` this excluded Optional<SomeObject> values, guess it might work for List too.

PR #257, opened 2019-04-26
Don't wrap optional List properties in Optional

kelnos:
There's no need to do this, and it results in `null` for the property in the JSON rather than an empty list, which is usually not desired.

Addresses https://github.com/twilio/guardrail/issues/247

**Contributing to Twilio**


blast-hardcheese:
> usually not desired.
we actually had this problem already, where an assumption like this forced folks to start representing their APIs as `Option[Option[A]]` to get around "present but empty" vs "not present".

I'd avoid this, as something like a `PUT` with an optional array field will always be filled, potentially resetting the value depending on the semantics of the backend.

kelnos:
Yeah, that's a good point.  Jackson does have support for "leave out this property entirely if it's null/absent", so that might be a reasonable compromise.

Issue #225, opened 2019-04-03
Monitoring per operation


hanny24:
This issue might be a more of a question rather than issue. I'm interested in a way how http4s generated code handles monitoring. Right now we are able to monitor things globally, ie. we can keep track of HTTP status codes returned by a server, but I'm not able to keep a track of status codes per endpoints/operationId.  Similar thing applies http4s server. How do you solve this? Or do you event solve it at all?

I have an idea how to modify current generator in order to suit our needs. Essentially it requires to be able to pass `String => Request[F] => F[Response[F]]` into every Client and Resource (server).

blast-hardcheese:
This seems very close to `TraceBuilder`, https://github.com/twilio/guardrail/blob/320afebc2c26f800de7699bd009b314d4ed4c6be/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sGenerator.scala#L53

Is what you're trying to do possible with that?

@cponomaryov Do you have any thoughts around this?

hanny24:
`TraceBuilder` has two issues from my point of view:
- it works only for client side monitoring, not a server one
- it forces us to propagate `TraceBuilder` across whole application, which is something I would very much like to avoid.

hanny24:
Let's try other way. How do you guys monitor http4s Responses (latency, success & failure rate)? From what I see it is only possible to monitor it globally, i.e not on single endpoint granularity.

I have an idea how to possibly modify/extend currently generated code, but I would like to do it only when it's absolutely needed. It might be a case I missed something.

blast-hardcheese:
We currently use akka-http, so I don't know how this would work in http4s.

Our monitoring is effectively global, as you've described, by sanitizing the response URLs and emitting metrics from outside the application.

Looking again at `String => Request[F] => F[Response[F]]`, how would you implement this? This function, if passed in to the `Resource`, wouldn't have access to creating a `Response[F]`, would it?

Given that, wouldn't it be closer to `String => Request[F] => F[Response[F]] => F[Response[F]]` or `String => Request[F] => Response[F] => Response[F]`?

I could see the first being implemented as something like a response transformer, giving you complete access to the value coming from the response of whatever `Handler` method was called, where the second would be trying to inject into the router (at least the way akka-http does it, this could be different for http4s)

I agree with the direction you're going here, I didn't give enough consideration before when asking about `TraceBuilder`; it's necessary for us, due to needing to inject context headers into downstream client requests, but I can see how others may not need that.

[...]

PR #259, opened 2019-04-27
Issue225

tomasherman:
Hey guys,

I took a stab at #225  ... what do you think? There might be some corners to polish but do you feel like the general strokes are correct?


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -501,8 +502,14 @@ object Http4sServerGenerator {
           case Nil        => responseInMatch
           case generators => q"for {..${generators :+ enumerator"response <- $responseInMatch"}} yield response"
         }
-        val routeBody       = entityProcessor.fold[Term](responseInMatchInFor)(_.apply(responseInMatchInFor))
-        val fullRoute: Case = p"case req @ $fullRouteWithTracingMatcher => $routeBody"
+        val routeBody = entityProcessor.fold[Term](responseInMatchInFor)(_.apply(responseInMatchInFor))
+        val fullRoute: Case =
+          p"""case req @ $fullRouteWithTracingMatcher => 
+             val response = {
```
blast-hardcheese: This is really good -- The one consideration is that we are trying to reduce top-level bindings inside of an individual route. The way things are written currently, parameters named `response` will now conflict.

tomasherman: good point, i can skip the `val response` and just add the computation into the tuple directly, i did it this way to make the code nicer but it's not required to be this way

tomasherman:
i changed the implementation, pls take a look ... Also i'm wondering if we should make some sort of case class that describes all possible arguments to class that does the pattern matching, as the handlerWrapper and tracing signature is quite big and kinda confusing (a lot of function literals) ... maybe if we had some:
```scala
case class HandlerArguments(
  handlerWrapper: (String, Request[F], F[Response[F]]) => Response[F],
  tracer: ...
)
```
and than just pass that to the class, it would make for cleaner code ... what do you think @blast-hardcheese ?

blast-hardcheese:
I hesitate to do it unless there just get to be way too many parameters. Two or three parameters isn't enough motivation for me right now, unless you're trying to solve a particular problem?

tomasherman:
i'm not ... with that said, i don't have nothing more to add to this PR unless you have any other comments. So feel free to merge this if it makes sense to you

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -77,10 +77,13 @@ object Http4sServerGenerator {
       case GetExtraRouteParams(tracing) =>
         for {
           _ <- Target.log.debug("Http4sServerGenerator", "server")(s"getExtraRouteParams(${tracing})")
-          res <- if (tracing) {
-            Target.pure(List(param"""trace: String => Request[F] => TraceBuilder[F]"""))
-          } else Target.pure(List.empty)
-        } yield res
+          handlerWrapper = param"""handlerWrapper: (String, Request[F], F[Response[F]]) => F[Response[F]] = (_: String, _: Request[F], r: F[Response[F]]) => r"""
```
blast-hardcheese: `handlerWrapper` reads as though it's wrapping the entire `Handler`, where as we're just wrapping an individual route or operation. Maybe `mapRoute`?

It's also somewhat undesirable that the user can now arbitrarily transform the response, potentially violating the specification.

blast-hardcheese: Additionally, do we need to `(String, Request[F], F[Response[F]]) => F[Response[F]]`, or can we just `(String, Request[F], F[Response[F]]) => F[Unit]` and just `<*` them?

Do you have an example of how this will be used?

Once we're done setting this up here, I'll port the interface we settle on here into the akka-http interface, and @kelnos may be interested in it for Java, though I seem to remember that Dropwizard is more based on registering mutable transformers on the routing core.

tomasherman: i like the `mapRoute` name, lets use that

well one of the use cases we thought about was for example when error occurs, we might be able to provide some better error messages in the body (include operation method etc) but we don't have a stronger argument for it. If you feel F[Unit] is better, we can live with it

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -501,8 +504,11 @@ object Http4sServerGenerator {
           case Nil        => responseInMatch
           case generators => q"for {..${generators :+ enumerator"response <- $responseInMatch"}} yield response"
         }
-        val routeBody       = entityProcessor.fold[Term](responseInMatchInFor)(_.apply(responseInMatchInFor))
-        val fullRoute: Case = p"case req @ $fullRouteWithTracingMatcher => $routeBody"
+        val routeBody = entityProcessor.fold[Term](responseInMatchInFor)(_.apply(responseInMatchInFor))
+        val fullRoute: Case =
+          p"""case req @ $fullRouteWithTracingMatcher => 
+             handlerWrapper($operationId, req, {$routeBody})
```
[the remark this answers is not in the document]
tomasherman: The main reason for wrapping it in { } is that `routeBody` is a block of code, i need to somehow delimit it (unless i'm missing something) ... the other alternative was:

```scala
          p"""case req @ $fullRouteWithTracingMatcher => 
             val routeBody = () = {
               $routeBody
             }
             handlerWrapper($operationId, req, routeBody())
```

But this will cause problems if parameter `routeBody` is in the path and doesn't really add anything.

Let me know if you prefer the version with lambda so that i can change it, but i think the block is better

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
tomasherman: Pls let me know if you'd prefer the `F[Unit]` so that i can change the PR if needed. I think returning `F[Response[F]]` gives more flexibility but i don't have strong feelings either way

tomasherman: If we decide to go with `F[Unit]`, we can rename the method to something like `observeRoute`

blast-hardcheese: Let's go with `F[Unit]` for now, until we definitely know that we need the more flexibility. `F[Unit]` => `F[Response[F]]` is a fairly straightforward change, adding functionality. It would be much more difficult to go from `F[Response[F]]` to `F[Unit]` once folks start depending on that behaviour.

blast-hardcheese:
One last note, please update `MIGRATING.md` with migration notes for this change, as it will require syntax changes for folks that are using `tracing`.

I'm also debating on default parameters here, vs overloaded constructors.
- Default values require no multiple constructor juggling, and can be auto-curried
- Overloaded constructors can permit both `apply(tracing)`, `apply(observeRoute)`, and `apply(observeRoute, tracing)`, making migration to the new style a deprecation warning instead of "incompatible types"

I'm also debating on whether `tracing` should come before `observeRoute` or after.
- After makes sense, as every generated resource will have this parameter
- Before makes sense, `observeRoute` has a default value, whereas `traceBuilder` does not

tomasherman:
@blast-hardcheese I don't disagree. The reason why i chose to put tracing before routeObserve is to make source code backward compatible. I don't have strong feelings either way so i decided to just choose working solution that is backward compatible.

If you reach different conclusion, i am willing to change it (altho adding multiple constructors will be more difficult, but probably doable). Let me know what you want me to do

tomasherman:
when we make up our mind about it, i will fill in the migration guide ... also we will nneed something like this for client side also but we can do that as separate issue, server side is more important

tomasherman:
one other consideration - how would you feel about introducing a trait for observeRoute? The method signature is quite long and it might make sense to add trait or at least type alias for it, perhaps into companion object of generated class?

tomasherman:
Also, how do you feel about putting route body into 
```scala
val response = { $routeBody}
```

and then generate the val name `response` in such a way so that it doesn't conflict with path params? I think we will need to do that in order to easily apply the observeRoute, because you need to pass it into the method and also eval it and return it

tomasherman:
So i thought about this again :)

if we decide to go with this:
```
observe: (String, Request[F], F[Response[F]] => F[Unit]
```

than there is no way to run it, because 

```scala
val response: F[Response] = ???
observe(str, req, response) <* response
```
will run response twice

so we either need to return `F[Response[F]]` from the observe or change the signature to:
`observe: (String, Request[F], F[Response[F]] => F[Unit]`
and run it like:

```scala
val response: F[Response] = ???
response.flatMap(r => observe(str, req, r).map(_ => r))
```
which is ok, but a bit limiting, because it for example won't let you measure duration of the response processing ... so i would suggest to move back, again, to returning F[Reponse[F]] from the obesrve method ... thoughts?

blast-hardcheese:
Yeah... not `flatMap`ping would permit more analysis. I guess the best thing would be to go with the types of the middleware from http4s, to greatly increase code reuse.

Adapting existing middleware could be as simple as `middleware.contramap[(String, Request[F], F[Response[F]])](_._3)` if they don't need additional functionality.

OK. Rename to `mapRoute` and I'm onboard.

tomasherman:
So i gave a try to implementing multiple constructors and it's not trivial ... i would suggest to move this into a different issue as it will require some refactoring across the board ... for example i would suggest adding Constructor into language abstraction and than use those constructs, instead of extraRouteParameters

But i don't really have the time to go into it right now and i think the solution with default value for mapRoute is good enough ... what do you think?

tomasherman:
ok so this is my latest attempt ... i agree that multiple constructors would be better solution but i think it warrants bigger refactoring that is not in scope of this PR ... i would be really happy if we could merge this without that. It can come later, perhaps. 

With that said, i chose to use the default value in order to generated code backward compatible

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -501,8 +504,17 @@ object Http4sServerGenerator {
           case Nil        => responseInMatch
           case generators => q"for {..${generators :+ enumerator"response <- $responseInMatch"}} yield response"
         }
-        val routeBody       = entityProcessor.fold[Term](responseInMatchInFor)(_.apply(responseInMatchInFor))
-        val fullRoute: Case = p"case req @ $fullRouteWithTracingMatcher => $routeBody"
+        val routeBody    = entityProcessor.fold[Term](responseInMatchInFor)(_.apply(responseInMatchInFor))
+        val responseName = generateSafeRouteValName("response", pathArgs)
```
blast-hardcheese: This still seems like unnecessary complexity. If this was avoiding allocations or something, I'd understand it, but this is just for readability of the generated code?

Having arbitrary integer suffixes that change based on name collisions seems only slightly better, not enough to warrant the additional complexity.

That being said, I understand you've put a lot of effort into this PR at this time. I don't anticipate this causing any failures, though I expect it will eventually be unified to not require an additional binding.

modules/codegen/src/test/scala/core/issues/Issue225.scala
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    compare(genHandler, handler)
+
+    // Cause structure is slightly different but source code is the same the value converted to string and then parsed
+    compare(genResource.toString().parse[Stat].get, resource)
+  }
+  
+  private def compare(actual: Tree, expected: Tree): Unit = {
+    println(s"actual: ${actual.syntax}")
```
blast-hardcheese: No `println`s during normal operation please

tomasherman: sorry, leftover from debugging :)

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
tomasherman: well i added this functionality when we were going to do `F[Unit]` ... with `F[Response[F]]` I can remove this

blast-hardcheese: Ah, I understand -- `<*` with real bindings definitely would make it _much_ easier to read. Since we've settled on `mapRoute`, I would appreciate simplifying this section.

No objections beyond that though.

tomasherman:
ok i believe this one is ready :)

blast-hardcheese:
Thanks!

Issue #260, opened 2019-04-29
http4s server generates multiple companions for date type path params


kkreuning:
When I generate a http4s server with `date` types for path params I end up with multiple `object LocalDateVar`s in the generated `Resources` class. This breaks compilation because
```
LocalDateVar is already defined as object LocalDateVar
```
I expect that only one such object is created even if it's used in multiple paths.

PR #261, opened 2019-04-29
Fixed issue 260

kkreuning:
Fix for issue: https://github.com/twilio/guardrail/issues/260

Using date as path param type in multiple paths will no longer generate multiple LocalDateVar objects.

This might also apply to other 'non native' path param types.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
@@ -60,7 +60,7 @@ object Http4sServerGenerator {
             List(combinedRouteTerms),
             List.empty,
             methodSigs,
-            renderedRoutes.flatMap(_.supportDefinitions),
+            renderedRoutes.flatMap(_.supportDefinitions).groupBy(_.structure).map(_._2.head).toList, // Only unique supportDefinitions by structure
             renderedRoutes.flatMap(_.handlerDefinitions)
           )
         }
--- a/modules/codegen/src/test/scala/core/issues/Issue260.scala
+++ b/modules/codegen/src/test/scala/core/issues/Issue260.scala
@@ -1,0 +1,92 @@
+package core.issues
+
+import com.twilio.guardrail.generators.Http4s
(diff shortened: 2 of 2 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -60,7 +60,7 @@ object Http4sServerGenerator {
             List(combinedRouteTerms),
             List.empty,
             methodSigs,
-            renderedRoutes.flatMap(_.supportDefinitions),
+            renderedRoutes.flatMap(_.supportDefinitions).groupBy(_.structure).map(_._2.head).toList, // Only unique supportDefinitions by structure
```
blast-hardcheese: This could still cause name collisions, as the bodies could be different. Thoughts around prefixing support definitions by their associated operationId?

kkreuning: @blast-hardcheese That would be preferable indeed,

it's easy to prefix the generated `*Var` with the `operationId` in `com.twilio.guardrail.generators.Http4sServerGenerator.ServerTermInterp#generatePathParamExtractors`, however

I can't find a neat way to prefix the endpoint matcher (`case req @ GET ...`) with `operationId`, AFAIK the path matchers are generated in `com.twilio.guardrail.SwaggerUtil.paths#generateUrlHttp4sPathExtractors`, but I don't have access to the RouteMeta in there and it doesn't feel right to pass the route meta object all the way from `com.twilio.guardrail.generators.Http4sServerGenerator.ServerTermInterp#generateRoute` into the Swagger utils. Any ideas on how to handle this?

blast-hardcheese: Iirc when I was reviewing yesterday, it may be a decent strategy to pass a common prefix into the support generator method, bubbling up what other parameters use to a single definition, then propagating out.

blast-hardcheese:
This may come back to bite us in the future, but let's proceed with what you have here. The context and tests around this are sufficient for now.

PR #262, opened 2019-04-29
Fix NPEs in SwaggerGenerator schema type extraction

kelnos:
Also beefs up the logging around a few missing things so spec authors can find the issue easier.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/SwaggerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/SwaggerGenerator.scala
@@ -1,17 +1,21 @@
 package com.twilio.guardrail
 package generators
 
-import _root_.io.swagger.v3.oas.models._
 import cats.implicits._
-import cats.syntax.either._
 import cats.~>
 import com.twilio.guardrail.extract.ScalaPackage
 import com.twilio.guardrail.languages.LA
-import com.twilio.guardrail.shims._
 import com.twilio.guardrail.terms._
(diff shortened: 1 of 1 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/SwaggerGenerator.scala
```diff
@@ -56,31 +60,31 @@ object SwaggerGenerator {
         } yield className
 
       case GetParameterName(parameter) =>
-        Target.fromOption(Option(parameter.getName()), "Parameter missing \"name\"")
+        Target.fromOption(Option(parameter.getName()), s"Parameter missing 'name': ${parameter}")
```
blast-hardcheese: swagger-parser's models log each entry with `null` fields and whatnot. I don't know if `parameter`s are `Schema[_]`s, but https://github.com/twilio/guardrail/blob/d6f2dbb74d37cd8374b0c9304bde8f4e4a6bc272/modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerms.scala#L58 may be useful here

kelnos: `Parameter`s are not `Schema[_]`s, as it turns out.

kelnos: (And they only have 7 fields, so I guess it's no big deal if we log nulls.)

Issue #264, opened 2019-04-29
Dashed-case enum names with numeric components don't get converted properly


kelnos:
```
type: string
enum:
    - foo-bar-5
    - baz-quux-6
```
gives us names like `FooBar-5` and `BazQuux-6`.  Scala manages to deal with it because it can backtick-escape the identifiers.  Java (well, JacksonGenerator) ends up doing the conversion again (it wants upper-snake-case), but for this particular case it's too late because ProtocolGenerator tries to call `pureTermName()` on the (incorrectly) generated hybrid PascalDashedCase name, and that's not a valid identifier in Java.

We really have a couple problems:
1. ProtocolGenerator always calls `.toPascalCase` on enum strings.  This is because that's what Scala always wants, and when we added Java support we never abstracted that, and I was lazy and cut corners and just do the upper-snake-conversion in JacksonGenerator (which is too late for this case).
2. `.toPascalCase` (and really all our case converters) just can't handle values that have groupings that start with a number.

PR #265, opened 2019-04-30
Fix enum naming

kelnos:
Addresses https://github.com/twilio/guardrail/issues/264

This also cleans up some deficiencies with our case converters, namely not handling spaces as separators in the original identifier, and not handling numbers as the initial character in a grouping when converting to pascal case.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
@@ -316,7 +316,7 @@ object JacksonGenerator {
                 ).mapN((_, _))
               )(Function.const(Target.pure((tpe, defaultValue))) _)
             (finalDeclType, finalDefaultValue) = _declDefaultPair
-            term <- safeParseParameter(s"final ${finalDeclType} ${argName.escapeReservedWord}")
+            term <- safeParseParameter(s"final ${finalDeclType} ${argName.escapeIdentifier}")
             dep = classDep.filterNot(_.value == clsName) // Filter out our own class name
           } yield ProtocolParameter[JavaLanguage](term, name, dep, readOnlyKey, emptyToNull, defaultValue)
         }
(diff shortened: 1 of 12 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -103,7 +103,7 @@ object JacksonGenerator {
           case (value, termName, _) =>
             new EnumConstantDeclaration(
               new NodeList(),
-              new SimpleName(termName.getIdentifier.toSnakeCase.toUpperCase(Locale.US)),
```
blast-hardcheese: This is in "Add FormatEnumName term to ScalaTerms", but it doesn't appear to be doing that, what's going on here?

modules/sample/src/main/resources/issues/issue264.yaml
```diff
@@ -0,0 +1,22 @@
...
+      responses:
+        204:
+          description: OK
+components:
+  schemas:
+    FooEnum:
+      type: string
+      enum:
+        - foo-bar-1
+        - baz-quux-2-flurb
```
blast-hardcheese: You mentioned "not handling numbers as the initial character in a grouping", but what if the numbers are the initial character in the entire enumeration?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
kelnos: Without this change it'll generate very wrong names with the addition of a (correct) `FormatEnumName` for Java.  In general I like to make each commit as correct as possible.

modules/sample/src/main/resources/issues/issue264.yaml
kelnos: Welp, regardless, it'll be an invalid name in Java at least.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
blast-hardcheese: Ah, due to the double snake case. Makes sense. Thanks!

Issue #266, opened 2019-04-30
Rename scala-specific vendor extensions that are language-agnostic or JVM-generic


kelnos:
A few of the vendor extensions have "scala" in the name but aren't really scala-specific.  We should do the following (but maintain back-compat and print a deprecation warning for the old names):

* `x-scala-tracing-label`: rename to `x-tracing-label`
* `x-scala-package`: rename to `x-jvm-package`
* `x-scala-empty-is-null`: at the very least, rename to `x-jvm-empty-is-null`, but it might even be generic enough for `x-empty-is-null`
* `x-scala-file-hash`: rename to `x-file-hash` (note that only the scala akka-http server generator supports this so far)

PR #268, opened 2019-04-30
Vendor extension rename

kelnos:
Addresses https://github.com/twilio/guardrail/issues/266

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/extract/package.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/extract/package.scala
@@ -1,19 +1,32 @@
 package com.twilio.guardrail
 
 package object extract {
-  def ScalaTracingLabel[F: VendorExtension.VendorExtensible](v: F): Option[String] =
-    VendorExtension(v).extract[String]("x-scala-tracing-label")
-  def ScalaPackage[F: VendorExtension.VendorExtensible](v: F): Option[String] =
-    VendorExtension(v).extract[String]("x-scala-package")
-  def ScalaType[F: VendorExtension.VendorExtensible](v: F): Option[String] =
-    VendorExtension(v).extract[String]("x-scala-type")
+  private def extractWithFallback[F: VendorExtension.VendorExtensible, T: Extractable](v: F, name: String, fallbackName: String): Option[T] = {
+    val ve = VendorExtension(v)
+    ve.extract[T](name)
+      .orElse({
+        val value = ve.extract[T](fallbackName)
+        if (value.isDefined) {
+          println(s"WARNING: Deprecated vendor extension '${fallbackName}'; please migrate to use '${name}' instead")
+        }
+        value
+      })
+  }
+
+  def TracingLabel[F: VendorExtension.VendorExtensible](v: F): Option[String] =
+    extractWithFallback[F, String](v, "x-tracing-label", "x-scala-tracing-label")
+
+  def JvmPackage[F: VendorExtension.VendorExtensible](v: F): Option[String] =
+    extractWithFallback[F, String](v, "x-jvm-package", "x-scala-package")
 
   def ServerRawResponse[F: VendorExtension.VendorExtensible](v: F): Option[Boolean] =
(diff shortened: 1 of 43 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/extract/package.scala
```diff
@@ -1,19 +1,32 @@
...
+        }
+        value
+      })
+  }
+
+  def TracingLabel[F: VendorExtension.VendorExtensible](v: F): Option[String] =
+    extractWithFallback[F, String](v, "x-tracing-label", "x-scala-tracing-label")
+
+  def JvmPackage[F: VendorExtension.VendorExtensible](v: F): Option[String] =
+    extractWithFallback[F, String](v, "x-jvm-package", "x-scala-package")
```
blast-hardcheese: Thoughts on having `x-scala-package`, `x-java-package`, and `x-jvm-package`? Different aesthetics for organization by language, perhaps? Alternately, diamond dependencies on one client generated for Scala and the same one generated for Java?

kelnos: Ehhhhhhhhh, does this really seem worth pre-emptively adding before someone asks for it?

blast-hardcheese: I wouldn't think twice about it other than "This is deprecated! Change to this other thing!" _later_ "We brought it back!"

kelnos: Yeah, true.  If you really want it, I'll add it; just annoying because it means adding another ScalaTerm a la `CustomTypePrefixes`

blast-hardcheese: I can't honestly say I'm comfortable merging without it. This might be an opportunity to generalize the previous one-off target-interpolating setting hierarchy?

kelnos: Sure.  Got a problem, though.  `JvmPackage` is used in `SwaggerGenerator` (in `GetClassName`), which is conceptually lower in the stack than `ScalaGenerator`/`JavaGenerator` (which is what provides the list of language-specific prefixes to use).

One possibility I can think of is to put a call to ScalaTerm's `customPrefixes()` in `Common.prepareDefinitions()` (which is the only place that calls `getClassName()`, and also already has a reference to `ScalaTerms`), and then add a param to `getClassName()` to pass those prefixes in, so it can do the vendor extension extraction using the prefix list.  (Looks like `ServerGenerator` also calls `getClassName()`, so that call site would have to be changed as well.)

What do you think?

kelnos: (Ended up pushing a couple more commits to do just that.  LMK if that seems reasonable.)

modules/codegen/src/main/scala/com/twilio/guardrail/extract/package.scala
```diff
@@ -1,22 +1,38 @@
...
 
 package object extract {
-  def ScalaTracingLabel[F: VendorExtension.VendorExtensible](v: F): Option[String] =
-    VendorExtension(v).extract[String]("x-scala-tracing-label")
-  def ScalaPackage[F: VendorExtension.VendorExtensible](v: F): Option[String] =
-    VendorExtension(v).extract[String]("x-scala-package")
-  def ScalaType[F: VendorExtension.VendorExtensible](v: F): Option[String] =
-    VendorExtension(v).extract[String]("x-scala-type")
+  private def extractFromNames[F: VendorExtension.VendorExtensible, T: Extractable](v: F, names: List[String]): Option[T] =
+    names.foldLeft(Option.empty[T])(
```
blast-hardcheese: This is actually `names.collectFirstSome(VendorExtension(v).extract[T](_))`, fwiw

kelnos: Ooh, TIL.  Will update and then merge.

Issue #153, opened 2019-01-07
dashed-case conversion doesn't work for mixed case


blast-hardcheese:
Defining the following parameter:
```yaml
        - in: header
          name: X-Auth-User
          type: integer
          format: int64
          required: true
```

yields the following argument:
```scala
`x-Auth-User`: Long
```

whereas altering the case to
```yaml
        - in: header
          name: x-auth-user
          type: integer
          format: int64
          required: true
```

generates
```scala
xAuthUser: Long
```

andrewryno:
I just ran into this as well. This causes Java to not even compile since it can't do a string literal for the name.
```
[ERROR] Error: Unable to parse 'x-Actor' to a com.github.javaparser.ast.expr.Name: Encountered unexpected token: "-" "-"
    at line 1, column 2.

Was expecting:

    <EOF>
```

blast-hardcheese:
I _think_ https://github.com/twilio/guardrail/issues/264 may have fixed this for Java

blast-hardcheese:
going to do a release later today; If you wouldn't mind running `cli.sh` with relevant arguments against master to verify the fix, that'd be appreciated!

andrewryno:
Doesn't look like it. I just modified `issue148.yaml` to use `X-Header`/`X-Optional-Header` instead of lowercase and `javaTestSuite` failed with the same error.

```
Error: Unable to parse 'x-Header' to a com.github.javaparser.ast.expr.Name: Encountered unexpected token: "-" "-"
    at line 1, column 2.

Was expecting:

    <EOF>
```

PR #272, opened 2019-05-01
Rewrite case converters and add tests

kelnos:
Completely new approach; see comments in the code for a description.

Addresses https://github.com/twilio/guardrail/issues/153 finally.

**Contributing to Twilio**

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -1,3 +1,28 @@
+Migrating to 0.49.0
+===================
+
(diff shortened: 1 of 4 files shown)


kelnos:
Looks like http4s's response generation is affected by this; it wants `OK`, but we're now generating `Ok`.  I've updated the test to correspond to the change, but recognize that this may not be desired behavior.

kelnos:
Updated MIGRATING

PR #275, opened 2019-05-02
Upgrade http4s to version 0.20.0

mayboroda:
Upgrade of http4s library to current stable version 0.20.0

**Contributing to Twilio**


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -217,7 +217,7 @@ object Http4sServerGenerator {
             Target.pure(
               Param(
                 None,
-                Some((q"urlForm.values.get(${arg.argName.toLit}).map(_.map(Json.fromString(_).as[$tpe]))", p"Some(Seq(Right(${Pat.Var(arg.paramName)})))")),
+                Some((q"urlForm.values.get(${arg.argName.toLit}).flatMap(_.toList.headOption).map(Json.fromString(_).as[$tpe])", p"Some(Right(${Pat.Var(arg.paramName)}))")),
```
blast-hardcheese: Do we need to force `toList` here? If it's an iterable, can we not just `headOption` without first going to `List`?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -230,7 +230,7 @@ object Http4sServerGenerator {
                 None,
                 Some(
                   (
-                    q"urlForm.values.get(${arg.argName.toLit}).map(_.map(Json.fromString(_).as[$tpe]).sequence)",
+                    q"urlForm.values.get(${arg.argName.toLit}).flatMap(_.toList.headOption).map(Json.fromString(_).as[$tpe]).sequence",
```
blast-hardcheese:
```suggestion
                    q"urlForm.values.get(${arg.argName.toLit}).flatMap(_.toList.headOption).traverse(Json.fromString(_).as[$tpe])",
```

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -246,7 +246,7 @@ object Http4sServerGenerator {
                 None,
                 Some(
                   (
-                    q"urlForm.values.get(${arg.argName.toLit}).map(_.map(Json.fromString(_).as[$tpe]).sequence).sequence",
+                    q"urlForm.values.get(${arg.argName.toLit}).flatMap(_.toList.headOption).map(Json.fromString(_).as[$tpe]).sequence.sequence",
```
blast-hardcheese: Doesn't the `headOption` defeat the purpose of `multiOpt` here? Maybe I'm missing something?

If you resolve the formatting issues (which are done first to prevent unnecessarily long builds that end with formatting failure at the very end) it would be interesting to see if all the regression tests still pass.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
mayboroda: Agree. Question here would be is a `headOption` is enough? In prev. solution we had a `Option[Seq[Either]]` now it's just an `Option[Either]]`. Maybe it's better to have it like `Seq[Either]`?

blast-hardcheese: The semantics are different here, definitely, but the tradeoffs are this:

If there are multiple values supplied with the same name,
- 🚀... do we attempt to parse all of them and fail if any of them are invalid, and only if all of them are valid do we take the first one?
- 😄... do we take the first one and ignore all others, even if that one is invalid but others are valid?
- ❤️... do we take the _last_ one and ignore all others, even if that one is invalid but others are valid?
- 🎉... do we validate first, ignoring errors, then take the first/last?

Vote with emoji!

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
mayboroda: Should be fixed

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
mayboroda: Fixed

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
mayboroda: I've took a deep look at `directivesFromParams` that is used here and the first parameter assumes that we have one mandatory fields, so vote with 😄

mayboroda:
@blast-hardcheese I guess we resolved all the issue, may be we can merge this?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -233,7 +236,7 @@ object Http4sServerGenerator {
                 None,
                 Some(
                   (
-                    q"urlForm.values.get(${arg.argName.toLit}).map(_.map(Json.fromString(_).as[$tpe]).sequence)",
+                    q"urlForm.values.get(${arg.argName.toLit}).flatMap(_.toList).traverse(Json.fromString(_).as[$tpe])",
```
blast-hardcheese: I'm very wary of `.values.get("foo").flatMap(_.toList)`, I haven't had time to test it to be certain, but my intuition is that `urlForm.values.get(...)` would be `Map`, which would mean `get` would return `Option[_]`, which would mean `flatMap(_.toList)` would fail.

If you're available to hit this codepath or otherwise test this construct in the REPL, I'd appreciate it. I'd like to get this out, I'd just like to get it out without breaking functionality (as I expect you can appreciate).

Please pardon the delay around this, I should have been more communicative.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
kelnos: How about a fifth option?  If we are expecting a single value but receive multiple, fail immediately, because clearly the client is doing something wrong and will get unexpected behavior

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
mayboroda: Deal, I'll try to do my best to give to a REPL example and a fix if needed :)

mayboroda: @blast-hardcheese, instead of REPL I've added `multi` and `multiOpt` cases for `formData.yaml` and put more tests. So, I hope this will be a better proof than REPL ;)

modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
```diff
@@ -305,7 +305,7 @@ object AkkaHttpServerGenerator {
       directivesFromParams(
         arg => tpe => Target.pure(q"formField(Symbol(${arg}).as[${tpe}])"),
         arg => tpe => Target.pure(q"formField(Symbol(${arg}).as[${tpe}].*)"),
-        arg => tpe => Target.pure(q"formField(Symbol(${arg}).as[${tpe}].*).map(xs => Option(xs).filterNot(_.isEmpty)).apply"),
+        arg => tpe => Target.pure(q"formField(Symbol(${arg}).as[${tpe}].*).map(xs => Option(xs).filterNot(_.isEmpty))"),
```
mayboroda: I had to update akkaHttp for `multiOpt` version only. I hope this is correct. Have no experience with akkaHttp.

mayboroda:
@blast-hardcheese I've rebased on `upstream/master`, please check latest changes. I think I've covered all your doubts with tests according `multi formData`.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
mayboroda: @kelnos do you think it's possible to test? I mean we generate client and server and we assume having a client that accepts certain types. I guess it's not a valid case.

kelnos: Sure.  You wouldn't be able to use the autogenerated client to test (since it would of course conform to the correct behavior), but you could build `HttpRequest`s from scratch and send them to the server impl.

mayboroda: @kelnos please check my latest commits I think I've resolved all the issues.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaGenerator.scala
```diff
@@ -180,6 +180,7 @@ object ScalaGenerator {
                 implicit val showLocalDate = build[java.time.LocalDate](_.format(java.time.format.DateTimeFormatter.ISO_DATE))
                 implicit val showOffsetDateTime = build[java.time.OffsetDateTime](_.format(java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME))
                 implicit val showJavaURL = build[java.net.URI](_.toString)
+                implicit def showIterable[K](implicit s: Show[K]) = build[Iterable[K]](_.map(s.show).mkString(","))
```
blast-hardcheese: This is very undesirable, as it establishes a new, stringy protocol.

If I try to `Show[String]` of `List("foo", "bar,baz", "blix")`, then the consumer attempts to derive the members again, I'd be forced to `values.split(",")`, which would give me `Array("foo", "bar", "baz", "blix")`, which does not round-trip.

Why is this necessary? If it's for query string or form parameters, we should expand into successive single values, like `Foo=1&Foo=2&Foo=3`

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
blast-hardcheese: Thank you!

modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaGenerator.scala
mayboroda: This was done for header values. In case we have a multi-value header we apply this implicit. This is used only for akka-http as far as I understood

mayboroda:
> I think this may be getting too broad. I believe everything except `Show[Iterable[A]]` is great, but once we add that instance, I think it's no longer attempting to upgrade to 0.20.0, it's now trying to add new functionality.
> 
> Also, my apologies for the delay in response

Good to hear,
So, how I can help to fix this?

blast-hardcheese:
@mayboroda Let me take a look tomorrow. I've been at a conference last week, so I'm a bit behind, though I'd like to get this PR unblocked.

mayboroda:
Great,
I've just got back from vacation :) So, take your time I will wait.

blast-hardcheese:
OK, finally finished the fiddly aspects of #326. Taking a look at this now.

blast-hardcheese:
OK, there were issues in all Scala clients:

- `Option[_]` and `Iterable[_]` of _non_ `String` values would not always be passed through `Formatter.show`
- Endpoints client generator actually had a duplicate codepath entirely. Deduplicating that fixed the compiler errors for that client

While this does compile, I'll take a look with fresh eyes tomorrow before merging. I think we've finally gotten to the end of this process though, and with some significant improvements as well.

Thanks again for your patience!

PR #277, opened 2019-05-02
Flatten Jackson inheritance hierarchies where needed

kelnos:
This is to prevent us from generating uncompileable multiple-inheritance hierarchies.  It works like this:

1. If there's more than one parent class that has a discriminator, we throw an error.
2. If there's a single parent that has a discriminator, we extend from that parent class.  If there are other parents, we flatten them so the generated class has the properties of all those other parents.
3. If there's just a single parent, we extend from that parent class.
4. Otherwise there's no composition or polymorphism going on, so we just emit a class that extends from nothing.

Unfortunately this necessitated removing the static `builder()` method from the POJO class, because you can't have a static method in a subclass with the same arg list that has a different return type.  (Which is an odd restriction since Java doesn't support static method overriding, but perhaps they did it for future-proofing.)  To compensate, the builder constructors are now public.

I'm _not_ including a test spec with this PR as it causes the Scala build to fail, which should hopefully be fixed when https://github.com/twilio/guardrail/pull/234 lands.

Note that the vast majority of this PR is only whitespace change as the indentation level changed for a bunch of code.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
@@ -275,6 +275,348 @@ object JacksonGenerator {
     }
   }
 
+  private def renderDTOClass(clsName: String,
+                             selfParams: List[ProtocolParameter[JavaLanguage]],
+                             parents: List[SuperClass[JavaLanguage]]): Target[TypeDeclaration[_ <: TypeDeclaration[_]]] = {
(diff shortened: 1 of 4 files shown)


tomasherman:
FYI your build failed

kelnos:
@tomasherman yeah, the scala 2.11 compiler doesn't like it.  we're considering dropping 2.11 support; it's becoming a headache fighting against compiler bugs that will never get fixed.

blast-hardcheese:
#280

PR #281, opened 2019-05-04
Fix copypasta in JacksonGenerator causing maps to be lists

kelnos:
Also parse both in a safer way that makes it more clear where errors lie.

**Contributing to Twilio**


kelnos:
right?  i feel super dumb now.

PR #282, opened 2019-05-04
Preparing for enabling -Wall

blast-hardcheese:
I'm interested in getting ahead of code quality issues by ensuring that all warnings are treated as errors, as well as enabling select wartremover warts.

This PR contains some initial work towards that, by at least removing unused imports, as well as minor formatting.

**Contributing to Twilio**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -156,6 +156,7 @@ val codegenSettings = Seq(
   scalacOptions in ThisBuild ++= Seq(
     "-Ypartial-unification",
     "-Ydelambdafy:method",
+    "-Ywarn-unused-import",
     "-feature",
     "-unchecked",
     "-deprecation",
--- a/build.sbt
+++ b/build.sbt
@@ -156,7 +156,7 @@ val codegenSettings = Seq(
   scalacOptions in ThisBuild ++= Seq(
     "-Ypartial-unification",
     "-Ydelambdafy:method",
-    "-Ywarn-unused-import",
+    // "-Ywarn-unused-import",  // TODO: Enable this! https://github.com/twilio/guardrail/pull/282
     "-feature",
     "-unchecked",
     "-deprecation",
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/Args.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/Args.scala
@@ -1,7 +1,5 @@
 package com.twilio.guardrail
 
-import cats.syntax.either._
-
 case class Args(
     kind: CodegenTarget,
     specPath: Option[String],
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/CLI.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/CLI.scala
@@ -1,7 +1,6 @@
 package com.twilio.guardrail
 
 import java.nio.file.Path
-import cats.Applicative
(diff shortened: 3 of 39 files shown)


tomasherman:
How do you feel about cats.implicits._ ... it's my understanding that it might be better to use explicits imports as it makes compilation/ide support faster... i actually tried to do a PR with this by my cats knowledge left me hanging at some files. .. is that something you would like to be solved?

blast-hardcheese:
I'm on the fence about `cats.implicits._`. Initially I was very much against it, but over the years I've come to value the ease of having all instances available when I need them. Additionally, I find that it lowers the barrier to entry to folks so they don't need to remember `cats.instances.traverse._` or `cats.instances.foldable._` in order to do something.

I generally have the cats source open if I find that what I'm doing is too much boilerplate, evaluating whether someone else has already solved the same problem, but I guess some folks use IDEs to solve the same problem, by using completion or otherwise browsing available methods. As I prefer vim, I don't have a preference one way or another, as it's just another compiler error, easy enough to resolve and it happens fairly infrequently.

> my cats knowledge left me hanging at some files

What do you mean by this, you couldn't identify which implicits were necessary?

tomasherman:
> What do you mean by this, you couldn't identify which implicits were necessary?

Yes, i think it was this file: `modules/codegen/src/main/scala/com/twilio/guardrail/WriteTree.scala`

blast-hardcheese:
@tomasherman I took a look:
```
[error] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/WriteTree.scala:26:19: value traverse is not a member of Option[Int]
[error]           diffIdx.traverse { diffIdx =>
[error]                   ^
```
requires `cats.instances.option._` to adapt the stdlib to cats, as well as `cats.syntax.traverse._`

```
[error] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/WriteTree.scala:28:28: could not find implicit value for evidence parameter of type cats.Applicative[[V]cats.data.Writer[List[String],V]]
[error]           diffIdx.traverse { diffIdx =>
[error]                            ^
```
requires `cats.instances.list._`, [as there's only an instance for `Applicative` for `Writer` if `L` (`List[String]`) has an instance of `Monoid[_]`](https://github.com/typelevel/cats/blob/910df7807a890f0f2ba26a750ab51e1072bf37ee/core/src/main/scala/cats/data/WriterT.scala#L176)
```
[error] .../guardrail/modules/codegen/src/main/scala/com/twilio/guardrail/WriteTree.scala:47:27: value map is not a member of com.twilio.guardrail.WriteTree => cats.data.Writer[List[String],java.nio.file.Path]
[error]     unsafeWriteTreeLogged.map {
[error]                           ^
```
requires `cats.instances.function._` and `cats.syntax.functor._` as it's trying to `map` over a `Function1[A, B]`

blast-hardcheese:
I think this is an indicator that whatever performance gains we'd get are not worth the mental burden on potential contributors, unless compilation speed improved by orders of magnitude.

I'd love to see benchmarks, if you've got links 😀

tomasherman:
I don't have any, it's just a feeling that all-you-can-eat import slowers down intellij and scalac :) Thanks for taking the time to take a look at this!

blast-hardcheese:
Definitely onboard for revisiting this in the future, thanks for raising the concern!

Issue #255, opened 2019-04-26
Unknown type for the following structure (type: string, class: io.swagger.v3.oas.models.media.PasswordSchema)


tomasherman:
**Update** This issue is tracking errors generated by following schemas:

```yaml
swagger: "2.0"
info:
  title: Whatever
  version: 1.0.0
host: localhost:1234
schemes:
  - http
definitions:
  Foo:
    type: object
    properties:
      somePassword:
        type: string
        format: password
      someFile:
        type: file
      someBinary:
        type: string
        format: binary
```

They all generate errors similar to this one:
```
Error: Unknown type for the following structure (type: string, class: io.swagger.v3.oas.models.media.PasswordSchema):
  class PasswordSchema {
      class Schema {
      }
      type: string
      format: password
  }
```

----

Original issue:

Hi,

when using follwing definition:
```json
{"password":{"format":"password","type":"string"}}
```

I get the following error:
```
Error: Unknown type for the following structure (type: string, class: io.swagger.v3.oas.models.media.PasswordSchema):
  class PasswordSchema {
      class Schema {
      }
      type: string
      format: password
  }
```

Any idea what to do about it? I suspect it's a format that has been removed in swagger v3 or something like that

tomasherman:
Ok i think i figured it out for password (see #256 ) ... but i have same problem with `type: file`, which is trickier ... i'm not sure how to handle that one

tomasherman:
this is an example of the offending yml...
```yml
get:
  summary: get timezone data
  operationId: getTimezoneData
  description: |-
    Get timezone data (also called tzdata)
  responses:
    200:
      description: Succesfully retrieved tzdata.
      schema:
        description: User specific tzdata file.
        type: file
```

i assume raw response could be used for this but this fails the compilation ... i wonder how to solve this ... perhaps generate response that takes `fs2` stream and pipe it to raw response in the generated code?

kelnos:
What's the compiler error?  We've supported `type: file` for some time now; not sure why it wouldn't work.

blast-hardcheese:
@tomasherman akka-http has support for `type: file`, not sure that http4s has support for that yet. The naive implementation in akka-http uses a helper method to build a `java.io.File` object, which the file is streamed into.

This isn't a great solution though, I'd much rather go with something `InputStream` based to stream into a file or to anything else that can behave like an `InputStream`.

tomasherman:
Like i said, the error comes up during generaetion phase:
```
[...]

PR #283, opened 2019-05-05
#255 improvement - Fixed issue with type:file

tomasherman:
Fixes issue with `type: file` and also adds test for mentioned failing case, but it seems to be working

**Contributing to Twilio**


modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
```diff
@@ -251,9 +251,9 @@ object SwaggerUtil {
             case ("boolean", fmt)              => booleanType(fmt).map(log(fmt, _))
             case ("array", fmt)                => arrayType(fmt).map(log(fmt, _))
             case ("file", fmt) =>
-              fileType(fmt).map(log(fmt, _))
-            case ("binary", _) =>
-              fileType(None).map(log(None, _))
+              fileType(None).map(log(fmt, _))
```
blast-hardcheese: Why is this `None` required here? Logging `fmt` but passing `None` is misleading

modules/codegen/src/test/scala/core/issues/Issue255.scala
```diff
@@ -23,6 +23,11 @@ class Issue255 extends FunSuite with Matchers with SwaggerSpecRunner {
                            |      somePassword:
                            |        type: string
                            |        format: password
+                           |      someFile:
```
blast-hardcheese: These are described in the comments of #255, could you edit the summary in the issue to describe the full scope of the issues that are tracked in it for reference?

If this regression test starts failing due to `type: file` starting to break and they see "Unknown type for the following structure (type: string, class: io.swagger.v3.oas.models.media.PasswordSchema)" I suspect they would be confused.

modules/codegen/src/test/scala/core/issues/Issue255.scala
```diff
@@ -31,7 +36,14 @@ class Issue255 extends FunSuite with Matchers with SwaggerSpecRunner {
...
       ) = runSwaggerSpec(swagger)(Context.empty, Http4s)
-    
-    c1.structure shouldBe q"case class Foo(somePassword: Option[String] = None)".structure
+
+    val expected = q"case class Foo(somePassword: Option[String] = None, someFile: Option[java.io.File] = None, someBinary: Option[String] = None)"
+    compare(c1, expected)
+  }
+  
+  def compare(t: Tree, t2: Tree) = {
+    println(t)
```
blast-hardcheese: 😉

modules/codegen/src/test/scala/core/issues/Issue255.scala
tomasherman: will do

modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
tomasherman: binary has default format `binary` and then is rendered as type `binary`, which doesn't compile ... would you prefer to change signature if filetype to not take any format? I'm not sure what kind of format files can have anyway and i didn't manage to find any format definition

tomasherman:
I'm not really sure how `type: string` `format: binary` should be handled...i patched it up to work for the person asking in #255 but i think we should open different issue on how to handle binary data properly

modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
blast-hardcheese: well, this is unfortunate.

It turns out [`type: string`, `format: binary`](https://swagger.io/specification/#dataTypes) (as well as `type: string`, `format: byte`). I hope that the 2.x -> 3.x handles this mapping for us 🤔

tomasherman: yep, do you want me to add support for type: byte also?

blast-hardcheese: I think that's very much scope creep, attempting to deal with the Base64 encoding issue 😁

That being said, I'd appreciate verifying that both swagger 2.x and 3.x behave the same way here. As `binary` has been demoted to a `format` of `string`, this should definitely not be `case ("binary", None)`, it should be `case ("string", "binary")` and we should ensure that both 2.x and 3.x generate the same code (as I mentioned earlier).

tomasherman: well i was not gonna solve that issue, i was just gonna add code that will render `format` `byte` as file so that users are not completely screwed and can use it on their own

i don't understand what you mean by `behave the same way here` - do you mean that swagger 2.0 defintion with `type:file` generates the same code as 3.0 definition with `type: string format: binary`?

blast-hardcheese: Ah -- rendering `type: string; format: byte` is almost certainly to ensure that non-latin1 charactersets are not destroyed in transit. This is apparently a common technique in some non-english speaking countries. Treating them as `File` would definitely be inconvenient.

"behave the same way" meaning ensuring that they generate the same code, that is correct. I think having a regression test for the same `definition` in both OpenAPI 2.x and OpenAPI 3.x then comparing the `structure`s they produce would be overkill, considering that this is behaviour that should be handled in `swagger-parser-v2-converter`. Just verifying that it works at all would be sufficient for moving forward here (changing the `type: binary` to `type: string`, `format: binary`)

blast-hardcheese: Whoops, https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java only has one place that has the literal string `"binary"`, but that's when dealing with `type: file`.

It looks like we'll have to duplicate these patterns here.

blast-hardcheese: Please verify that this assertion is correct before working around it; I can open an issue (and possibly fix it) upstream.

tomasherman: Sorry, perhaps im a little dense, but ... i assume you are talking about swagger converter because guardrail uses it internally when reading swagger 2 files ... those get converterd to openapi 3 and then guardrail works with them, is that correct? Are you sure this is happening? Because when working with `type: file`, it was represented in as `FileSchema`, not as `string with binary format` which is what i would expect if the assumption that v2 sources are converted to v3 sources

tomasherman: hmm, it looks like swagger converts `type: file` to `binary string` if and only if it's as part of response: (https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java#L816)

blast-hardcheese: No need to apologise! You are correct. Internally, we use the `swagger-parser-v2-converter` in order to support both 2.x and 3.x.

Unfortunately, as that converter is bespoke and not generated via some specification of how to map 2.x to 3.x, my only reason to assume anything works is to [examine the source code in that project](https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java) to see if the functionality is there.

As for the file handling stuff you mentioned, it does appear that they explicitly handle the file type conversion. As I'm reading this converter source, it doesn't seem to do the same conversion for `"binary"`, which is why I'm asking that we verify before inadvertently removing functionality from the 2.x codebase :smile:

I would do it myself, but I'm unfortunately away from my computer until tomorrow

tomasherman: yep, when reverting changes in this PR, putting type file into response fails with 
```
[info]   class FileSchema {
[info]       class Schema {
[info]           type: string
[info]           format: binary
[info]       }
[info]   }
```

while putting them into definitions it fails with:

```
[info]   class FileSchema {
[info]       class Schema {
[info]           type: file
[info]           format: binary
[info]       }
[info]   }
```

tomasherman: so perheps we could split the concerns ... my main goals with this PR was to solve our issue with `type: file` in responses. I believe this pr solves those issues.

There is however unresolved issue with `type: string format:byte`, i would suggest fixing that one as different issue/pr

blast-hardcheese: I agree, `format: byte` is a different concern. Thanks for looking into this!

tomasherman: sweet ... i will open issue for binary improvements tomorrow ... anything else blocking this from being merged? Thanks for quick responses btw!

blast-hardcheese: I don't think anything else is blocking this from being merged, provided that both OpenAPI 2.x and 3.x behave the same way for _all_ changed cases in `typeName`.

tomasherman: well i think the `case "binary"` branch is dead and never gets called anyway (`type: binary` is not a valid in v2 or v3) and the `file` branch is only called from openapi v2 i think so it should be fine

blast-hardcheese: I think all of this may have been from a misunderstanding when I was reading the specification. I misread that `type: binary` was in the Swagger 2.0 specification.

Sorry for so much confusion! I appreciate you spelling it out to me 😁

tomasherman: no worries :) if i might be so bold, could you approve the PR then? ;)

blast-hardcheese:
Thanks again for the discussion!

Issue #280, opened 2019-05-03
Drop support for 2.11


blast-hardcheese:
Attempting to work around odd bugs in Scala 2.11's typesystem is not worth maintaining crossbuilds anymore.

If anyone is actively depending on 2.11 and can provide justification for continuing to release for it, please let us know.

Keep in mind, this is just for the code generator, not the generated code. The maven and sbt plugins use 2.12, but the generated code can be compiled as far back as 2.10. This question is only about build system integrations that do not support 2.12+

PR #286, opened 2019-05-06
Add scala 2.11.8 and 2.12.3 to .travis.yml

karolchmist:
I was trying to reproduce a problem of compiling a project using scala 2.11, so I've created a variation in travis. In the end it didn't help in my case, but maybe you'll find it useful to have this travis configuration in guardrail. 

But I'm not sure if there is any value added since guardrail is already cross compiled to both 2.11 and 2.12. 

**Contributing to Twilio**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
 language: scala
+scala:
+  - 2.11.8
+  - 2.12.3
 script:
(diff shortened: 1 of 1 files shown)


karolchmist:
I've just seen https://github.com/twilio/guardrail/issues/280, so I'm closing this PR

Issue #210, opened 2019-03-27
Never empty collectionFormat: multi


rtfpessoa:
When I have a parameter like:
```
parameters:
        - in: query
          name: fileIds
          description: File identifiers
          required: false
          type: array
          items:
            type: integer
            format: int64
          # HACK: Force type Long
          x-scala-type: 'Iterable[Long]'
          collectionFormat: multi
```

In the handler it generates something like `Option[Iterable[Long]]`, which is nice.

The problem I have is that even when nobody passes any fileIds I get a `Some(List())` instead of a `None`. Any ideas?

blast-hardcheese:
Seems as though https://github.com/twilio/guardrail/blob/6cae43c7b8c9fd274930c4f1b7ec0ae14cb75717/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala#L298 doesn't actually `.filter(_.nonEmpty)`.

If you (or others) are inclined to fix this, please make sure to apply the same patch to `http4s` server generation (if applicable!)

rtfpessoa:
Today I was trying to pick this but seems like it was fixed. The only issue I have is that the code does not compile:
```
[error] missing argument list for method apply in trait Function1
[error] Unapplied methods are only converted to functions when a function type is expected.
[error] You can make this conversion explicit by writing `apply _` or `apply(_)` instead of `apply`.
```

Did anyone use this or am I doing something wrong?

blast-hardcheese:
Confirmed that [`filterNot(_.isEmpty)`](https://github.com/blast-hardcheese/guardrail/blob/master/modules/scala-akka-http/src/main/scala/dev/guardrail/generators/scala/akkaHttp/AkkaHttpServerGenerator.scala#L464) is present, looks like this issue just fell off in the shuffle.

Thank you 🙂

Issue #184, opened 2019-02-01
Parameters of "type: string" reject numeric values as "malformed"


griffio:
💂‍♀️🚃We are sending query parameters that can be either a number or alpha-numeric input

In the latest version and earlier. The following simple fragment when generated as a route will only allow alpha-numeric values to be passed in the request.

For example: using a numeric value is rejected as "The query parameter 'a' was malformed: String"
 
```paths:
  /foo:
    get:
      operationId: getFoo
      x-scala-package: foo
      produces:
        - application/json
      parameters:
        - name: a
          in: query
          type: string      

🙉Hope that makes sense that we may expect the decoder to allow numerics to be converted to a string rather than be rejected.

griffio:
The parameter is turned into a 🔢JSNumber by the J.decodeJson(json) and sends it into a decodeString: Decoder[String] circe thing.
👁‍🗨https://github.com/twilio/guardrail/blob/c476988b5bf3024bde0fa91c73580b0c441842e1/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpGenerator.scala#L132

dsilvasc:
Is the advice here then to annotate every swagger string input parameter as `x-scala-type: Either[String, Long]` in case the input characters are all digits?

blast-hardcheese:
@dsilvasc I honestly haven't had a chance to circle around and fix the underlying bug here. I'd rather this not be the case, but for now I can't offer a better solution.

What needs to happen is the underlying types from swagger need to be reflected in akka-http's directives (`as[Long]` or equivalent) before being wrapped (currently all parameters are wrapped with circe) and decoded into domain-specific types (again, currently done via circe).

Right now, we're just taking everything as a string, attempting to parse it as JSON, then treating it as a string before proceeding. This creates an unfortunate situation where if a valid JSON symbol (numbers, boolean literals) are supplied, they'll get parsed as those literals, failing the string decoder. This has the additional unfortunate impact of removing surrounding double quotes from strings, as they're parsed as JSON string literals.

PR #287, opened 2019-05-08
fix: Remove extra apply in generated multiple query parameters

rtfpessoa:
Trying to fix codegen in #210 

When generating the akka code it does not compile. After removing it and compiling just the akka part it compiles in my tests. But seems like the tests in this repo fail. Any ideas?

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
@@ -713,7 +713,9 @@ object AkkaHttpServerGenerator {
                         acc(Term.Apply(directive, List(next)))
                     case xs =>
                       next =>
-                        acc(Term.Apply(directive, List(Term.Function(xs.map(x => Term.Param(List.empty, x, None, None)), next))))
+                        acc(
+                          Term.Apply(Term.Select(directive, Term.Name("apply")), List(Term.Function(xs.map(x => Term.Param(List.empty, x, None, None)), next)))
+                        )
                   }
               }))
           val methodMatcher  = bindParams(List((akkaMethod, List.empty)))
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
@@ -316,7 +316,7 @@ object AkkaHttpServerGenerator {
       directivesFromParams(
         arg => tpe => um => Target.pure(q"parameter(${param(um)(arg)(tpe)})"),
         arg => tpe => um => Target.pure(q"parameter(${param(um)(arg)(tpe)}.*)"),
-        arg => tpe => um => Target.pure(q"parameter(${param(um)(arg)(tpe)}.*).map(xs => Option(xs).filterNot(_.isEmpty)).apply"),
+        arg => tpe => um => Target.pure(q"parameter(${param(um)(arg)(tpe)}.*).map(xs => Option(xs).filterNot(_.isEmpty))"),
         arg => tpe => um => Target.pure(q"parameter(${param(um)(arg)(tpe)}.?)")
       ) _
     }
--- a/modules/codegen/src/test/scala/core/issues/Issue127.scala
+++ b/modules/codegen/src/test/scala/core/issues/Issue127.scala
@@ -149,7 +149,7 @@ class Issue127 extends FunSuite with Matchers with SwaggerSpecRunner {
                   }
                   maybe.fold(reject(_), tprovide(_))
               }))
-            }: Directive[Tuple1[(File, Option[String], ContentType)]])(file => complete(handler.uploadFile(uploadFileResponse)(file)))))
+            }: Directive[Tuple1[(File, Option[String], ContentType)]]).apply(file => complete(handler.uploadFile(uploadFileResponse)(file)))))
           }
         }
         sealed abstract class uploadFileResponse(val statusCode: StatusCode)
--- a/modules/codegen/src/test/scala/tests/generators/akkaHttp/AkkaHttpServerTest.scala
+++ b/modules/codegen/src/test/scala/tests/generators/akkaHttp/AkkaHttpServerTest.scala
@@ -151,13 +151,13 @@ class AkkaHttpServerTest extends FunSuite with Matchers with SwaggerSpecRunner {
           {
             get(pathEndOrSingleSlash(discardEntity(complete(handler.getRoot(getRootResponse)()))))
           } ~ {
-            put(path("bar")(parameter(Symbol("bar").as[Long])(bar => discardEntity(complete(handler.putBar(putBarResponse)(bar))))))
+            put(path("bar")(parameter(Symbol("bar").as[Long]).apply(bar => discardEntity(complete(handler.putBar(putBarResponse)(bar))))))
           } ~ {
             get((pathPrefix("foo") & pathEndOrSingleSlash)(discardEntity(complete(handler.getFoo(getFooResponse)()))))
           } ~ {
-            get(path("foo" / LongNumber)(bar => discardEntity(complete(handler.getFooBar(getFooBarResponse)(bar)))))
+            get(path("foo" / LongNumber).apply(bar => discardEntity(complete(handler.getFooBar(getFooBarResponse)(bar)))))
           } ~ {
-            get(path("store" / "order" / LongNumber)(orderId => parameter(Symbol("status").as[OrderStatus](stringyJsonUnmarshaller.andThen(unmarshallJson[OrderStatus])))(status => discardEntity(complete(handler.getOrderById(getOrderByIdResponse)(orderId, status))))))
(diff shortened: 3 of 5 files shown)


blast-hardcheese:
I'm interested to see whether this compiles; this was added to fix #249, it may trip the regression test. I think #250 may have resolved #210 already, but I'm interested as to why you're getting compilation failures about `Function1`. If you could provide an OpenAPI specification that produces the failure you're seeing, that'd be excellent to take a look at.

rtfpessoa:
Here is an example that breaks, anything else I can help with?
```
swagger: "2.0"
info:
  title: Whatever
  version: 1.0.0
host: localhost:1234
schemes:
  - http
definitions:
  User:
    type: object
    required:
      - id
      - address
    properties:
      id:
        type: string
      address:
        $ref: '#/definitions/UserAddress'
  UserAddress:
    type: object
    properties:
      line1:
        type: string
      line2:
        type: string
      line3:
        type: string
paths:
  /commit/{commitId}/files:
    get:
      tags:
        - file
      summary: Finds all files for commit
      description: 'Finds all files for a commit.'
      operationId: commitCommitIdFiles
      produces:
        - application/json
      parameters:
        - in: path
          name: commitId
          description: Commit identifier
          required: true
          type: integer
          format: int64
          x-example: 1
        - in: query
          name: fileIds
          description: File identifiers
          required: false
          type: array
          items:
            type: integer
            format: int64
          # HACK: Force type Long
          x-scala-type: 'Iterable[Long]'
          collectionFormat: multi
        - in: query
          name: fileNames
          description: File names
          required: false
          type: array
          items:
            type: string
          collectionFormat: multi
      responses:
        '200':
          schema:
            $ref: '#/definitions/User'
        '404':
          description: Not found
```

blast-hardcheese:
Ah, it seems as though this is a divergence of behaviour between 2.11 and 2.12. Finding a structure that compiles in both will be interesting. Are you available to try and fix this? If not, I can take a look after I'm done on my current PR, attempting to address #184

rtfpessoa:
I would gladly do it, but I am not even understanding the exact issue here.

blast-hardcheese:
My understanding is as follows:
```scala
parameter(Symbol("foo").as[String].*).map(Option.apply _)
```
compiled in 2.11 and 2.12. This had the unfortunate effect of returning a `Seq[String]` with no other validation.
```scala
parameter(Symbol("foo").as[String].*).map(xs => Option(xs).filterNot(_.isEmpty))
```
compiles in 2.12 but _not_ 2.11, due to some syntactic ambiguity and related type divergence.
```scala
parameter(Symbol("foo").as[String].*).map(xs => Option(xs).filterNot(_.isEmpty)).apply
```
compiles in 2.11 and 2.12, and tests pass. That being said, your proffered specification _does not_ compile, exposing a previously untested limitation of this suffixed `apply` approach.

Can you post the surrounding code that resulted in
```
[error] missing argument list for method apply in trait Function1
[error] Unapplied methods are only converted to functions when a function type is expected.
[error] You can make this conversion explicit by writing `apply _` or `apply(_)` instead of `apply`.
```
please? It sounds like we're ending up with something like:
```scala
path("foo") {
  parameter(Symbol("foo").as[String].*).map(xs => Option(xs).filterNot(_.isEmpty)).apply
}
```
causing that `apply` to not end up with another `Directive` or `Term.Block` to be applied against.

rtfpessoa:
I just generated with `./cli.sh scala --server --specPath modules/sample/src/main/resources/tes1.yaml --outputPath modules/sample-akkaHttp/target/generated --packageName tes1 --framework akka-http`

Where `tes1.yaml` a file with the spec above.

Anyway here is the content of method around the code:

```
def routes(handler: FileHandler)(implicit mat: akka.stream.Materializer): Route = {
    {
      get(path("commit" / LongNumber / "files")(commitId => (parameter(Symbol("fileIds").as[Long].*).map(xs => Option(xs).filterNot(_.isEmpty)).apply & parameter(Symbol("fileNames").as[String].*).map(xs => Option(xs).filterNot(_.isEmpty)).apply)((fileIds, fileNames) => discardEntity(complete(handler.commitCommitIdFiles(commitCommitIdFilesResponse)(commitId, fileIds, fileNames))))))
    }
  }
```

And the full full Routes.scala:

```
/*
 * This file was generated by Guardrail (https://github.com/twilio/guardrail).
 * Modifications will be overwritten; instead edit the OpenAPI/Swagger spec file.
 */
package tes1.file
import akka.http.scaladsl.model._
import akka.http.scaladsl.model.headers.RawHeader
import akka.http.scaladsl.unmarshalling.{ Unmarshal, Unmarshaller, FromEntityUnmarshaller, FromRequestUnmarshaller, FromStringUnmarshaller }
import akka.http.scaladsl.marshalling.{ Marshal, Marshaller, Marshalling, ToEntityMarshaller, ToResponseMarshaller }
import akka.http.scaladsl.server.Directives._
import akka.http.scaladsl.server.{ Directive, Directive0, Directive1, ExceptionHandler, MalformedHeaderRejection, MissingFormFieldRejection, Rejection, Route }
import akka.http.scaladsl.util.FastFuture
import akka.stream.{ IOResult, Materializer }
import akka.stream.scaladsl.{ FileIO, Keep, Sink, Source }
import akka.util.ByteString
import io.circe.Decoder
import cats.{ Functor, Id }
import cats.data.EitherT
import cats.implicits._
import scala.concurrent.{ ExecutionContext, Future }
import scala.language.higherKinds
import scala.language.implicitConversions
import java.io.File
import java.security.MessageDigest
import java.util.concurrent.atomic.AtomicReference
import scala.util.{ Failure, Success }
import scala.language.higherKinds
import _root_.tes1.Implicits._
import _root_.tes1.AkkaHttpImplicits._
import _root_.tes1.definitions._
trait FileHandler { def commitCommitIdFiles(respond: FileResource.commitCommitIdFilesResponse.type)(commitId: Long, fileIds: Option[Iterable[Long]] = None, fileNames: Option[Iterable[String]] = None): scala.concurrent.Future[FileResource.commitCommitIdFilesResponse] }
object FileResource {
  def discardEntity: Directive0 = extractMaterializer.flatMap { implicit mat => 
    extractRequest.flatMap { req => 
      req.discardEntityBytes().future
      Directive.Empty
    }
  }
  def routes(handler: FileHandler)(implicit mat: akka.stream.Materializer): Route = {
    {
      get(path("commit" / LongNumber / "files")(commitId => (parameter(Symbol("fileIds").as[Long].*).map(xs => Option(xs).filterNot(_.isEmpty)).apply & parameter(Symbol("fileNames").as[String].*).map(xs => Option(xs).filterNot(_.isEmpty)).apply)((fileIds, fileNames) => discardEntity(complete(handler.commitCommitIdFiles(commitCommitIdFilesResponse)(commitId, fileIds, fileNames))))))
    }
  }
  sealed abstract class commitCommitIdFilesResponse(val statusCode: StatusCode)
  case class commitCommitIdFilesResponseOK(value: User) extends commitCommitIdFilesResponse(StatusCodes.OK)
  case object commitCommitIdFilesResponseNotFound extends commitCommitIdFilesResponse(StatusCodes.NotFound)
  object commitCommitIdFilesResponse {
    implicit val commitCommitIdFilesTRM: ToResponseMarshaller[commitCommitIdFilesResponse] = Marshaller { implicit ec => 
      resp => commitCommitIdFilesTR(resp)
    }
    implicit def commitCommitIdFilesTR(value: commitCommitIdFilesResponse)(implicit ec: scala.concurrent.ExecutionContext): scala.concurrent.Future[List[Marshalling[HttpResponse]]] = value match {
      case r @ commitCommitIdFilesResponseOK(value) =>
        Marshal(value).to[ResponseEntity].map {
          entity => Marshalling.Opaque {
            () => HttpResponse(r.statusCode, entity = entity)
          } :: Nil
        }
      case r: commitCommitIdFilesResponseNotFound.type =>
        scala.concurrent.Future.successful(Marshalling.Opaque {
          () => HttpResponse(r.statusCode)
        } :: Nil)
    }
    def apply[T](value: T)(implicit ev: T => commitCommitIdFilesResponse): commitCommitIdFilesResponse = ev(value)
    implicit def OKEv(value: User): commitCommitIdFilesResponse = OK(value)
    def OK(value: User): commitCommitIdFilesResponse = commitCommitIdFilesResponseOK(value)
    def NotFound: commitCommitIdFilesResponse = commitCommitIdFilesResponseNotFound
  }
}
```

blast-hardcheese:
Oh. Of course that won't work.

blast-hardcheese:
If you `runScalaExample` and `akkaHttpSample/test:compile`, you should see the compiler errors that caused this build to fail. If you're inclined to throw some time at figuring out a syntax that makes all use cases work (yours included) I'd be more than happy to merge it and cut a bugfix release. If not, I'll try and find some time to resolve this, as I'm sure it'll bite others sooner than later.

blast-hardcheese:
Ok -- when rebasing, the tests continued to fail with the same problem that was originally happening.

I was able to work around it by putting the `apply` at the point at which we create the inner function literal to consume the supplied fields. This seems to force the magnet pattern to solidify, and type inference works again after that point.

rtfpessoa:
@blast-hardcheese nice. I tried a couple times, but I was not able to get anything close to the real issue.
Thanks for the help.

Issue #290, opened 2019-05-10
Multiple date query params generate ambiguous implicit values


kkreuning:
When I run the http4s server code generator with the following (minimal) specification

```yaml
openapi: "3.0.0"
 info:
   title: Generator Error Sample
   version: 1.0.0
 paths:
   /events:
     parameters:
       - $ref: '#/components/parameters/FromParam'
       - $ref: '#/components/parameters/ToParam'
     get:
       operationId: getEvents
       responses:
         200:
           description: Requested events
 components:
   parameters:
     FromParam:
       name: from
       in: query
       schema:
         type: string
         format: date
     ToParam:
       name: to
       in: query
       schema:
         type: string
         format: date
```

It generates the following code:

```scala
Http4sDsl[F] {
  implicit val toQueryParamDecoder: QueryParamDecoder[java.time.LocalDate] = (value: QueryParameterValue) => Json.fromString(value.value).as[java.time.LocalDate].leftMap(t => ParseFailure("Query decoding failed", t.getMessage)).toValidatedNel
  implicit val fromQueryParamDecoder: QueryParamDecoder[java.time.LocalDate] = (value: QueryParameterValue) => Json.fromString(value.value).as[java.time.LocalDate].leftMap(t => ParseFailure("Query decoding failed", t.getMessage)).toValidatedNel
  object GetEventsToMatcher extends OptionalQueryParamDecoderMatcher[java.time.LocalDate]("to")
  object GetEventsFromMatcher extends OptionalQueryParamDecoderMatcher[java.time.LocalDate]("from")
  def routes(handler: Handler[F]): HttpRoutes[F] = HttpRoutes.of {
    {
      case req @ GET -> Root / "events" :? GetEventsFromMatcher(from) +& GetEventsToMatcher(to) =>
        handler.getEvents(GetEventsResponse)(from, to) flatMap {
          case GetEventsResponse.Ok =>
            Ok()
        }
    }
  }
}
```

When I compile, I get the following error:

```
[error] /.../Routes.scala:40:42: ambiguous implicit values:
[error]  both value fromQueryParamDecoder in class Resource of type => org.http4s.QueryParamDecoder[java.time.LocalDate]
[error]  and value toQueryParamDecoder in class Resource of type => org.http4s.QueryParamDecoder[java.time.LocalDate]
[error]  match expected type org.http4s.QueryParamDecoder[java.time.LocalDate]
[...]

PR #291, opened 2019-05-10
Fix issue 290.

kkreuning:
Fixes https://github.com/twilio/guardrail/issues/290

This fix generates implicit query param decoders looking like this:

```scala
implicit val `java.time.LocalDateQueryParamDecoder`: QueryParamDecoder[java.time.LocalDate] = (value: QueryParameterValue) => Json.fromString(value.value).as[java.time.LocalDate].leftMap(t => ParseFailure("Query decoding failed", t.getMessage)).toValidatedNel
```

It's not pretty but it works, then again implicits are not shown to the user and the compiler doesn't care about nice names...

Another option would be to have default `QueryParamDecoder` for types like `java.time.LocalDate` (and `java.time.LocalDateTime` etc.) since they are 'standard' types returned by the Swagger generator. That way only custom types (using `x-scala-type`) would need specific generated implicits. Any ideas on this?

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
@@ -651,7 +651,7 @@ object Http4sServerGenerator {
             }
             if (!List("Unit", "Boolean", "Double", "Float", "Short", "Int", "Long", "Char", "String").contains(elemType.toString())) {
               val queryParamDecoder = q"""
-                implicit val ${Pat.Var(Term.Name(s"${argName.value}QueryParamDecoder"))}: QueryParamDecoder[$elemType] = (value: QueryParameterValue) =>
+                implicit val ${Pat.Var(Term.Name(s"${elemType}QueryParamDecoder"))}: QueryParamDecoder[$elemType] = (value: QueryParameterValue) =>
                     Json.fromString(value.value).as[$elemType]
                       .leftMap(t => ParseFailure("Query decoding failed", t.getMessage))
                       .toValidatedNel
--- a/modules/codegen/src/test/scala/core/issues/Issue290.scala
+++ b/modules/codegen/src/test/scala/core/issues/Issue290.scala
@@ -1,0 +1,65 @@
+package core.issues
+
+import com.twilio.guardrail.{Context, Server, Servers}
+import com.twilio.guardrail.generators.Http4s
+import org.scalatest.{FunSpec, Matchers}
+import support.SwaggerSpecRunner
+
+class Issue290 extends FunSpec with Matchers with SwaggerSpecRunner {
+
+  describe("Multiple date query params generate ambiguous implicit values") {
+
+    val specification = """
+      | openapi: "3.0.0"
+      | info:
+      |   title: Generator Error Sample
+      |   version: 1.0.0
+      | paths:
+      |   /events:
+      |     parameters:
+      |       - $ref: '#/components/parameters/FromParam'
+      |       - $ref: '#/components/parameters/ToParam'
+      |     get:
+      |       operationId: getEvents
+      |       responses:
+      |         200:
+      |           description: Requested events
+      | components:
+      |   parameters:
+      |     FromParam:
+      |       name: from
+      |       in: query
+      |       schema:
+      |         type: string
+      |         format: date
+      |     ToParam:
+      |       name: to
+      |       in: query
+      |       schema:
+      |         type: string
+      |         format: date""".stripMargin
+
+    val (
+      _, // ProtocolDefinitions
+      _, // clients
+      Servers(
+        Server(
+          _, // pkg
+          _, // extraImports
+          _, // genHandler
+          serverDefinition :: _
+        ) :: Nil,
+        _ // supportDefinitions
+      )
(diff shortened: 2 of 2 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -636,7 +636,7 @@ object Http4sServerGenerator {
             }
             if (!List("Unit", "Boolean", "Double", "Float", "Short", "Int", "Long", "Char", "String").contains(elemType.toString())) {
               val queryParamDecoder = q"""
-                implicit val ${Pat.Var(Term.Name(s"${argName.value}QueryParamDecoder"))}: QueryParamDecoder[$elemType] = (value: QueryParameterValue) =>
+                implicit val ${Pat.Var(Term.Name(s"${elemType}QueryParamDecoder"))}: QueryParamDecoder[$elemType] = (value: QueryParameterValue) =>
```
blast-hardcheese: These `elemType`s are not guaranteed to be fully qualified, as we won't always know that information, consider:

```
- name: foo
  x-scala-type: Foo
- name: foo2
  x-scala-type: com.example.Foo
```
given `imports=List("com.example._")` in build.sbt.

I wonder if
```
implicit def decoderDecoder[A: Decoder]: QueryParamDecoder[A] =
  Json.fromString(value.value).as[A]
    .leftMap(t => ParseFailure("Query decoding failed", t.getMessage))
    .toValidatedNel
```
would be more stable, relying on scalac to do the wiring up for us.

One unfortunate aspect here is that `Json.fromString` prevents non-string query parameters. I'd expect
```
- name: numeric
  type: integer
  format: int64
```
would fail to parse anything, see #184 and #292. This is outside of the scope of your library upgrade here, just raising awareness of an issue.

blast-hardcheese:
> Another option would be to have default QueryParamDecoder for types like java.time.LocalDate (and java.time.LocalDateTime etc.) since they are 'standard' types returned by the Swagger generator. That way only custom types (using x-scala-type) would need specific generated implicits. Any ideas on this?

Circling back around to this PR, I think I misunderstood what you meant here.

Currently, we have two support definitions: One for the parameter extractors, and another for the type decoders, is this correct?

If so, I believe that the reusable type decoders would make sense to go in `Http4sImplicits`; as you say, these are common and can be shared. That would leave the individual parameter extractors functioning as they currently do, which makes sense.

The way we would do extension in this way would be to either rely on the circe `Decoder`s like we currently do, or by consumers providing their own decoders by way of `imports=...` in their project configuration.

Sorry for having kept this open for as long as we have; if you've got thoughts around this, I'd appreciate it.

eli-jordan:
We just ran into this issue on my team, so if I can help to get it resolved, I'd be happy to help!

Perhaps a simple fix would be to pass the `QueryParamDecoder` explicitly, and not have it be an `implicit val`. Similar to what is done for the `EntityEncoder` and `EntityDecoder` instances

blast-hardcheese:
@eli-jordan If the simple fix of passing the implicits explicitly is straightforward, that may be sufficient. Alternately, as I mentioned in my previous comment, perhaps defining the (effectively constant) implicits in a single place in `Http4sImplicits`, that's what that object is designed to hold in the first place -- type class instances that are then used in all the different parts of guardrail-generated clients and servers.

blast-hardcheese:
I'm onboard with fixing this as simply as possible, weighing pragmatism against future-proofing. I'm a bit at a disadvantage, however, as my experience with http4s is somewhat limited.

blast-hardcheese:
@hanny24 Maybe you have some ideas around this, as you've been in and around this part of the codebase recently? I'd like to get this resolved, but I don't know what would be the most idiomatic from an http4s usage perspective. Perhaps just a reusable adapter in the `Http4sImplicits` package object?

eli-jordan:
Adding the standard types to the implicits object will help. However, I don’t think it will cover the case where the same custom type is used in multiple places. That’s why I suggested passing them explicitly, even though that’s not really the most idiomatic solution. Otherwise, you would need to dedup instances for the same type.

hanny24:
I think that the best was how to deal with this is to add standard types to `Implicits` object and let user solve any other case by providing custom implicit instance of `QueryParamDecoder` by the means of imports.

Another solution would be to create a newtype in request's companion for each parameter. Something like this:

```scala
object GetEventsRequest {
  case class FromParam(value: java.time.LocalDate)
  object FromParam {
      implicit val toQueryParamDecoder: QueryParamDecoder[java.time.LocalDate] = (value: QueryParameterValue) => Json.fromString(value.value).as[java.time.LocalDate].leftMap(t => ParseFailure("Query decoding failed", t.getMessage)).toValidatedNel
   }
   // the same for ToParam
}
```

This solution is more difficult to implement and I'm not really sure it's worth it.

blast-hardcheese:
I agree with this approach. Unifying the type decoder into the core implicits makes sense to me. I don't have time to resolve this, but if it gets done I'm onboard with merging and releasing it.

blast-hardcheese:
Circling back, I ended up adopting and extending this technique in #506. It would have probably been good to abstract, but at least it'll work in the interim

blast-hardcheese:
Closing this, sorry for not merging earlier.

PR #292, opened 2019-05-11
#184 unmarshaller shuffle

blast-hardcheese:
Resolving #184 

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
@@ -315,7 +315,7 @@ object SwaggerUtil {
           } yield res
         case o: ObjectSchema =>
           for {
-            _ <- log.debug(s"Not attempting to process properties from ${o.showNotNull}")
+            _   <- log.debug(s"Not attempting to process properties from ${o.showNotNull}")
             res <- objectType(None).map(Resolved[L](_, None, None)) // TODO: o.getProperties
           } yield res
 
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
@@ -145,11 +145,11 @@ object SwaggerUtil {
   sealed class ModelMetaTypePartiallyApplied[L <: LA, F[_]](val dummy: Boolean = true) {
     def apply[T <: Schema[_]](
         model: T
-    )(implicit Sc: ScalaTerms[L, F], Sw: SwaggerTerms[L, F], F: FrameworkTerms[L, F]): Free[F, ResolvedType[L]] =
+    )(implicit Sc: ScalaTerms[L, F], Sw: SwaggerTerms[L, F], Fw: FrameworkTerms[L, F]): Free[F, ResolvedType[L]] =
       Sw.log.function("modelMetaType") {
         import Sc._
         import Sw._
-        import F._
+        import Fw._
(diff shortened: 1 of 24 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpGenerator.scala
```diff
@@ -91,6 +94,28 @@ object AkkaHttpGenerator {
                     Json.fromString(data.decodeString("utf-8"))
                 })
 
+            // Translate HttpEntity => Json (for `text/plain`, relying on the Decoder to reject incorrect types.
+            //   This permits not having to manually construct ToStringMarshaller/FromStringUnmarshallers.
+            //   This is definitely lazy, but lets us save a lot of scalar parsers as circe decoders are fairly common.)
+            final val sneakyJsonEntityUnmarshaller: FromEntityUnmarshaller[${jsonType}] =
```
kelnos: sneaky!  :D

modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
```diff
@@ -213,28 +213,36 @@ object AkkaHttpServerGenerator {
...
 
     def directivesFromParams(
-        required: Term => Type => Target[Term],
-        multi: Term => Type => Target[Term],
-        multiOpt: Term => Type => Target[Term],
-        optional: Term => Type => Target[Term]
+        required: Term => Type => Option[Term] => Target[Term],
+        multi: Term => Type => Option[Term] => Target[Term],
+        multiOpt: Term => Type => Option[Term] => Target[Term],
+        optional: Term => Type => Option[Term] => Target[Term]
```
kelnos: Any readability concerns with a function-function-function?

blast-hardcheese: I tried to maintain the level of precedence for relevancy. I have no objections to this style.

That being said, an alternative would be a tuple or a named parameter with explicit `unapply` overrides to prevent every field needing to be explicitly updated any time anything changes with the source API.

blast-hardcheese:
One major downside of this PR in its current state is that the "base" type is unable to be traversed for deferred datatypes. I'm updating the branch at this point, though I need to go back and reevaluate some of this approach.

modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
```diff
@@ -141,11 +143,13 @@ object SwaggerUtil {
...
   sealed class ModelMetaTypePartiallyApplied[L <: LA, F[_]](val dummy: Boolean = true) {
-    def apply[T <: Schema[_]](model: T)(implicit Sc: ScalaTerms[L, F], Sw: SwaggerTerms[L, F], Fw: FrameworkTerms[L, F]): Free[F, ResolvedType[L]] =
+    def apply[T <: Schema[_]](
+        model: T
+    )(implicit Sc: ScalaTerms[L, F], Sw: SwaggerTerms[L, F], F: FrameworkTerms[L, F]): Free[F, ResolvedType[L]] =
       Sw.log.function("modelMetaType") {
         import Sc._
         import Sw._
-        import Fw._
+        import F._
```
kelnos: Minor nit: I kinda prefer this as `Fw` so it doesn't get confused with the type param `F`.

blast-hardcheese: Yep, not sure why this got changed, but changing it back now.

PR #296, opened 2019-05-16
Security plumbing

kelnos:
This doesn't yet add security support to any client/server generators, but threads through the defined security schemes to enable that.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/extract/Extractable.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/extract/Extractable.scala
@@ -1,15 +1,21 @@
 package com.twilio.guardrail.extract
-import com.twilio.guardrail.{ EmptyIsEmpty, EmptyIsNull, EmptyToNullBehaviour }
 
-import scala.util.Try
+import com.twilio.guardrail.{ EmptyIsEmpty, EmptyIsNull, EmptyToNullBehaviour }
+import scala.util.{ Success, Try }
+import scala.reflect.ClassTag
+import scala.collection.JavaConverters._
+import java.util
 
 trait Extractable[T] {
   def extract(v: Any): Try[T]
 }
 
 object Extractable {
+  private[this] def buildExceptionally[T](f: PartialFunction[Any, Try[T]]): Extractable[T] =
+    (v: Any) => Try(f(v)).flatten
+
   def build[T](f: PartialFunction[Any, T]): Extractable[T] =
-    (v: Any) => Try(f(v))
+    buildExceptionally(f.andThen(Success(_)))
 
   implicit val defaultExtractableBoolean: Extractable[Boolean] =
     build[Boolean]({ case x: Boolean => x })
@@ -37,4 +43,16 @@ object Extractable {
       case x: Boolean if x  => EmptyIsNull
       case x: Boolean if !x => EmptyIsEmpty
     })
+
+  implicit def defaultExtractableList[T: Extractable: ClassTag]: Extractable[List[T]] =
+    buildExceptionally[List[T]](validateSeq.andThen(xs => Try(validateListItems[T].apply(xs))))
+
+  private[this] def validateSeq: PartialFunction[Any, List[_]] = {
+    case x: Seq[_]       => x.toList
+    case x: util.List[_] => x.asScala.toList
+  }
+
+  private[this] def validateListItems[A](implicit cls: ClassTag[A]): PartialFunction[List[_], List[A]] = {
+    case xs: List[_] if xs.forall(x => cls.runtimeClass.isAssignableFrom(x.getClass)) => xs.asInstanceOf[List[A]]
+  }
 }
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/extract/VendorExtension.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/extract/VendorExtension.scala
@@ -1,9 +1,10 @@
 package com.twilio.guardrail.extract
 
-import io.swagger.v3.oas.models.media._
 import io.swagger.v3.oas.models._
+import io.swagger.v3.oas.models.media._
 import io.swagger.v3.oas.models.parameters.{ CookieParameter, Parameter }
 import io.swagger.v3.oas.models.responses.ApiResponse
+import io.swagger.v3.oas.models.security.SecurityScheme
 
 object VendorExtension {
   trait VendorExtensible[-F] {
@@ -66,6 +67,8 @@ object VendorExtension {
     implicit val defaultVendorExtensibleCookieParameter: VendorExtensible[CookieParameter] =
       build[CookieParameter](m => key => Option(m.getExtensions).map(_.get(key)))
 
+    implicit val defaultVendorExtensibleSecurityScheme: VendorExtensible[SecurityScheme] =
+      build[SecurityScheme](m => key => Option(m.getExtensions).map(_.get(key)))
   }
 
   case class VendorExtensibleAdapter[F](from: F) extends AnyVal {
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
@@ -32,7 +32,13 @@ object AkkaHttpClientGenerator {
         .fold(param"host: String")(v => param"host: String = ${Lit.String(v.head.toString())}")
 
     def apply[T](term: ClientTerm[ScalaLanguage, T]): Target[T] = term match {
-      case GenerateClientOperation(_, _ @RouteMeta(pathStr, httpMethod, operation), methodName, tracing, parameters, responses) =>
+      case GenerateClientOperation(_,
+                                   RouteMeta(pathStr, httpMethod, operation, securityRequirements),
+                                   methodName,
+                                   tracing,
+                                   parameters,
+                                   responses,
+                                   securitySchemes) =>
         def generateUrlWithParams(path: String, pathArgs: List[ScalaParameter[ScalaLanguage]], qsArgs: List[ScalaParameter[ScalaLanguage]]): Target[Term] =
           Target.log.function("generateUrlWithParams") {
             for {
@@ -309,7 +315,7 @@ object AkkaHttpClientGenerator {
       case GenerateResponseDefinitions(operationId, responses, protocolElems) =>
         Target.pure(Http4sHelper.generateResponseDefinitions(operationId, responses, protocolElems))
 
-      case GenerateSupportDefinitions(tracing) =>
+      case GenerateSupportDefinitions(tracing, securitySchemes) =>
         Target.pure(List.empty)
 
       case BuildStaticDefns(clientName, tracingName, serverUrls, ctorArgs, tracing) =>
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
@@ -28,7 +28,13 @@ object EndpointsClientGenerator {
         .fold(param"host: String")(v => param"host: String = ${Lit.String(v.head.toString())}")
 
     def apply[T](term: ClientTerm[ScalaLanguage, T]): Target[T] = term match {
-      case GenerateClientOperation(className, route @ RouteMeta(pathStr, httpMethod, operation), methodName, tracing, parameters, responses) =>
+      case GenerateClientOperation(className,
+                                   RouteMeta(pathStr, httpMethod, operation, securityRequirements),
+                                   methodName,
+                                   tracing,
+                                   parameters,
+                                   responses,
+                                   securitySchemes) =>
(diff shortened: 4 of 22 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/extract/Extractable.scala
```diff
@@ -37,4 +39,9 @@ object Extractable {
       case x: Boolean if x  => EmptyIsNull
       case x: Boolean if !x => EmptyIsEmpty
     })
+  implicit def defaultExtractableList[T]: Extractable[List[T]] =
+    build[List[T]]({
+      case x: Seq[T]       => x.toList
```
blast-hardcheese: This seems dangerous. Any new warnings around erasure here?

blast-hardcheese: I'd slap on an `Extractable[T]` constraint and fold through the sequence to ensure match

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
```diff
@@ -157,8 +158,19 @@ case class RouteMeta(path: String, method: HttpMethod, operation: Operation) {
       })
 }
 
+sealed trait SecurityScheme {
+  def typeName: Option[String]
+  def extensions: Map[String, AnyRef]
```
blast-hardcheese: Is there a well-typed variant that we could use instead of `AnyRef`? Refactors or API changes that would need to be changed in more than one place would be more difficult to find if we use reflection all over the place

kelnos: This is just a reflection of what the swagger parser gives us: `java.util.Map[String, Object]`.  There isn't really a better way to represent it since keys could be anything supported in yaml.

blast-hardcheese: What are they consumed by? Something is going to need to do reflection, just wondering if there's some sort of schema we can apply earlier in the process

kelnos: Soooo, nothing, it turns out.  I was originally going to support non-standard security types via vendor extensions on the `SecurityScheme` instance, but I ended up finding it's easier/nicer to just (ab)use the auth-scheme field when the security type is `http`.  So I can just remove this field entirely.

blast-hardcheese: Sick. We can explicitly add vendor extensions as necessary.

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
```diff
@@ -157,8 +158,18 @@ case class RouteMeta(path: String, method: HttpMethod, operation: Operation) {
       })
 }
 
+sealed trait SecurityScheme {
+  def typeName: Option[String]
```
blast-hardcheese: Why are these `String`?  Is this to avoid `L#Type`?

kelnos: We could do `L#Type`, but unfortunately `SwaggerGenerator` doesn't have an instance of `ScalaTerms`.

blast-hardcheese: Forgot about that point. We could add it, as there's no penalty, but we'll see what the consumer looks like

kelnos: I have a consumer in a branch that doesn't yet work... it ends up feeding the string to something that returns the language-equivalent of `L#Type`, so this _would_ be useful.

kelnos:
I ended up semi-redoing this.  I noticed that the swagger parser doesn't implement "inheritance".  That is, if there's a toplevel `security:` item, it doesn't get automatically set on all the `Operation` instances.

I created a new `SecurityRequirements` case class that is now a member of `RouteMeta`.  It propagates the security requirements on the `Operation` (falling back to the requirements on the root object if there aren't any), the security schemes marked as optional (via our vendor extension), and whether or not the requirement comes from the global/root level or not.

For the last bit, I want clients to be able to know if the security requirement is there for all operations, or just some.  This way a client can make the API a little nicer: if the requirement is at the global level the constructor for the client itself can take credentials, and then the individual operation methods can take creds as an optional param.  If the requirement is at the operation level, then the client constructor won't take credentials, and the operation methods will take creds as required params.

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
```diff
@@ -29,7 +32,16 @@ object RouteMeta {
     }
   }
 }
-case class RouteMeta(path: String, method: HttpMethod, operation: Operation) {
+
+object SecurityRequirements {
+  type SecurityScopes = List[String]
+
+  def apply(requirements: util.List[SecurityRequirement], optionalSchemes: List[String], isGlobal: Boolean): SecurityRequirements =
```
blast-hardcheese: I'm seeing `isGlobal = true` and `isGlobal = false`. Initially I thought that was to override default parameters, but now I see it's for clarity. I'd request `IsGlobal` defined as a `sealed trait` with `Global` and `Local` (or `NonGlobal`, or something else) members, explicitly with the intent to disambiguate.

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
blast-hardcheese: On further reflection, I can't actually see a good way to call `parseType` given what we end up with here. The only alternative is something like a `case class SecuritySchemeConstructor[L <: LA](tpe: String, builder: L# Type => SecurityScheme)`, but it's clearly working around the paradigms, not working with them.

Thinking through how we resolve this practice in other places, consumers of `extractSecuritySchemes` could be rewritten to:
```scala
      securitySchemes <- Option(swagger.getComponents)
        .flatMap(components => Option(components.getSecuritySchemes))
        .map(_.asScala.toMap)
        .getOrElse(Free.pure[F, Map[String, SwSecurityScheme]](Map.empty))
        .toList
        .traverse { case (schemeName, scheme) =>
          // Decompose processor behaviour into distinct chunks with explicit extractors instead of a monolith that owns all responsibility
        }
```
... but I wonder if this is ceremony for ceremony's sake. What do you think?

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
blast-hardcheese: Thanks!

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
kelnos: I actually started going down that path in the first place, but then found that I had to copy/paste the same code into the client generator, server generator, and one of the test runners, so I opted to do it this way.

Alternatively, I could stash this code in `SwaggerUtil` and just call it from those three places.  That way I think we'd be able to do `parseType()` there as well, right?

blast-hardcheese: Yeah; a lot of `SwaggerUtil` is becoming just `Free` utility functions, so this would be reasonable. The motivator here is to not accidentally hard couple exactly which type parser we're using, keeping everything overridable by interpreters. Thanks for considering this approach!

modules/codegen/src/main/scala/com/twilio/guardrail/extract/VendorExtension.scala
```diff
@@ -66,6 +67,8 @@ object VendorExtension {
     implicit val defaultVendorExtensibleCookieParameter: VendorExtensible[CookieParameter] =
       build[CookieParameter](m => key => Option(m.getExtensions).map(_.get(key)))
 
+    implicit val defaultVendorExtensibleSecurityScheme: VendorExtensible[SecurityScheme] =
+      build[SecurityScheme](m => key => Option(m.getExtensions).map(_.get(key)))
```
blast-hardcheese: I hesitate to ask, but what happens if `get(key)` fails? `null` or `Exception`?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
```diff
@@ -32,7 +32,13 @@ object AkkaHttpClientGenerator {
         .fold(param"host: String")(v => param"host: String = ${Lit.String(v.head.toString())}")
 
     def apply[T](term: ClientTerm[ScalaLanguage, T]): Target[T] = term match {
-      case GenerateClientOperation(_, _ @RouteMeta(pathStr, httpMethod, operation), methodName, tracing, parameters, responses) =>
+      case GenerateClientOperation(_,
+                                   _ @RouteMeta(pathStr, httpMethod, operation, securityRequirements),
```
blast-hardcheese: `_@RouteMeta` => `RouteMeta`

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
```diff
@@ -28,7 +28,13 @@ object EndpointsClientGenerator {
         .fold(param"host: String")(v => param"host: String = ${Lit.String(v.head.toString())}")
 
     def apply[T](term: ClientTerm[ScalaLanguage, T]): Target[T] = term match {
-      case GenerateClientOperation(className, route @ RouteMeta(pathStr, httpMethod, operation), methodName, tracing, parameters, responses) =>
+      case GenerateClientOperation(className,
+                                   route @ RouteMeta(pathStr, httpMethod, operation, securityRequirements),
```
blast-hardcheese: seems as though `route` is unused

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sClientGenerator.scala
```diff
@@ -32,7 +32,13 @@ object Http4sClientGenerator {
         .fold(param"host: String")(v => param"host: String = ${Lit.String(v.head.toString())}")
 
     def apply[T](term: ClientTerm[ScalaLanguage, T]): Target[T] = term match {
-      case GenerateClientOperation(className, route @ RouteMeta(pathStr, httpMethod, operation), methodName, tracing, parameters, responses) =>
+      case GenerateClientOperation(className,
+                                   route @ RouteMeta(pathStr, httpMethod, operation, securityRequirements),
```
blast-hardcheese: seems as though `route` is unused

modules/codegen/src/main/scala/com/twilio/guardrail/extract/VendorExtension.scala
kelnos: It's just a `java.util.Map`, I believe, so `null`.  At any rate, that's how all the other `VendorExtensible` impls work, so I assumed you believed `null` was ok there, and that something else in the extension extraction would take care of it?  Seeing as the vendor extensions actually work and never blow up, seems like it must be ok.

blast-hardcheese: So these should actually all be `flatMap` on `v => Option(v.get(key))`?

blast-hardcheese: otherwise we're relying on the trinary state of somehow ending up with an empty security model?

kelnos: Literally all the other `VendorExtensible` implementations do exactly the same thing.  The reason why it works is because the `PartialFunction` impls for `Extractable` all match on types that are not `Object`, and so they'll never get a `null` value passed to them.

kelnos: Essentially it works like this:
```
scala> val xs = List(1, "a", null, true)
xs: List[Any] = List(1, a, null, true)

scala> val pf: PartialFunction[Any, String] = {
     | case s: String => s
     | }
pf: PartialFunction[Any,String] = <function1>

scala> xs.map(x => scala.util.Try(pf(x)).toOption)
res0: List[Option[String]] = List(None, Some(a), None, None)
```

blast-hardcheese: Yep, there's just a completely different set of neurons firing when reading other people's PRs, even if it's code I originally wrote or maintained. Just looking at continuing to improve the codebase

kelnos: Yeah... I think there's room for improvement here; if someone ever changes `Extractable` to not be based on partial functions (or something) or tries to remove the `Try()` wrapper on calling the extractor, then things might stop working.  But I think we can leave that for antoher day...

blast-hardcheese: Fair. I'll create an issue to track

PR #299, opened 2019-05-21
Fix sbt assembly command

kelnos:
Adding the Eclipse JDT formatter for Java code broke assembly because of a bunch of duplicate files in the Eclipse JARs that need merge strategies.

**Contributing to Twilio**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -23,6 +23,16 @@ val dropwizardVersion = "1.3.9"
 val jerseyVersion     = "2.25.1"
 
 mainClass in assembly := Some("com.twilio.guardrail.CLI")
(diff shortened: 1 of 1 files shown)


kelnos:
sbt-assembly has a few built-in rules I think for what to do with some kinds of files.  But if it encounters something else, and the files don't have the same contents, it errs on the side of failing vs. possibly creating a broken JAR.  In this case, yeah, different contents for the files I added.  I think we could actually just `discard` them all, because the code paths I use don't do any of Eclipse's magical plugin/extension-point class loading, but I figured might as well make a best effort to make it happy.

PR #278, opened 2019-05-03
Better content type support for AsyncHttpClient

kelnos:
Adds text/plain and application/octet-stream support for AHC request and response bodies.

Also improves error reporting from the Java code formatter rather than just dumping a stack trace to console.

**Contributing to Twilio**

> All third party contributors acknowledge that any contributions they provide will be made under the same open source license that the open source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #300, opened 2019-05-22
Better content type support for AsyncHttpClient

kelnos:
Take two of https://github.com/twilio/guardrail/pull/278

Adds text/plain and application/octet-stream support for AHC request and response bodies.

Also improves error reporting from the Java code formatter rather than just dumping a stack trace to console.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
@@ -2,6 +2,7 @@ package com.twilio.guardrail.generators.Java
 
 import cats.data.NonEmptyList
 import cats.instances.list._
+import cats.syntax.foldable._
 import cats.syntax.traverse._
 import cats.~>
 import com.github.javaparser.JavaParser
@@ -12,15 +13,21 @@ import com.github.javaparser.ast.body._
 import com.github.javaparser.ast.expr.{ MethodCallExpr, NameExpr, _ }
 import com.github.javaparser.ast.stmt._
 import com.twilio.guardrail.generators.Java.AsyncHttpClientHelpers._
-import com.twilio.guardrail.generators.ScalaParameter
+import com.twilio.guardrail.generators.{ ScalaParameter, ScalaParameters }
 import com.twilio.guardrail.generators.syntax.Java._
 import com.twilio.guardrail.languages.JavaLanguage
 import com.twilio.guardrail.protocol.terms.Response
 import com.twilio.guardrail.protocol.terms.client._
+import com.twilio.guardrail.shims._
 import com.twilio.guardrail.terms.RouteMeta
+import com.twilio.guardrail.terms.RouteMeta.ContentType
 import com.twilio.guardrail.{ RenderedClientOperation, StaticDefns, SupportDefinition, SwaggerUtil, Target }
+import io.swagger.v3.oas.models.Operation
+import io.swagger.v3.oas.models.media.MediaType
 import java.net.URI
 import java.util
+import java.util.Locale
+import scala.collection.JavaConverters._
 
 object AsyncHttpClientClientGenerator {
   private val URI_TYPE                       = JavaParser.parseClassOrInterfaceType("URI")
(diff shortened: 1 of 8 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
```diff
@@ -471,18 +555,47 @@ object AsyncHttpClientClientGenerator {
...
               new BlockStmt(
-                callBuilderFinalFields
-                  .map[Statement, List[Statement]]({
-                    case (_, name) =>
-                      new ExpressionStmt(new AssignExpr(new FieldAccessExpr(new ThisExpr, name), new NameExpr(name), AssignExpr.Operator.ASSIGN))
-                  })
-                  .toNodeList
+                (
+                  callBuilderFinalFields
+                    .map[Statement, List[Statement]]({
```
blast-hardcheese: Is this just to avoid `List[Statement](new ExpressionStmt(..))`?

modules/sample/src/main/resources/petstore.json
```diff
@@ -788,8 +788,7 @@
         "description": "",
         "operationId": "loginUser",
         "produces": [
-          "application/xml",
```
blast-hardcheese: This left because we didn't know what `application/xml` was? Any reason why this can't just be handled by handling the "best" one for now? My concern is folk who don't have direct ownership of the swagger doc to the point where they mandate an incomplete spec be curated just for them

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
```diff
@@ -37,6 +44,52 @@ object AsyncHttpClientClientGenerator {
...
+    )
+
+    response.value
+      .map(_._1)
+      .flatMap({ valueType =>
+        val allProduces = operation.getResponses.asScala
+          .get(response.statusCode.toString)
+          .flatMap(resp => Option(resp.getContent))
+          .fold(List.empty[(String, MediaType)])(_.asScala.toList)
+          .flatMap({ case (name, _) => ContentType.unapply(name.toLowerCase(Locale.US)) })
```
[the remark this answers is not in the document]
kelnos: (For the last bit: it's not a map, it's an option, so collectFirst is the way to go.)

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
kelnos: No, it's already a list.  unfortunately mapping without the type ascription gives us a `List[ExpressionStmt]`, then `NodeList[ExpressionStmt]`, which, due to variance (or that lack thereof?), makes the types not line up, because the `BlockStmt` constructor wants a `NodeList[Statement]`.

modules/sample/src/main/resources/petstore.json
kelnos: I needed something to test text/plain with and I guess hadn't noticed contentType-textPlain.yaml yet.  Will revert this part.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
blast-hardcheese: Ah, so this is to avoid `new ExpressionStmt(...): Statement`. Makes sense.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
```diff
@@ -37,6 +44,52 @@ object AsyncHttpClientClientGenerator {
...
 
+  def getBestConsumes(operation: Operation, parameters: ScalaParameters[JavaLanguage]): Option[ContentType] =
+    if (parameters.formParams.nonEmpty) {
+      if (parameters.formParams.exists(_.isFile)) {
+        Option(RouteMeta.MultipartFormData)
+      } else {
+        Option(RouteMeta.UrlencodedFormData)
+      }
+    } else if (parameters.bodyParams.isDefined) {
+      val allConsumes = operation.consumes.flatMap(ContentType.unapply)
```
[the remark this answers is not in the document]
kelnos: Agreed.

kelnos:
Ok, comments addressed, and also added a test.

PR #301, opened 2019-05-22
Remove AHC baseUrl/clientName final modifier when we don't want it

kelnos:
If the openapi spec gives us a `baseUrl` or `clientName`, we want to initialize it as the default, but allow overriding using a method on the builder (so no `final` modifier on the instance var).  If there is none provided, we take it in the constructor and offer no means to override (so yes `final` modifier on the instance var).

A side note: it looks like `FieldDeclaration#setFinal()` doesn't really work.  Odd.

**Contributing to Twilio**


kelnos:
Noted offline why I can't add a regression test (tracing is disabled for java here); will add testing in the future.

PR #311, opened 2019-05-24
Allow overriding required POJO properties with builder methods

kelnos:
Since required properties are passed to the builder constructor, there's usually no need to be able to call `withSomeProperty(...)` to later set the value (again).  However, since I added the "copy" constructor for the builder, this would mean that you wouldn't be able to use that constructor and then modify the value of a required property, which greatly reduces its usefulness.  So, we remove that restriction and add a few more checks to ensure that the object remains consistent.

**Contributing to Twilio**


kelnos:
I've so far been opinionated to the point that I don't want to see `null`s ever, and this just continues that.  the `withFoo(T foo)` methods will always error out if you pass `null` to that, and the `withFoo(Optional<T> foo)` methods are provided so you can blindly call the method even if you don't know if you have a value or not.  But I still won't accept `null`.

blast-hardcheese:
OK, that makes sense. If this is something you think we should be enforcing on the Java side of the world, I'm onboard.

PR #316, opened 2019-05-30
Add basic redaction via new vendor extension 'x-data-redaction'

kelnos:
Redacts fields in circe- and jackson-generated objects' `toString` methods.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
@@ -2,9 +2,10 @@ package com.twilio.guardrail
 package generators
 
 import _root_.io.swagger.v3.oas.models.media._
+import cats.data.NonEmptyList
 import cats.implicits._
 import cats.~>
-import com.twilio.guardrail.extract.{ Default, EmptyValueIsNull }
+import com.twilio.guardrail.extract.{ DataRedaction, Default, EmptyValueIsNull }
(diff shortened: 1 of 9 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -141,6 +142,7 @@ object CirceProtocolGenerator {
             case s: StringSchema    => EmptyValueIsNull(s)
             case _                  => None
           }).getOrElse(EmptyIsEmpty)
+          dataRedaction = DataRedaction(property).getOrElse(DataVisible)
```
blast-hardcheese: I think this feature will eventually need to be controllable via core interpreter, defining the default based on call site (possibly plugin configuration) such that for secure environments everything should be redacted by default

kelnos: Yeah, that makes sense.  Do you think we'd want to parse it out here, and then just give the core interp a chance to override it?

blast-hardcheese: Yeah -- I'm thinking a single `Term` that just `Target.pure(DataVisible)` defining the default. Eventually, it may consume the `config` parameters and use plugin configuration to differentiate, but I haven't thought that far ahead, mainly just interested in separating static defaults.

PR #317, opened 2019-06-03
Support BinarySchema for param/property types

kelnos:
This makes OpenAPI v3 binary support work for the case where you have a request body that uses a `$ref` to reference a defined schema.

This also removes the warning logging for `type: string`, `format: binary`; not sure why that was there in the first place since it's a perfectly valid combination, and, in OpenAPI v3, is the only correct way to declare stream of bytes (since they removed `type: file`).

**Contributing to Twilio**


kelnos:
Looks like there are some unexpected side-effects of this; getting this to work properly will require some refactoring of how we handle parameters.

PR #318, opened 2019-06-03
Use InputStream instead of File for file/binary types in Dropwizard

kelnos:
DW natively supports `InputStream`, but I had to write a shim for AsyncHttpClient.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientHelpers.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientHelpers.scala
@@ -2,6 +2,9 @@ package com.twilio.guardrail.generators.Java
 
 import com.github.javaparser.JavaParser
 import com.github.javaparser.ast.`type`.ClassOrInterfaceType
+import com.twilio.guardrail.{ SupportDefinition, Target }
+import com.twilio.guardrail.languages.JavaLanguage
+import com.twilio.guardrail.generators.syntax.Java.loadSupportDefinitionFromString
 
 object AsyncHttpClientHelpers {
   val DEFAULT_ASYNC_HTTP_CLIENT_CONFIG_BUILDER_TYPE: ClassOrInterfaceType = JavaParser.parseClassOrInterfaceType("DefaultAsyncHttpClientConfig.Builder")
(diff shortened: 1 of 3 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardGenerator.scala
```diff
@@ -10,7 +10,7 @@ import com.twilio.guardrail.terms.framework._
 object DropwizardGenerator {
   object FrameworkInterp extends (FrameworkTerm[JavaLanguage, ?] ~> Target) {
     def apply[T](term: FrameworkTerm[JavaLanguage, T]): Target[T] = term match {
-      case FileType(format)   => safeParseType(format.getOrElse("java.io.File"))
+      case FileType(format)   => safeParseType(format.getOrElse("java.io.InputStream"))
```
blast-hardcheese: so a concern here is what happens if there are more than one `type: file` parameters?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientHelpers.scala
```diff
@@ -12,5 +15,97 @@ object AsyncHttpClientHelpers {
...
+       *
+       * Unfortunately, AHC doesn't let us register custom part types, so we have
+       * to base this on ByteArrayPart, which means we don't actually stream the
+       * part; it has to be completely buffered into memory before sending.
+       *
+       * Unfortunately, this also means that the constructor can block the calling
+       * thread if the passed InputStream can block on read.
+       */
+      public class InputStreamPart extends ByteArrayPart {
+          private static final int BUFFER_SIZE = 1024 * 1024;
```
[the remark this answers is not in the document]
kelnos: Eh, probably not.  I'm actually digging in to find a better way to do this.  AHC appears to have InputStreamMultipartPart, even tho for some reason it doesn't have InputStreamPart.  The APIs for turning InputStreamMultipartPart into are request body are a little more involved than what I'm using now, but I think it's worth it.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardGenerator.scala
kelnos: Shouldn't be a problem, I don't think.  Why would it be?  They still get unique parameter names.  No reason the caller can't provide more than one.

blast-hardcheese: `curl example -F file1=@file1.iso -F file2=@file2.iso` and `curl example -F file2=@file2.iso -F file1=@file1.iso` are _very_ different.

kelnos: Not sure I understand what you think the issue is, or how using InputStream instead of File changes anything about this.

blast-hardcheese: Actually, even having two input streams overcomplicates the problem. It's simpler to just have one:
```scala
def doFoo(a: Boolean, b: InputStream)
```
So, this is fine:
1. `curl http://... -F a=false -F b=@file.iso`
1. `a` is a valid `Boolean`
1. `b` is streamed directly to `InputStream`, unbuffered
1. Everyone's happy

... but this would be not great
1. `curl http://... -F b=@file.iso -F a=false`
1. `b` is streamed directly to `InputStream`, buffered?
1. `a` may be a valid boolean, but we have no way to know yet
1. `doFoo` can't be called yet, as `a` has not been supplied yet and no mutable variables
1. server runs out of memory

kelnos: Oh, you're talking about server, not client.

Yeah, that's an issue, to be sure.  But this is also why servers have global max bytes per request settings.

blast-hardcheese: Oh, yes, very much so.

blast-hardcheese: So, we're OK with buffering in ram for arbitrary file sizes up until the global max bytes, and if anyone wants different behaviour it's up to them to ensure that only file content is supplied via post body?

kelnos: I'm not sure, honestly.  I don't think `File` is ideal, because allowing the server to spool out data to a temporary file comes with its own problems.  Buffering in RAM is also not great, but at least the user has control over the max body size allowed.  Multipart is just kinda crappy from a streaming perspective, unfortunately.

kelnos: It occurs to me that spec writers can also override this behavior with a `x-jvm-type: java.io.File` if they want.  I _do_ think that `InputStream` is a better default, though.

blast-hardcheese: What's the behaviour if they do override it with `java.io.File`, `mktemp()`? Sounds dangerous

kelnos: Not sure, will have to read the Jersey source to figure that out, but I assume it uses https://docs.oracle.com/javase/7/docs/api/java/io/File.html#createTempFile(java.lang.String,%20java.lang.String,%20java.io.File)

If they do this, then they should be aware of the risks.  This is the current behavior, anyway.

blast-hardcheese:
@kelnos Reconsidering this, I'm onboard with merging this if you have (or add) a check to see if multiple file upload bodies would be present for a single upload and either emitting a warning or failing the codegen. No sense in blocking useful stuff due to hard questions not being answered yet.

kelnos:
> multiple file upload bodies would be present for a single upload

Not sure I understand what you mean by this.

blast-hardcheese:
Hah! I meant multiple binary parameters for a single operation.

blast-hardcheese:
To summarize an offline discussion:

The problem we have here is if there are multiple `form` parameters, one or more of which are `file`, the caller has compete control over whether the file is delivered before any other parameters, causing the file to be required to be consumed before any other validation can occur.

As a result of this, the current strategy we're proceeding with is to emit a warning indicating that a route requires special attention and configuration.

Even if all it does is start the user on the journey of finding this conversation so they can weigh in, at least then they'd either learn or weigh in on the overall situation.

PR #319, opened 2019-06-05
Use NonEmptyList/NonEmptyMap for SecurityRequirements

kelnos:
We give an `Option[SecurityRequirements]` to terms that might need to do stuff with security.  The `SecurityRequirements` case class contains a list of requirements.  If that list is empty, there's no point in having ` SecurityRequirements` instance at all, so we guarantee that there will be at least one item in the list, and also guarantee that the item in the list is a non-empty map, since it's pointless for it to be empty.

**Contributing to Twilio**


kelnos:
@blast-hardcheese yeah, `NonEmptyMap` requires ` SortedMap` as input.  Not sure why, I guess maybe to ensure that `.head` always returns a consistent value?

PR #325, opened 2019-06-09
Deduplicate identical parameters across different contentTypes

### LINKED ISSUE #324: Server routes with more than one `consumes` duplicates parameters (2to3 shims bug)
In Swagger specifications, there was no differentiation between the different content types for HTTP bodies submitted to any given endpoint.

As a result, when converting swagger to OpenAPI 3.x, it's necessary to duplicate all parameters for each content type in the resulting specification.

Unfortunately, it seems as though [parameters are concatted irrespective of mediaType](https://github.com/twilio/guardrail/blob/4ef618d9ea9d213cd410e062313bd4252cc4e2fc/modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala#L106-L107) without deduplicating them.

This results in:
```
[error] .../PetClient.scala:45:98: name is already defined as value name
[error]   def updatePetWithForm(petId: Long, name: Option[String] = None, status: Option[String] = None, name: Option[String] = None, status: Option[String] = None, headers: List[HttpHeader] = Nil): EitherT[Future, Either[Throwable, HttpResponse], UpdatePetWithFormResponse] = {
[error]                                                                                                  ^
```

blast-hardcheese:
Resolves #324 

When multiple content types are specified in `consumes`, deduplicate the parameters and mark ones that don't explicitly occur across all media types as not required.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
@@ -3,6 +3,8 @@ package terms
 
 import cats.data.{ NonEmptyList, NonEmptyMap }
 import cats.free.Free
+import cats.data.State
+import cats.implicits._
 import cats.kernel.Order
 import com.twilio.guardrail.generators.{ ScalaParameter, ScalaParameters }
 import com.twilio.guardrail.languages.LA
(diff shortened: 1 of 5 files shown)


blast-hardcheese:
@kelnos I suspect that this will conflict with what you're doing in supporting distinct media types, but this is an attempt at squeezing slightly more mileage out of the content body shims.

blast-hardcheese:
Additionally, before moving this, I'll add a regression test that exercises this behaviour.

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
```diff
@@ -170,7 +175,26 @@ case class RouteMeta(path: String, method: HttpMethod, operation: Operation, sec
             p.setExtensions(Option(schema.getExtensions).getOrElse(new java.util.HashMap[String, Object]()))
             p
         }
+      })
+      .traverse[State[ParameterCountState, ?], Parameter] { p =>
```
kelnos: Out of curiosity, why the `?` here?

blast-hardcheese: `?` comes from https://github.com/typelevel/kind-projector
It's a simple macro that expands things of the shape `F[?]` into `({ type λ[A] = F[A] })#λ`, letting us write `.traverse[State[Foo, ?], Bar]` instead of `.traverse[({ type λ[A] = State[Foo, A] })#λ, Bar]`. Very handy.

blast-hardcheese: It's very common in FP libraries, fwiw.

kelnos: Ah, right.  I remembered the kind-projector thing, but not what it expanded to.

PR #326, opened 2019-06-09
Akka form data consumption

### LINKED ISSUE #312: akka-http: Substream Source cannot be materialized more than once
Given the following specification:
```yaml
swagger: 2.0
basePath: /v1
schemes:
  - http
paths:
  /pets:
    post:
      operationId: createPets
      x-scala-package: pets
      parameters:
        - name: name
          in: formData
          type: string
        - name: status
          in: formData
          type: string
#        - name: file  # Uncomment this block to switch between streaming `multipart: form/data` parsing and raw (broken) `formField` directives
#          in: formData
#          type: file
      responses:
        200:
          schema:
            type: string
```
I get the following behaviour:
```
blast$ curl localhost:8080/v1/pets -H 'application/x-www-form-urlencoded' --data "name=foo&status=bar"; echo
"(Some(foo),Some(bar))"
blast$ curl localhost:8080/v1/pets -F name=foo -F status=bar; echo
The form field 'status' was malformed:
Substream Source cannot be materialized more than once
```
This only seems to happen with `multipart/form-data` content type when there are no `type: file` parameters present.

What should probably happen is:
- the `type: file` multipart form-data processor should be expanded in scope to handle all `multipart: form-data` content types
- both the `x-www-form-urlencoded` and `multipart/form-data` are registered as `FromEntityUnmarshaller`s, provided:
  - `consumes` doesn't explicitly select one over another
  - there are no `type: file` parameters, in which case `x-www-form-urlencoded` is invalid.

blast-hardcheese:
Resolves #312 

- Switches to `Unmarshaller`s for `formData` consumption to support both `application/x-www-form-urlencoded` _and_ `multipart/form-data`, depending on what's in `consumes`
- Relies heavily on parameter deduplication (#325)
- If any `consumes` are `application/x-www-form-urlencoded`, all `file` parameters are forced to be optional.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
@@ -147,7 +147,8 @@ case class RouteMeta(path: String, method: HttpMethod, operation: Operation, sec
     type HashCode            = Int
     type Count               = Int
     type ParameterCountState = (Count, Map[HashCode, Count])
-    val contentTypes = Option(requestBody.getContent()).toList.flatMap(_.keySet.asScala.toList)
+    val contentTypes: List[RouteMeta.ContentType] =
+      Option(requestBody.getContent()).toList.flatMap(_.keySet.asScala.toList).flatMap(RouteMeta.ContentType.unapply)
     val ((maxCount, instances), ps) = Option(requestBody.getContent())
       .fold(List.empty[MediaType])(x => Option(x.values()).toList.flatMap(_.asScala))
       .flatMap({ mt =>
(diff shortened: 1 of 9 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
```diff
@@ -173,6 +174,11 @@ case class RouteMeta(path: String, method: HttpMethod, operation: Operation, sec
 
             p.setRequired(isRequired)
             p.setExtensions(Option(schema.getExtensions).getOrElse(new java.util.HashMap[String, Object]()))
+
+            if (Option(schema.getType()).exists(_ == "file") && contentTypes.contains("application/x-www-form-urlencoded")) {
```
kelnos: Should you use `RouteMeta.ContentType` for this to avoid string compares?

blast-hardcheese: Yep, good call

PR #337, opened 2019-07-08
Do not generate framework implicits when generating model only

hanny24:
When generating only model classes, current behavior is to generate also a file with implicits for given framework. This effectively means that generated code has a build dependency on http4s/akka-http/... even though it's not needed.

This PR changes that such that no file with framework specific implicits is generated when generating model classes only.

**Contributing to Twilio**


blast-hardcheese:
Fwiw, to keep the noise in the git history down, I'm going to rebase your changes -- if you wouldn't mind opening future PRs off master, that would be appreciated!

PR #344, opened 2019-07-12
WIP - an array of integers in query parameters is generated as strings

kubum:
Hi everyone,

I would like to ask what does prevent guardrail to generate a parameter with the type `Option[Iterable[Int]]` from an array of integers in the `server.yaml` for akka-http?

There is a test case in the branch, I would be happy if you can point me out where does it get confused and still generates `Option[Iterable[String]]`?

Not sure if the generated code in the branch helps to understand the issue better.

Thank you!


blast-hardcheese:
Hey @kubum! Could you try to generate your servers using `./cli.sh` off the current master? #292 may have resolved this

kubum:
Hi @blast-hardcheese!

The example `issueX.yaml` still has generated strings:

```scala
parameter(Symbol("keys").as[String].*).map(xs => Option(xs).filterNot(_.isEmpty)
```

Steps to reproduce:

```bash
$ ./cli.sh --server --specPath modules/sample/src/main/resources/issues/issueX.yaml --outputPath modules/sample/src/main/scala --packageName issues.issueX
```

The issue yaml contains:

```yaml
        - name: keys
          in: query
          type: array
          items:
            type: integer
            format: int32
```

Do you have an idea where it gets confused while generating the routes? I wonder if I can help.

Thank you!

blast-hardcheese:
@kubum My apologies for not responding sooner! I've been unfortunately in and out over the past few weeks, and some of these issues have gotten away from me.

Your `runExample` task was very helpful, by the way. In the future, you could also leave a comment to the effect of "Please run `cli --server --specPath modules/sample/src/main/resources/issues/issueX.yaml --outputPath modules/sample/src/main/scala --packageName issues.issueX`", to save yourself the trouble, but again, that was very useful.

I think I may have found the problem:

https://github.com/twilio/guardrail/blob/master/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaGenerator.scala#L121

Tracing back, you seem to have hit the motherlode of legacy cruft.

```diff
diff --git a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaParameter.scala b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaParameter.scala
index 871802187..a560fb9c3 100644
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaParameter.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaParameter.scala
@@ -70,11 +70,7 @@ object ScalaParameter {
 
       def resolveParam(param: Parameter, typeFetcher: Parameter => Free[F, String]): Free[F, ResolvedType[L]] =
         for {
-          tpeName <- typeFetcher(param)
-          fmt = Option(param.getSchema).flatMap(s => Option(s.getFormat))
-          customTypeName <- SwaggerUtil.customTypeName(param)
-          res <- (SwaggerUtil.typeName[L, F](tpeName, Option(param.format()), customTypeName), getDefault(tpeName, fmt, param))
-            .mapN(SwaggerUtil.Resolved[L](_, None, _, Some(tpeName), fmt))
+          res <- Option(param.getSchema).fold(fallbackParameterHandler(param))(SwaggerUtil.propMeta[L, F](_))
         } yield res
 
       def paramHasRefSchema(p: Parameter): Boolean = Option(p.getSchema).exists(s => Option(s.get$ref()).nonEmpty)
```

(For context, I rebased your branch off master, so this patch may not be 100% accurate unless you do as well)

Given that patch, your test case passes, but a number of other tests start failing. This turns out to be _another_ bug, this time in how we've been handling the various purpose-built `Schema` subclasses in `swagger-parser`:

https://github.com/twilio/guardrail/blob/master/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala#L331

It turns out that reflecting on these types is unsound for Swagger 2to3 converted specifications, as the converter just creates the raw supertype every time:

https://github.com/swagger-api/swagger-parser/blob/master/modules/swagger-parser-v2-converter/src/main/java/io/swagger/v3/parser/converter/SwaggerConverter.java#L695-L697

As a result of _this_, what we'll likely have to do is build ontop of the patch I pasted above, then come up with a way to make the syntax not suck for handling both the ideal case of having thees purpose-built `Schema` subclasses as well as handling the fallback type with (most of) the same logic.

---

Some initial thoughts, including loads of boilerplate:

```scala
object BooleanSchema {
  type RawType = String
  type RawFormat = String
  type Specialized = BooleanSchema
  def unapply(schema: Schema[_]): Option[(RawType, Option[RawFormat], Option[Specialized])] = ???
}

...

  case schema@BooleanSchema(rawType, rawFormat, specialized) =>
    for {
      customTpeName <- customTypeName(schema)
      res <- (
        typeName[L, F](rawType, None, customTpeName),
        specialized.flatTraverse(b => Default(b).extract[Boolean].traverse(litBoolean(_)))
      ).mapN(Resolved[L](_, None, _, Some(rawType), rawFormat))
    } yield res
```

where `specialized` is only present if we found it upon reflection.

---

Whew. Hopefully this provides some context to the problem you're facing. The same bug exists on the Java side of the world as well, so any fixes we make here will likely need to be made there as well.

If you feel so inclined to proceed with any of this, I'm available to answer questions.

If not, I'll see when I can get to it -- this will be a significant number of lines removed, combined with consolidation and fixing a major bug in how 2to3 is handled, so I'm very interested, but I'm also not overburdened with free time as of late. Your assistance would almost certainly get to a resolution quicker.

blast-hardcheese:
This has been resolved in #1407. Sorry for the delay in resolution 🙂

Issue #94, opened 2018-09-10
Server readOnly properties are applied to responses instead of requests


kthornberrysc:
Hi, 
See example below with the readonly property TestObject.bar:

```
---
  swagger: "2.0"
  schemes:
    - http
  consumes:
    - application/json
  produces:
    - application/json
  paths:
    /test:
      get:
        operationId: testGet
        responses:
          200:
            schema:
              $ref: "#/definitions/TestObject"
      post:
        operationId: testSet
        parameters:
          - name: body
            in: body
            schema:
              $ref: "#/definitions/TestObject"
        responses:
          200:
            description: OK
  definitions:
    TestObject:
      type: object
      properties:
        foo:
          type: string
        bar:
          type: string
          readOnly: true
```

The server code generated from this removes the readOnly properties from responses when encoding as JSON (in definitions/TestObject.scala), and does not perform any validation on incoming request objects. Is this the wrong way round? As I understand it, readOnly properties are allowed on response objects and forbidden on request objects (https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#schemaReadOnly).

I guess it's straightforward to remove the code that removes readOnly properties from encoded responses, but I'm not sure how to add validation and error reporting if a request contains one of the properties.

Any advice much appreciated, thanks!

blast-hardcheese:
Ah, this is actually because `readOnly` was written for clients first, and was never revisited for servers. Good catch!

PR #346, opened 2019-07-15
Issure #94: Add server/client aware context for circe encoders/decoders

dtennander:
Hello, and thank you for an awesome tool!

I have run into #94 so I thought I might give it a try to figure out a solution.

Right now my solution would include creating and using an implicit `EncodingContext` (name in development) that would specify if the definition is used on server side or client side and thus generate two different results.

The generated servers and clients would then define their respective `EncodingContext`.
Do you think they should be defined the server/client package or should they maybe be defined in the protocol package and be imported into the server/client respectively?

Client encoders should exclude readOnly and include writeOnly were as 
Server encoders should include readOnly and exclude WriteOnly.

This would be a bit of a rewrite but would allow guardrails circe encodings to have different behavior on client and server side.

WDYT? 

### DIFF
--- a/modules/codegen/src/test/scala/core/issues/Issue94.scala
+++ b/modules/codegen/src/test/scala/core/issues/Issue94.scala
@@ -1,0 +1,67 @@
+package core.issues
+
+import com.twilio.guardrail.generators.Http4s
+import com.twilio.guardrail.{ClassDefinition, Context, ProtocolDefinitions, StaticDefns}
+import org.scalatest.{FunSpec, Matchers}
+import support.SwaggerSpecRunner
+
+import scala.meta._
+
+class Issue94 extends FunSpec with Matchers with SwaggerSpecRunner {
+
+  describe("ReadOnly parameter should be returned from server") {
+
+    val swagger: String = """
+      | openapi: "3.0.0"
+      | info:
+      |   title: Server should write out readOnly parameters.
+      |   version: 1.0.0
+      | components:
+      |   schemas:
+      |     ObjectWithWriteOnly:
+      |       type: object
+      |       properties:
+      |         prop:
+      |           type: string
+      |           writeOnly: true
+      |     ObjectWithReadOnly:
+      |       type: object
+      |       properties:
+      |         prop:
+      |           type: string
+      |           readOnly: true""".stripMargin
+
+    val (
+      ProtocolDefinitions(
+      ClassDefinition(_, _, _, StaticDefns(_, _, writeOnlyEncoder :: _ :: Nil), _) ::
+      ClassDefinition(_, _, _, StaticDefns(_, _, readOnlyEncoder :: _ :: Nil), _) :: Nil, _, _, packageObjectContents
+      ), // ProtocolDefinitions
+      _, // clients
+      _, // Servers
+    ) = runSwaggerSpec(swagger)(Context.empty, Http4s)
+
+    it("Ensure correct code is generated") {
+
+      val expectedReadOnlyResult =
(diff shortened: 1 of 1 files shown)


dtennander:
Creating a draft PR to get early feedback on the solution before implementing it. Hope this is an allowed process.

blast-hardcheese:
@dtennander I think this matches what I'd expect, as far as current considerations go. Additionally, we could utilize this as an opportunity to filter out parameters before even rendering instead of `filterKeys` after encoding.

One of the bigger considerations that may not be covered here is enforcing that `readOnly` and `writeOnly` constructors are `Option`al:

|            | Received object through server | Creating object for client |
| ------ | ----- | ----- |
| Read Only  | Must not be optional           | Must be optional           |
| Write Only | Must be optional               | Must not be optional       |

So, to flip it around slightly:

```scala
class ContextAwareCodec[A](val readOnly: Encoder[A], val writeOnly: Encoder[A])
object ContextAwareCodec {
  def instance[A](implicit codec: ContextAwareCodec[A], ctx: CodecContext): Encoder[A] = ctx match {
    case CodecContext.ReadOnly => codec.readOnly
    case CodecContext.WriteOnly => codec.writeOnly
  }
}

implicit def encodeObjectWithReadOnly: ContextAwareCodec[ObjectWithReadOnly] = new ContextAwareCodec[ObjectWithReadOnly](
  new ObjectEncoder[ObjectWithReadOnly] { final def encodeObject(a: ObjectWithReadOnly): JsonObject = JsonObject.fromIterable(Vector(("prop", a.prop.asJson))) },
  new ObjectEncoder[ObjectWithReadOnly] { final def encodeObject(a: ObjectWithReadOnly): JsonObject = JsonObject.fromIterable(Vector()) }
)
```

the unpleasant thing is not being able to have different constructors for required vs not required for the different contexts, so everything must become `Option`al.

Thoughts?

dtennander:
@blast-hardcheese: Nice I like the changes you made!

Yeah... The `Option` thing is a pickle. Should we document this requirement and/or fail compilation if both required and writeOnly/readOnly is set `true`? Or should we maybe ignore the required flag if writeOnly/readOnly is set?

blast-hardcheese:
@dtennander Sorry for dropping the ball on this! I've been in and out over the past few weeks, if that's any explanation.

I really don't think there's a way to do a compile-time-safe version of this without resorting to something like shapeless and generating all definitions as record types, which would permit the entire nested record to be typechecked at once at every call site.

I'm going to kick this over to one of my coworkers who may have an opinion on this, but to summarize my current thoughts (after coming back to the problem):

- We could make `respond.Xyz` return `Either[Pointer, FooResponse]`, which taints response codes that have these read/write-only fields, forcing users to handle the error (This assumes that any parameter tainted with read/write only is forced to be Optional, which removes type safety)
- We could have specific builders and a phantom parameter that signifies use, like `Foo.writeOnly(1,2,3): Foo with WriteOnly` or `: Foo[A <: ReadWriteEither]` where `ReadWriteEither` would track usage intent and be absolutely horrendous to use for most users, and a converter that would explicitly keep track of the necessary parameters that are different in the other variant.
- We could explicitly split the definitions, permanently tainting them with one behaviour or the other (this is actually currently done, biased from the client's perspective)

None of these really stand out as being excellent to me, so I'm looking for some other opinions on the problem.

blast-hardcheese:
Closing this out due to stale, though I'm sympathetic to this ask. Hopefully it gets picked up eventually, in a way that doesn't burden the user.

PR #355, opened 2019-07-24
Support arbitrary primitive types and enums as discriminators

kelnos:
For Jackson, we assume that polymorphic POJO discriminators are of type string, which is of course a bit limiting.  Ideally they can be any type, including enums.

Unfortunately, it's not possible for us to support any arbitrary type with how we handle discriminators right now, but we can at least explicitly support all primitive types, as well as enums.

Stuff I'm not happy about:

1. To figure out which type to use, we "reverse engineer" the parsed type that the core gives us.
2. If the type the core gives us isn't a recognized primitive type, or one of the types explicitly listed as being not supported, we just assume it's an enum type, which of course it may not be.  This will result in generated code that won't compile.  It would be better to present an error at generation time, but on the plus side, at least this doesn't give you runtime errors.

This already works out of the box with Circe (sorta; I think number types might get mis-JSONed as strings) because of differences in how poloymorphism are handled, as well as the fact that Scala has typeclasses, so as long as you have an `Encoder` in scope, you can convert anything into a JSON value.  You can't really make things work that way with Java/Jackson.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
@@ -5,8 +5,9 @@ package Java
 import _root_.io.swagger.v3.oas.models.media._
 import cats.data.NonEmptyList
 import cats.implicits._
+import cats.instances.map
 import cats.~>
-import com.github.javaparser.ast.`type`.{ PrimitiveType, Type, UnknownType }
+import com.github.javaparser.ast.`type`.{ ClassOrInterfaceType, PrimitiveType, Type, UnknownType }
 import com.twilio.guardrail.Discriminator
 import com.twilio.guardrail.extract.{ DataRedaction, Default, EmptyValueIsNull }
 import com.twilio.guardrail.generators.syntax.Java._
@@ -20,17 +21,22 @@ import com.github.javaparser.ast.stmt._
 import com.github.javaparser.ast.Modifier.{ ABSTRACT, FINAL, PRIVATE, PROTECTED, PUBLIC, STATIC }
 import com.github.javaparser.ast.body._
 import com.github.javaparser.ast.expr._
+import java.math.BigInteger
 import java.util
+import java.util.Locale
 import scala.language.existentials
 import scala.util.Try
 
 object JacksonGenerator {
-  private val BUILDER_TYPE = JavaParser.parseClassOrInterfaceType("Builder")
+  private val BUILDER_TYPE        = JavaParser.parseClassOrInterfaceType("Builder")
+  private val BIG_INTEGER_FQ_TYPE = JavaParser.parseClassOrInterfaceType("java.math.BigInteger")
+  private val BIG_DECIMAL_FQ_TYPE = JavaParser.parseClassOrInterfaceType("java.math.BigDecimal")
 
   private case class ParameterTerm(propertyName: String,
                                    parameterName: String,
                                    fieldType: Type,
                                    parameterType: Type,
+                                   rawType: RawParameterType,
                                    defaultValue: Option[Expression],
                                    dataRedacted: RedactionBehaviour)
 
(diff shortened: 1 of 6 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -296,13 +300,55 @@ object JacksonGenerator {
...
+                t => Target.raiseError[T](s"Unable to parse '$discriminatorValue' as $friendlyName: ${t.getMessage}"),
+                Target.pure[T]
+              )
+
+            term.fieldType.simpleName
+              .fold(
+                Target.raiseError[Expression](s"Field type for discriminator ${term.propertyName} is not valid for this type")
+              )({
+                case "String"              => Target.pure[Expression](new StringLiteralExpr(discriminatorValue))
+                case "boolean" | "Boolean" => parseLiteral(_.toBoolean, "boolean").map(new BooleanLiteralExpr(_))
```
blast-hardcheese: Why is it possible to get both cases here? Unboxed vs boxed?

kelnos: Yep.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -296,13 +301,78 @@ object JacksonGenerator {
...
+              case Some(tpe @ "boolean") => parseLiteral(x => new BooleanLiteralExpr(x.toBoolean), tpe)
+              case Some(tpe @ "integer") =>
+                term.rawType.format match {
+                  case Some(fmt @ "int32") => parseLiteral(x => new IntegerLiteralExpr(x.toInt), fmt)
+                  case Some(fmt @ "int64") => parseLiteral(x => new LongLiteralExpr(x.toLong), fmt)
+                  case Some(fmt) =>
+                    Target.raiseError[Expression](s"Unsupported discriminator type '$tpe' with format '$fmt' for property '${term.propertyName}'")
+                  case None =>
+                    parseLiteral({ x =>
+                      new BigInteger(x); new ObjectCreationExpr(null, BIG_INTEGER_FQ_TYPE, new NodeList(new StringLiteralExpr(discriminatorValue)))
```
blast-hardcheese: oh jeez. Could we not serialize the `BigInteger` back into something that can be stuffed into `ObjectCreationExpr`, lest some linter start calling it out as a non-unit expression?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -296,13 +301,78 @@ object JacksonGenerator {
...
+                }
+              case Some(tpe @ "number") =>
+                term.rawType.format match {
+                  case Some(fmt @ "float")  => parseLiteral(x => new DoubleLiteralExpr(x.toFloat), fmt)
+                  case Some(fmt @ "double") => parseLiteral(x => new DoubleLiteralExpr(x.toDouble), fmt)
+                  case Some(fmt) =>
+                    Target.raiseError[Expression](s"Unsupported discriminator type '$tpe' with format '$fmt' for property '${term.propertyName}'")
+                  case None =>
+                    parseLiteral({ x =>
+                      new java.math.BigDecimal(x); new ObjectCreationExpr(null, BIG_DECIMAL_FQ_TYPE, new NodeList(new StringLiteralExpr(discriminatorValue)))
```
blast-hardcheese: same

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -296,13 +301,78 @@ object JacksonGenerator {
...
+                    }, "BigDecimal")
+                }
+              case Some(tpe) =>
+                Target.raiseError[Expression](s"Unsupported discriminator type '$tpe' for property '${term.propertyName}'")
+              case None =>
+                term.fieldType match {
+                  case cls: ClassOrInterfaceType =>
+                    // hopefully it's an enum type; nothing else really makes sense here
```
blast-hardcheese: 🤞

modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
```diff
@@ -52,14 +52,23 @@ object Java {
...
     def name: Option[String] =
       tpe match {
         case cls: ClassOrInterfaceType =>
           Some(cls.getScope.asScala.fold("")(_.getName.asString + ".") + cls.getNameAsString)
-        case _ => None
+        case pt: PrimitiveType => Option(pt.asString)
+        case _                 => None
+      }
+
+    def simpleName: Option[String] =
```
blast-hardcheese: no longer seems used

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
kelnos: I guess could do `new ObjectCreationExpr(..., new BigInteger(x).toString())`?  in theory it shouldn't lose precision, since that's the point of `BigInteger`, yeah?

kelnos: Did this ^. Seems to be ok, checked some numbers with very very very very very large numbers both to the right and left of the decimal, as well as small, and there's no loss in precision.

PR #358, opened 2019-07-31
Safety and exhaustivity refactors

blast-hardcheese:
This PR largely applies to akka-http, though the intent is as more functionality gets fleshed out, more of the requirements will move to core and trickle out into the rest of the frameworks/languages.

- Starting with changing `consumes` to `NonEmptyList` (flipped for the akka-http client generator)
- Two new exceptions in cases where we don't have either `Free` or `Target` (these will continue to flag as warnings as time goes on, and will be resolved as we go back to reduce the number of warnings in the build, moving towards `-Wall`)
- Removing some NPEs in the 2to3 shim (we were doing stuff like `Option(getSchema.getType)`)

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
@@ -155,7 +155,7 @@ object AkkaHttpClientGenerator {
                   formDataParams: Option[Term],
                   headerParams: Term,
                   responses: Responses[ScalaLanguage],
-                  produces: Seq[RouteMeta.ContentType],
+                  produces: NonEmptyList[RouteMeta.ContentType],
                   consumes: Seq[RouteMeta.ContentType],
                   tracing: Boolean)(tracingArgsPre: List[ScalaParameter[ScalaLanguage]],
                                     tracingArgsPost: List[ScalaParameter[ScalaLanguage]],
@@ -259,7 +259,9 @@ object AkkaHttpClientGenerator {
           // Placeholder for when more functions get logging
           _ <- Target.pure(())
 
-          produces = operation.produces.toList.flatMap(RouteMeta.ContentType.unapply(_))
+          produces = NonEmptyList
+            .fromList(operation.produces.toList.flatMap(RouteMeta.ContentType.unapply(_)))
+            .getOrElse(NonEmptyList.one(RouteMeta.ApplicationJson))
           consumes = operation.consumes.toList.flatMap(RouteMeta.ContentType.unapply(_))
 
           headerArgs = parameters.headerParams
@@ -405,10 +407,10 @@ object AkkaHttpClientGenerator {
         Target.pure(NonEmptyList(Right(client), Nil))
     }
 
-    def generateCodecs(methodName: String, responses: Responses[ScalaLanguage], produces: Seq[RouteMeta.ContentType]): List[Defn.Val] =
+    def generateCodecs(methodName: String, responses: Responses[ScalaLanguage], produces: NonEmptyList[RouteMeta.ContentType]): List[Defn.Val] =
       generateDecoders(methodName, responses, produces)
 
-    def generateDecoders(methodName: String, responses: Responses[ScalaLanguage], produces: Seq[RouteMeta.ContentType]): List[Defn.Val] =
+    def generateDecoders(methodName: String, responses: Responses[ScalaLanguage], produces: NonEmptyList[RouteMeta.ContentType]): List[Defn.Val] =
       for {
         resp <- responses.value
         tpe  <- resp.value.map(_._1).toList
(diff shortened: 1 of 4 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
```diff
@@ -259,7 +259,9 @@ object AkkaHttpClientGenerator {
           // Placeholder for when more functions get logging
           _ <- Target.pure(())
 
-          produces = operation.produces.toList.flatMap(RouteMeta.ContentType.unapply(_))
+          produces = NonEmptyList
+            .fromList(operation.produces.toList.flatMap(RouteMeta.ContentType.unapply(_)))
+            .getOrElse(NonEmptyList.one(RouteMeta.ApplicationJson))
```
kelnos: No change needed, just something I've had on my mind on and off: should we be basing a default consumes/produces based on the type of the body?  Like if it's `type: string`, maybe text/plain.  If it's a `$ref` to some `type: object`, JSON... etc.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
```diff
@@ -522,88 +522,100 @@ object AkkaHttpServerGenerator {
...
+                (List(fru), NonEmptyList.one(unmarshallerTerm))
               }
-          """), Option(unmarshallerTerm))
-            } else (List.empty, None)
+
+              case RouteMeta.ApplicationJson => throw new Exception(s"Unable to generate unmarshaller for application/json")
+
+              case RouteMeta.OctetStream => throw new Exception(s"Unable to generate unmarshaller for application/octet-stream")
+
+              case RouteMeta.TextPlain => throw new Exception(s"Unable to generate unmarshaller for text/plain")
```
kelnos: Wartremover is gonna be pissed at you for these ;)

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
```diff
@@ -71,17 +72,29 @@ case class SecurityRequirements(requirements: NonEmptyList[NonEmptyMap[String, S
                                 location: SecurityRequirements.Location)
 
 case class RouteMeta(path: String, method: HttpMethod, operation: Operation, securityRequirements: Option[SecurityRequirements]) {
+  object MediaType {
+    def unapply(value: MediaType): Option[(Option[Schema[_]], Option[Map[String, Encoding]], Option[Map[String, Object]])] = {
+      val schema: Option[Schema[_]] = Option(value.getSchema)
+      val encoding                  = Option(value.getEncoding()).map(_.asScala.toMap)
+      val extensions                = Option(value.getExtensions()).map(_.asScala.toMap)
+      Option((schema, encoding, extensions))
```
kelnos: Is it worth defining a case class to make this 3-tuple clearer?

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
```diff
@@ -103,12 +117,11 @@ case class RouteMeta(path: String, method: HttpMethod, operation: Operation, sec
     val content = for {
       content <- Option(requestBody.getContent)
       mt      <- content.values().asScala.headOption
-      ref     <- Option(mt.getSchema.get$ref())
+      schema  <- Option(mt.getSchema)
+      ref     <- Option(schema.get$ref())
```
kelnos: Yay fewer NPEs!

modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala
blast-hardcheese: :hide:

modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpClientGenerator.scala
blast-hardcheese: I didn't do it here, but I'm leaning more towards just straight up failing unless the produces/consumes is supplied. The use case I'm weighing against is if the user doesn't have access to modify the OpenAPI specification file they are generating from.

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala
blast-hardcheese: It wouldn't make it clearer, the only way it's even possible to be used is in the local scope, and it's just used in the `case MediaType(_, _, _) =>`. Changing it to `case MediaType(MediaTypeHolder(_, _, _)) =>` isn't clearer, imo.

That would be a useful technique to create safe extractors for OpenAPI types, but I don't believe `unapply` is the place to do that conversion -- it's too magical, and introduces a completely idiosyncratic pattern.

It may be useful for readability in the future, but right now it would be too big of a shift, both stylistically and mentally.

blast-hardcheese: > the only way it's even possible to be used is in the local scope

a quick follow-up: I found an edge case with companion objects defined for classes that don't have them and are defined in different packages. I tried defining this object in a shared location and importing it, but I got import conflicts. I think this is an edge case in scalac, but I don't care to investigate currently.

PR #359, opened 2019-08-01
Unmarshaller shuffle redux

### LINKED ISSUE #357: Extend unmarshaller shuffle to urlencoded parameters
As part of #292, it seems as though server-parsed urlencoded parameters were left out.

- [ ] Create tests for all field types, crossed against all encoding types, to verify that JSON literals are correctly treated as their stringy equivalents
- [ ] Fix whatever handlers are not correctly decoding these values.

This effectively boils down to finding anywhere we do `jawn.parse(...).orElse(Json.fromString(...))` and choosing one or the other.

blast-hardcheese:
Depends on #358 
Resolves #357 

#292 didn't fully resolve the problem, as query string parameters still used the old-style fallback parser.

I've added a test for it now, as well as deleting the dangerous unmarshaller entirely.

**Contributing to Twilio**

### DIFF
--- a/modules/sample/src/main/resources/issues/issue357.yaml
+++ b/modules/sample/src/main/resources/issues/issue357.yaml
@@ -1,0 +1,73 @@
+swagger: '2.0'
+host: localhost:1234
+schemes:
+- http
+paths:
+  /{path}:
+    delete:
+      operationId: deleteFoo
+      parameters:
+      - name: path
+        in: path
+        type: string
+        required: true
+      - name: query
+        in: query
+        type: string
+        required: true
+      - name: form
+        in: formData
+        type: string
+        required: true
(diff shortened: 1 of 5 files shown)


modules/sample/src/main/resources/issues/issue357.yaml
```diff
@@ -0,0 +1,73 @@
...
+      parameters:
+      - name: path
+        in: path
+        type: string
+        required: true
+      - name: query
+        in: query
+        type: string
+        required: true
+      - name: hoopdoop
```
kelnos: hoopdoop!

blast-hardcheese:
Diff from force push: https://github.com/twilio/guardrail/compare/e7e36bc..e452b46

PR #363, opened 2019-08-05
Remove generated io.circe.java8 imports and add documentation

### LINKED ISSUE #361: Generated code uses circe-java8 which does not exist on latest Circe version (HTTP4S, Scala 2.13)
I was trying Guardrail in combination with a Scala 2.13 http4s project, but I ran into an issue. Generated Guardrail code uses circe-java8, which is removed in Circe 0.12.0-M3 (see [release notes](https://github.com/circe/circe/releases/tag/v0.12.0-M3)). This is also the lowest version that supports Scala 2.13. This means that code generated by Guardrail will not compile on any Circe version higher than 0.12.0-M3 as it cannot find the package. 

Current behaviour:

```
[info] Compiling 8 Scala sources to /myFolder/guardrail-http4s-example/target/scala-2.13/classes ...
[error] /myFolder/guardrail-http4s-example/target/scala-2.13/src_managed/main/com/example/quickstart/endpoints/definitions/package.scala:7:17: object java8 is not a member of package io.circe
[error] import io.circe.java8.time._
[error]                 ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

Expected behaviour:

Guardrail does not give compile errors due to generated code that is not compatible with the latest Circe version.

Used versions:
- Guardrail: 0.52.1
- Circe: 0.12.0-M4
- Http4s: 0.21.0-M2

Example project: https://github.com/hugo-vrijswijk/guardrail-http4s-example

I would be interested in helping with a fix, if I can be pointed in the right direction. Perhaps remove the import when the Circe version 0.12.0-M2 or higher?

hugo-vrijswijk:
Fixes #361 

I've removed the imports and added some documentation. There could be more places where documentation needs to be updated.

**Contributing to Twilio**

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -23,6 +23,7 @@ guardrail version | akka-http | akka-stream | circe  | cats  | Notes
  "                | 10.0.15   |             |  "     |  "    |
  "                | 10.1.0    | 2.3.5       |  "     |  "    |
  "                | 10.1.7    | 2.5.19      |  "     |  "    |
+0.54.0            |  "        |  "          | 0.12.1 |  "    | For Scala 2.11, use --module circe-java8
 
 ### http4s
 
@@ -31,4 +32,5 @@ guardrail version | http4s | circe-core | cats  | Notes
 0.34.0 ⚠          | 0.18.0 | 0.9.0      | 1.0.1 | Only clients are supported
 0.39.0            | 0.18.0 |  "         |  "    | First server release
 0.41.1            | 0.19.0 | 0.10.0     | 1.4.0 |
-0.5x.x            |  "     |  "         |  "    | For Scala 2.11, Add `io.circe:circe-java8` to your dependencies and add `"io.circe.java8.time._"` to your guardrail imports configuration.
+0.51.0            | 0.20.0 |  "         |  "    |
+0.54.0            |   "    | 0.12.1     |  "    | For Scala 2.11, use --module circe-java8
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -31,3 +31,4 @@ guardrail version | http4s | circe-core | cats  | Notes
 0.34.0 ⚠          | 0.18.0 | 0.9.0      | 1.0.1 | Only clients are supported
 0.39.0            | 0.18.0 |  "         |  "    | First server release
 0.41.1            | 0.19.0 | 0.10.0     | 1.4.0 |
+0.5x.x            |  "     |  "         |  "    | For Scala 2.11, Add `io.circe:circe-java8` to your dependencies and add `"io.circe.java8.time._"` to your guardrail imports configuration.
--- a/modules/sample-akkaHttp/src/test/scala/core/issues/Issue148.scala
+++ b/modules/sample-akkaHttp/src/test/scala/core/issues/Issue148.scala
@@ -11,7 +11,7 @@ import org.scalatest.time.SpanSugar._
 import org.scalatest.{ EitherValues, FunSuite, Matchers }
 import scala.concurrent.Future
 import io.circe._
-import _root_.jawn.IncompleteParseException
+import org.typelevel.jawn.IncompleteParseException
 
 /** Changes
   *
@@ -153,7 +153,7 @@ class Issue148Suite extends FunSuite with Matchers with EitherValues with ScalaF
           .toEntity
       ) ~> route ~> check {
       rejection match {
-        case MalformedFormFieldRejection("foo", message, _) => message shouldBe "expected json value got b (line 1, column 1)"
+        case MalformedFormFieldRejection("foo", message, _) => message shouldBe "expected json value got 'blep' (line 1, column 1)"
       }
     }
 
@@ -171,7 +171,7 @@ class Issue148Suite extends FunSuite with Matchers with EitherValues with ScalaF
       ) ~> route ~> check {
       rejection match {
         case MalformedFormFieldRejection("foo", message, _) =>
-          message shouldBe "expected json value got b (line 1, column 1)"
+          message shouldBe "expected json value got 'blep' (line 1, column 1)"
       }
     }
 
@@ -190,7 +190,7 @@ class Issue148Suite extends FunSuite with Matchers with EitherValues with ScalaF
           .toEntity
       ) ~> route ~> check {
       rejection match {
-        case MalformedFormFieldRejection("bar", message, _) => message shouldBe "expected json value got b (line 1, column 1)"
+        case MalformedFormFieldRejection("bar", message, _) => message shouldBe "expected json value got 'blep' (line 1, column 1)"
       }
     }
 
@@ -210,7 +210,7 @@ class Issue148Suite extends FunSuite with Matchers with EitherValues with ScalaF
           .toEntity
       ) ~> route ~> check {
       rejection match {
-        case MalformedFormFieldRejection("bar", message, _) => message shouldBe "expected json value got b (line 1, column 1)"
+        case MalformedFormFieldRejection("bar", message, _) => message shouldBe "expected json value got 'blep' (line 1, column 1)"
       }
     }
   }
--- a/build.sbt
+++ b/build.sbt
@@ -335,7 +335,7 @@ lazy val dropwizardSample = (project in file("modules/sample-dropwizard"))
       "org.mockito"                %% "mockito-scala"          % "1.2.0"            % Test,
       "com.github.tomakehurst"     %  "wiremock"               % "1.57"             % Test,
       "io.dropwizard"              %  "dropwizard-testing"     % dropwizardVersion  % Test,
-			"org.glassfish.jersey.test-framework.providers" % "jersey-test-framework-provider-grizzly2" % jerseyVersion % Test
+      "org.glassfish.jersey.test-framework.providers" % "jersey-test-framework-provider-grizzly2" % jerseyVersion % Test
     ),
     unmanagedSourceDirectories in Compile += baseDirectory.value / "target" / "generated",
     crossPaths := false,  // strangely needed to get the JUnit tests to run at all
--- a/build.sbt
+++ b/build.sbt
@@ -13,7 +13,7 @@ crossScalaVersions in ThisBuild := Seq("2.12.10")
 val akkaVersion          = "10.0.14"
 val catsVersion          = "1.6.0"
 val catsEffectVersion    = "1.0.0"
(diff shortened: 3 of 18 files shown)


hugo-vrijswijk:
The build seems to be failing with the following error: 

```
[error] /home/travis/build/twilio/guardrail/modules/sample-endpoints/target/generated/additionalProperties/client/endpoints/definitions/package.scala:11:57: could not find implicit value for parameter instance: io.circe.Decoder[java.time.Instant]
```

From what I can tell, the `sample-endpoints` project is a ScalaJS project (it has `.enablePlugins(ScalaJSPlugin)`, so it makes sense to me that it cannot compile. I am not sure what the way forward here is.

modules/sample-akkaHttp/src/test/scala/core/issues/Issue148.scala
```diff
@@ -153,7 +153,7 @@ class Issue148Suite extends FunSuite with Matchers with EitherValues with ScalaF
           .toEntity
       ) ~> route ~> check {
       rejection match {
-        case MalformedFormFieldRejection("foo", message, _) => message shouldBe "expected json value got b (line 1, column 1)"
+        case MalformedFormFieldRejection("foo", message, _) => message shouldBe "expected json value got 'blep' (line 1, column 1)"
```
blast-hardcheese: This is somewhat worrisome; Do you have an context on why this changed?

hugo-vrijswijk: I think it's the change in Circe version

hugo-vrijswijk:
Is anyone available to have a look at this PR? I'd like to continue work on it and have Guardrail compatible with 2.13, but I am unsure how to continue.

blast-hardcheese:
It does appear that Travis is failing, investigating...

blast-hardcheese:
My assessment is that the most straightforward way to move forward here is to fork the scalajs dependency on circe to stay pinned to the older version, with the intent of further decoupling this functionality -- no reason to depend on an entire implementation of java8 time (instances, formatters, etc) if we don't actually use them.

If you'd like to review what I've pushed, it would be good to have another perspective of the direction I'm thinking of going here. If you think this is acceptable, I'm comfortable merging this (provided that the tests pass)

blast-hardcheese:
Actually, it seems as though this will become much easier in 0.12.x, as [the instances have been merged into `shared`](https://github.com/circe/circe/blob/v0.12.0-RC4/modules/core/shared/src/main/scala/io/circe/Encoder.scala), it's just 0.11.x that suffers from this unpleasantness.

blast-hardcheese:
Merging to resolve this for now. Past 0.12.x we can likely reverse this.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Endpoints.scala
```diff
@@ -11,7 +11,18 @@ import ScalaGenerator._
...
 
 object Endpoints extends (CodegenApplication[ScalaLanguage, ?] ~> Target) {
-  val interpDefinitionPM: DefinitionPM[ScalaLanguage, ?] ~> Target       = ProtocolSupportTermInterp or ModelProtocolTermInterp
+  import com.twilio.guardrail.protocol.terms.protocol.{ PackageObjectImports, ProtocolSupportTerm }
+  val java8timeCirceInterp = new (ProtocolSupportTerm[ScalaLanguage, ?] ~> Target) {
+    import scala.meta._
+    def apply[A](value: ProtocolSupportTerm[ScalaLanguage, A]): Target[A] = value match {
+      case PackageObjectImports() =>
+        ProtocolSupportTermInterp(value).map { values =>
+          values :+ q"import io.circe.java8.time._"
```
hugo-vrijswijk: Please tell me if I am missing something, but doesn't this just add back the lines I removed in [CirceProtocolGenerator.scala](https://github.com/twilio/guardrail/pull/363/files#diff-ef22233a73a3c72d40555514fad41afbL370)? That would break on Scala 2.13/Circe 0.12 again.

My suggested change is to remove the imports by default, since they are only needed on Scala 2.11, and add a clarification to the documentation for 2.11 users.

blast-hardcheese: It seems as though there's no workaround for scalajs currently, until the next release. For that reason, I added them back in, effectively locking scalajs projects to 2.12, but permitting Scala projects to move on to 2.13. An alternate strategy would be to convince the example runner to inject the import for the scalajs target; I went for this option to limit impact, as it'll get reverted on the next Circe release, I think.

Does this impact you? If so, feel free to force push to take a different approach with your branch, I was focused on striking a balance during this Circe migration.

blast-hardcheese:
@hugo-vrijswijk An alternate approach for this PR would be to just depend on the milestone release of circe 0.12. I'm ready to merge either way, just looking to see if this actually solves the problem for you.

hugo-vrijswijk:
I was able to find some time to properly test this PR. I tested the following scenarios (all for http4s):

|            | Scala 2.11     | Scala 2.12 | Scala 2.13 |
|------------|----------------|------------|------------|
| Circe 0.11 | Doesn't work ❌ | Works! ✅   | N/A        |
| Circe 0.12 | N/A            | Works! ✅   | Works! ✅   |

Circe 0.11 on Scala 2.11 does not work because there are some compile errors related to ambiguous implicit values of `java.time.LocalDate` decoders.

blast-hardcheese:
It's good that 2.12+ work, though I'm still holding out hope for a 2.11 solution as well; I'll give it another shot tonight; it may be beneficial to have the different versions of Scala be tested for compilation as well during travis

kimxogus:
Is there any updates?
Hope to have this quick as java8 import blocks us bumping circe to 0.12

blast-hardcheese:
Just coming back to this today, looking towards a release within the next few days including this work.

ma3574:
Have started work on a Scala 2.13 proof of concept and have hit this exact bug. Thanks for finding a fix. Any ETA on when this will be available in a release?

blast-hardcheese:
I've got most of what I think can really solve this going forward implemented, I'm spending some time on the developer experience to avoid introducing something broken. My initial estimates were inaccurate, unfortunately, but I've got a clear path now.

As a workaround, sticking
```
package io.circe.java8.time
```
somewhere in your build should let you continue until this issue is resolved. Sorry for the delay!

blast-hardcheese:
OK, I've taken a slightly different approach with this PR. Scope has expanded, but the bonus is that this is something I've wanted to do for a while.

In addition to `--framework`, I've introduced `--module`. They are mutually exclusive.

| framework | modules | scala version |
| --- | --- | --- |
| `--framework akka-http`  | `--module akka-http --module circe` | Scala 2.12+ |
| `--framework http4s`  | `--module http4s --module circe` | Scala 2.12+ |
| `--framework endpoints`  | `--module endpoints --module circe-java8` | Scala 2.11 |

This opens the door for `--module refined` and `--module akka-http --module json4s` and other sorts of customization.

One thing I'm on the fence about is whether `--module` should scope where the argument should apply, for example:
```
  --module-client akka-http --module-server http4s
```
(even though currently you can't generate clients and servers in the same execution).

@hugo-vrijswijk, @kelnos, @sullis, I'd appreciate your input here.

@hugo-vrijswijk Thanks for your patience while we work through this. The motivation behind not just removing 2.11 is because internally we still have some services running that version of Scala. I realize I didn't explain that earlier.

blast-hardcheese:
Build failed because circe 0.12.0 had a bug, bumped to 0.12.1

kelnos:
Does scalajs not have java8 datetime types at all?!
```
[info] Fast optimizing /home/travis/build/twilio/guardrail/modules/sample-endpoints/target/scala-2.12/endpointssample-test-fastopt.js
[error] Referring to non-existent class java.time.OffsetDateTime
```

blast-hardcheese:
@kelnos It does not, but
```diff
-      "io.github.cquiroz" %%% "scala-java-time"               % "2.0.0-M13",
+      "org.scala-js"      %%% "scalajs-java-time"             % "0.2.5",
```
should have provided it. I'll take a look, but what do you think about the `--module` approach?

I'm thinking it may be beneficial to say `--framework endpoints --module circe-java8`, where the framework is used as the "default" answer to the questions of which modules are available, instead of having to start entirely from scratch and select all your modules...

... but at the same time I think the full decomposition here gets us into a position where we may be able to infer based on available libraries the answer to which modules to use, only raising an error if there's a conflict, but delivering a useful error message for how to resolve the conflict (either by deleting the offending libraries or selecting which variant you want to generate).

Anyway, I'm just looking for options to not break the existing 2.11 users that can't immediately jump ship to 2.12+.

kelnos:
I *really* like the module approach.  It would be great to be able to do some mix-and-match between HTTP framework and ser/deser library.  We still have a ways to go before that'd actually work (some of the client/server generators would definitely have to be ser/deser-aware), but I think it's the right way to go.

blast-hardcheese:
@kelnos Reverting the scalajs library swap, as I inadvertently swapped to a dependency that doesn't have real implementations for most of the classes.

COMPATIBILITY.md
```diff
@@ -31,3 +31,4 @@ guardrail version | http4s | circe-core | cats  | Notes
 0.34.0 ⚠          | 0.18.0 | 0.9.0      | 1.0.1 | Only clients are supported
 0.39.0            | 0.18.0 |  "         |  "    | First server release
 0.41.1            | 0.19.0 | 0.10.0     | 1.4.0 |
+0.5x.x            |  "     |  "         |  "    | For Scala 2.11, Add `io.circe:circe-java8` to your dependencies and add `"io.circe.java8.time._"` to your guardrail imports configuration.
```
kelnos: Should this be updated for the actual version we're gonna put this in?

blast-hardcheese: Yep, 'sec

blast-hardcheese: Done!

blast-hardcheese:
@hugo-vrijswijk Thanks for your patience!

hugo-vrijswijk:
@blast-hardcheese Sorry I haven't been able to provide further input about the module system. The module system seems like a very clean way to deal with this problem (and for future use as well)! Due to lack of familiarity with Guardrail I can't really provide more input, but I look forward to a release with the changes!

blast-hardcheese:
@hugo-vrijswijk The release should be up now! 0.54.0 should have all the previous stuff, as well as module support.

Issue #364, opened 2019-08-06
`dtoPackage` maven configuration parameter is not used correctly


aleksandr-lapushkin:
Steps to reproduce:

1. Use Java
2. Have 2 Maven executions. Both with different Swagger specs and `dtoPackage` configuration parameters (e.g. `service-a` and `service-b`). Otherwise identical.

The following structure is generated (`target` and other maven directories omitted):
```
> root
  > service-a -> generated client for A
  > service-b -> generated client for B
  > definitions
    > service-a -> definitions for A (dtoPackage = service-a)
    > service-b -> definitions for B (dtoPackage = service-b)
```

This is correct. However, the resulting classes inside `service-a` and `service-b` do not have the correct package declaration.

Expected:
Classes inside `root.definitions.service-a` should have the package `<whatever_the_root_is>.definitions.service-a`.

Actual:
Classes inside `root.definitions.service-a` have the package `<whatever_the_root_is>.definitions`

PR #365, opened 2019-08-08
Use correct package for Java definitions

aleksandr-lapushkin:
The goal is for generated Java definitions to have the correct package
defined in case the `dtoPackage` flag is provided.

This PR aims to address Issue #364 

**Contributing to Twilio**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -35,41 +35,42 @@ assemblyMergeStrategy in assembly := {
     oldStrategy(x)
 }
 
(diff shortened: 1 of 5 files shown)


kelnos:
While you're here, can you also fix https://github.com/twilio/guardrail/blob/master/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaGenerator.scala#L289 ?

PR #368, opened 2019-08-14
Avoid rewriting files with identical contents

kelnos:
I think sbt handles this reasonably (by hashing file contents), but it appears that maven pays more attention to file modification times.  It slows down things enough already that guardrail runs each time a build is run, but having to recompile files that haven't changed just adds insult to injury.

1. Running guardrail for the first time should have roughly the same performance (an extra `stat()` per file, which will fail since the file doesn't exist).
2. Running guardrail when files need to be updated will be either roughly the same or a bit slower, depending: a) if the file sizes are different, it's just a `stat()` plus integer comparison extra, so roughly the same; b) if the file sizes are the same, it has to read the existing file off disk and compare, so things will be a bit slower.
3. Running guardrail when files exist already and don't need to be updated should be slightly faster (an extra `stat()` and integer comparison, as well as reading the existing file off disk to compare, but no write of the file).

Of course, guardrail still has to do all the work to render the new files' contents to memory no matter what; in general that should be the dominating time factor.  There's only one case that ends up being worst (where "worst" is probably a few hundreds of nanoseconds per file extra), which should be the very, very, very uncommon case (as it requires the files to be different, but also the same size).

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/core/CoreTermInterp.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/core/CoreTermInterp.scala
@@ -2,12 +2,28 @@ package com.twilio.guardrail
 package core
 
 import cats.data.{ NonEmptyList, State }
+import cats.free.Free
 import cats.implicits._
 import cats.{ FlatMap, ~> }
 import com.twilio.guardrail.languages.LA
 import com.twilio.guardrail.terms._
-import java.nio.file.Paths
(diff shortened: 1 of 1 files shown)


blast-hardcheese:
Didn't consider the implication here. Even if the files are not rewritten, they still needed to be returned to plugins to ensure they're passed to `scalac`. (Related: https://github.com/twilio/sbt-guardrail/issues/26 )

Issue #179, opened 2019-01-28
http4s generator creates invalid code for 401 response


countfloyd:
Using swagger snippet:

```
      responses:
        200:
          description: OK
          schema:
            $ref: '#/definitions/JwtToken'
        401:
          description: Login invalid
```

The http4s code generator creates:

```
trait AuthHandler[F[_]] { def authenticate(respond: AuthenticateResponse.type)(login: Login)(traceBuilder: TraceBuilder[F]): F[AuthenticateResponse] }
class AuthResource[F[_]](trace: String => Request[F] => TraceBuilder[F])(implicit E: Effect[F]) extends Http4sDsl[F] {
  val authenticateDecoder: EntityDecoder[F, Login] = jsonOf[F, Login]
  val authenticateOkEncoder = jsonEncoderOf[F, JwtToken]
  object usingForAuthenticate { def unapply(r: Request[F]): Option[(Request[F], TraceBuilder[F])] = Some(r -> trace("auth:authenticate")(r)) }
  def routes(handler: AuthHandler[F]): HttpRoutes[F] = HttpRoutes.of {
    {
      case req @ POST -> Root / "api" / "v1" / "auth" / "login" usingForAuthenticate traceBuilder =>
        req.decodeWith(authenticateDecoder, strict = false) { login => 
          handler.authenticate(AuthenticateResponse)(login)(traceBuilder) flatMap {
            case AuthenticateResponse.Ok(value) =>
              Ok(value)(E, authenticateOkEncoder)
            case AuthenticateResponse.Unauthorized =>
              Unauthorized()
          }
        }
    }
  }
}
sealed abstract class AuthenticateResponse {
  def fold[A](handleOk: JwtToken => A, handleUnauthorized: => A): A = this match {
    case x: AuthenticateResponse.Ok =>
      handleOk(x.value)
    case AuthenticateResponse.Unauthorized =>
      handleUnauthorized
  }
}
object AuthenticateResponse {
  case class Ok(value: JwtToken) extends AuthenticateResponse
  case object Unauthorized extends AuthenticateResponse
}
```

I then get a compile error:

```
Routes.scala:38:15: overloaded method value apply with alternatives:
[error]   [A](authenticate: org.http4s.headers.WWW-Authenticate, body: A, headers: org.http4s.Header*)(implicit F: cats.Monad[F], implicit w: org.http4s.EntityEncoder[F,A])F[org.http4s.Response[F]] <and>
[...]

PR #374, opened 2019-08-27
feat: add support for http4s AuthedRoutes middleware

ingarabr:
http4s include a middleware for authenticating users. To enable this feature you need to add the `security` configuration to the openApi specification (see server3 for an example) and include the `--http4sAuthedRoutes` argument when generating the code. Note, this is only the server side
implementation. The client already has the option to pass headers.

Upgraded http4s since `AuthedRoutes` isn't available in `0.20.0`. I also bumped into https://github.com/twilio/guardrail/issues/179 so that's why some of the code is commented out.

Signed-off-by: Ingar Abrahamsen <ingar.abrahamsen@tapad.com>

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/Context.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/Context.scala
@@ -1,7 +1,7 @@
 package com.twilio.guardrail
 
-case class Context(framework: Option[String], tracing: Boolean)
+case class Context(framework: Option[String], tracing: Boolean, http4sAuthedRoutes: Boolean)
 
 object Context {
-  val empty = Context(None, tracing = false)
+  val empty = Context(None, tracing = false, http4sAuthedRoutes = false)
 }
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
@@ -43,35 +43,43 @@ object Http4sServerGenerator {
           } else Target.pure(None)
         } yield res)
 
-      case GenerateRoutes(tracing, resourceName, basePath, routes, protocolElems, securitySchemes) =>
+      case GenerateRoutes(tracing, resourceName, basePath, routes, protocolElems, securitySchemes, authedRoutes) =>
+        def toListOfRoutes(l: List[Case]) = if (l.nonEmpty) combineRouteTerms(l).map(List(_)) else Target.pure(List.empty)
         for {
-          renderedRoutes <- routes
+          (authedRenderedRoutes, renderedRoutes) <- routes
             .traverse {
               case (operationId, tracingFields, sr @ RouteMeta(path, method, operation, securityRequirements), parameters, responses) =>
-                generateRoute(resourceName, basePath, sr, tracingFields, parameters, responses)
+                generateRoute(resourceName, basePath, sr, tracingFields, parameters, responses, authedRoutes)
             }
             .map(_.flatten)
-          routeTerms = renderedRoutes.map(_.route)
-          combinedRouteTerms <- combineRouteTerms(routeTerms)
-          methodSigs = renderedRoutes.map(_.methodSig)
+            .map(_.span(_.authedRoute))
+
+          routeTerms       <- toListOfRoutes(renderedRoutes.map(_.route))
+          authedRouteTerms <- toListOfRoutes(authedRenderedRoutes.map(_.route))
+          allRoutes  = renderedRoutes ++ authedRenderedRoutes
+          methodSigs = allRoutes.map(_.methodSig)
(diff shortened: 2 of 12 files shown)


modules/sample/src/main/resources/server3.yaml
```diff
@@ -0,0 +1,73 @@
...
+      security:
+        - basicAuth: []
+      produces:
+        - application/json
+      responses:
+        '200':
+          type: array
+          items:
+            $ref: '#/definitions/SecureAddress'
+#        '401':
```
blast-hardcheese: Was this commented out due to #179? If so, the fix has been merged, hopefully this can get uncommented

modules/codegen/src/main/scala/com/twilio/guardrail/Context.scala
```diff
@@ -1,7 +1,7 @@
 package com.twilio.guardrail
 
-case class Context(framework: Option[String], tracing: Boolean)
+case class Context(framework: Option[String], tracing: Boolean, http4sAuthedRoutes: Boolean)
```
blast-hardcheese: This PR represents a significant time investment, I appreciate what you've done so far. One of the design principles that may not have been clear is that the core of guardrail does not actually know anything about implementation details of any libraries; as a result, having a top-level context boolean that only applies to a single framework is undesirable.

An alternate strategy could be what we do in the [dropwizard generator](https://github.com/twilio/guardrail/blob/master/modules/codegen/src/main/scala/com/twilio/guardrail/generators/SwaggerGenerator.scala#L42-L44), where we build [`SecurityRequirements`](https://github.com/twilio/guardrail/blob/master/modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerm.scala#L47) both globally (set top-level in the OpenAPI specification) as well as locally (configured for an individual operation).

You could then use the presence or absence of `SecurityRequirements` to enable or disable this functionality.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -85,19 +93,36 @@ object Http4sServerGenerator {
...
-        } yield q"""
-          class ${Type.Name(resourceName)}[F[_]](..$extraRouteParams)(implicit F: Async[F]) extends Http4sDsl[F] {
-
+          types        = if (authedRouteTerms.isEmpty) List(Type.Name("F")) else List(Type.Name("F"), Type.Name("U"))
+          routesParams = List(param"handler: ${Type.Name(handlerName)}[..$types]")
+        } yield {
+          val authedRoutes =
+            if (authedRouteTerms.isEmpty) None
+            else Some(q"""
+            def authedRoutes[U](..${routesParams}): AuthedRoutes[U, F] = AuthedRoutes.of {
```
blast-hardcheese: Instead of defining a new `authedRoutes` function and requiring folk to call it, why not alter the return type of `routes` to yield a tuple of the non-authed routes as well as the authed routes?

If we take this approach, when a user switches on authed functionality, the compiler error makes it very obvious what needs to happen next.

modules/codegen/src/main/scala/com/twilio/guardrail/Context.scala
blast-hardcheese: One other consideration here is that as this is a global boolean, there's no convenient way to combine authed and unauthed paths in the same configuration. By using this `security` context configuration, it could be turned on and off by route.

ingarabr: I expected some feedback here. As you mention, it didn't fit so well with the rest of the design. The intention here was to not change the behavior of the existing solution if you didn't want to opt-in to it. I'll do another pass on it and see what I can find a better way to solve it.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
ingarabr: 👍 Should be easy to change.

francescopellegrini:
Any update on this? I'd really like to see `security` configuration support for http4s too! :)

blast-hardcheese:
Considering the inactivity on this PR, if someone wants to rebase off master and address the comments, that would be very much appreciated.

I don't anticipate any major breaking changes that would cause nontrivial merge conflicts coming up that can't be pushed off past when this PR is merged, and I'm also available to coach for new contributors, feel free to reach out to [@guardrail_code](https://twitter.com/guardrail_code) if you're interested!

blast-hardcheese:
I'm happy to say, this was implemented in a different way! #1342 

Closing this one out, and really appreciate your effort here, even though we weren't able to push it over the line at the time.

ingarabr:
I'm glad that someone picked it up. Thanks for the update :-)

Issue #53, opened 2018-04-19
Response headers


manuelbernhardt:
Given the following `response` definition:

        // ...
        "responses" : {
          "200" : {
            "description" : "foo",
            "headers" : {
              "version" : {
                "type" : "integer",
                "default" : 3
              }
            }
          }

The generated handler doesn't seem to include a way to set the header value. Is this a bug, or a missing feature? In case it is missing, any hints as to where to look for adding this would be welcome!

blast-hardcheese:
As for implementation of this feature, the code is all currently localized in the [`AkkaHttpServerGenerator`](https://github.com/twilio/guardrail/blob/e8fbbf3fbd27660db32cfa36cae2f32de3e08e27/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttpServerGenerator.scala#L154-L182).

Accumulating and then flattening a `List[Option[...]]`s representing the optional response body and subsequent `Some(...)`s for the various headers seems like a good strategy, and switching from `case Foo(value) => ... value ...` to an actual name supplier (`x1`, `x2`, ...) would eliminate the concerns around having headers named `value` and getting name conflicts.

manuelbernhardt:
Thanks @blast-hardcheese! I will try and have a look when time permits!

blast-hardcheese:
Thanks to @hanny24, this has been implemented for scala clients and servers!

PR #382, opened 2019-09-03
Issue #53 for http4s server/client, endpoints/akka clients

hanny24:
Issue #53 for http4s server/client, endpoints/akka clients

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
@@ -5,9 +5,10 @@ import cats.arrow.FunctionK
 import cats.data.NonEmptyList
 import cats.implicits._
 import com.twilio.guardrail.extract.{ ServerRawResponse, TracingLabel }
+import com.twilio.guardrail.generators.syntax._
 import com.twilio.guardrail.generators.syntax.Scala._
 import com.twilio.guardrail.languages.ScalaLanguage
-import com.twilio.guardrail.protocol.terms.{ Response, Responses }
+import com.twilio.guardrail.protocol.terms.{ Header, Response, Responses }
 import com.twilio.guardrail.protocol.terms.server._
 import com.twilio.guardrail.shims._
 import com.twilio.guardrail.terms.RouteMeta
@@ -468,14 +469,29 @@ object Http4sServerGenerator {
           .filter(_ == true)
           .fold[Term] {
             val marshallers = responses.value.map {
-              case Response(statusCodeName, valueType) =>
-                val responseTerm = Term.Name(s"${statusCodeName.value}")
-                valueType.fold[Case](
-                  p"case $responseCompanionTerm.$responseTerm => F.pure(Response[F](status = org.http4s.Status.${statusCodeName}))"
-                ) { _ =>
-                  val generatorName = Term.Name(s"$operationId${statusCodeName}EntityResponseGenerator")
-                  val encoderName   = Term.Name(s"$operationId${statusCodeName}Encoder")
-                  p"case $responseCompanionTerm.$responseTerm(value) => $generatorName(value)(F,$encoderName)"
+              case Response(statusCodeName, valueType, headers) =>
+                val responseTerm  = Term.Name(s"${statusCodeName.value}")
+                val respType      = Type.Select(responseCompanionTerm, Type.Name(statusCodeName.value))
+                val generatorName = Term.Name(s"$operationId${statusCodeName}EntityResponseGenerator")
+                val encoderName   = Term.Name(s"$operationId${statusCodeName}Encoder")
+                (valueType, headers.value) match {
+                  case (None, Nil) =>
+                    p"case $responseCompanionTerm.$responseTerm => F.pure(Response[F](status = org.http4s.Status.${statusCodeName}))"
+                  case (Some(_), Nil) =>
+                    p"case resp: $respType => $generatorName(resp.value)(F,$encoderName)"
+                  case (None, headersList) =>
+                    val (http4sHeaders, http4sHeadersDefinitions) = createHttp4sHeaders(headersList)
+                    p"""case resp: $respType =>
+                          ..$http4sHeadersDefinitions
+                          F.pure(Response[F](status = org.http4s.Status.$statusCodeName, headers = Headers($http4sHeaders)))
+                      """
+                  case (Some(_), headersList) =>
+                    val (http4sHeaders, http4sHeadersDefinitions) = createHttp4sHeaders(headersList)
+                    val valueTerm                                 = q"resp.value"
+                    p"""case resp: $respType =>
+                          ..$http4sHeadersDefinitions
+                          $generatorName($valueTerm, $http4sHeaders:_*)(F,$encoderName)
+                      """
                 }
             }
             q"$handlerCall flatMap ${Term.PartialFunction(marshallers)}"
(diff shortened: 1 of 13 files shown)


hanny24:
@blast-hardcheese Do you think you'll have a time to review this soon?

blast-hardcheese:
@hanny24 Yes, I'm still catching up with open issues/PRs from while I was in the mountains. I'll try to get to it within the week. Thank you for your patience.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -554,6 +574,21 @@ object Http4sServerGenerator {
...
+      val (names, definitions) = headers.map {
+        case Header(name, required, _, termName) =>
+          val nameLiteral = Lit.String(name)
+          val headerName  = Term.Name(s"${name.toCamelCase}Header")
+          val pattern     = Pat.Var(headerName)
+          val v           = if (required) q"List(Header($nameLiteral, $termName))" else q"$termName.map(Header($nameLiteral,_)).toList"
+          (headerName, q"val $pattern = $v")
+      }.unzip
+      val patterns    = headers.map(header => Pat.Var(header.term))
+      val headersTerm = Term.Name("responseHeaders")
```
blast-hardcheese: This is unfortunately in the same namespace as user-controlled values. Could you alter the http4s server generator to be like
```diff
-            case LoginUserResponse.Ok(value, xRateLimit, xExpiresAfter) =>
-              val xRateLimitHeader = xRateLimit.map(Header("X-Rate-Limit", _)).toList
-              val xExpiresAfterHeader = xExpiresAfter.map(Header("X-Expires-After", _)).toList
-              val responseHeaders = List(xRateLimitHeader, xExpiresAfterHeader).flatten
-              loginUserOkEntityResponseGenerator(value, responseHeaders: _*)(F, loginUserOkEncoder)
+            case resp: LoginUserResponse.Ok =>
+              val xRateLimitHeader = resp.xRateLimit.map(Header("X-Rate-Limit", _)).toList
+              val xExpiresAfterHeader = resp.xExpiresAfter.map(Header("X-Expires-After", _)).toList
+              val responseHeaders = List(xRateLimitHeader, xExpiresAfterHeader).flatten
+              loginUserOkEntityResponseGenerator(resp.value, responseHeaders: _*)(F, loginUserOkEncoder)
```
? That would completely disambiguate values we control from values the user controls. This is how the other generators work (I checked by diffing the output from current master against the output from this branch)

hanny24: Thanks, great suggestion! It also makes the code much simpler.

hanny24: Generated client code is solved differently thought: we append `Header` suffix to each header such that we can be sure that are no conflicts.

blast-hardcheese: The edge case would be if I had two parameters, `foo` and `fooHeader`, then suffixed both with `Header` -- I'd then get a conflict.

blast-hardcheese: We've already had cases like this, where someone has a data model named `Request` which conflicts with http4s' built-in types. I'm staying extra vigilant until proper tests for unconstrained terms/types have been written.

hanny24: Ahh, I see. Do you have any suggestion?

blast-hardcheese: The resulting diff of https://github.com/twilio/guardrail/pull/382#discussion_r322992570 would solve the problem, as all access would be constrained by the `resp` object.

hanny24: That's already done for the server side. What I'm talking about is a client side where we currently generate something like this: 
```scala
case Ok(resp) =>
        val xOptionalHeaderHeader =
          resp.headers.get("x-optional-header".ci).map(_.value)
        val result =
          for (xRequiredHeaderHeader <- EitherT.fromEither[F](
                 resp.headers
                   .get("x-required-header".ci)
                   .map(_.value)
                   .toRight(
                     ParseFailure(
                       "Missing required header.",
                       "HTTP header 'X-Required-Header' is not present."
                     )
                   )
               ))
            yield
              GetFooResponse.Ok(xRequiredHeaderHeader, xOptionalHeaderHeader)
        result.value.flatMap({
          case Right(v) =>
            F.pure(v)
          case Left(e) =>
            F.raiseError(e)
        })
```

blast-hardcheese: Oh, sorry, I misunderstood. I took that diff from the server generator. I think you're right for the client generator, other than currying and partially applying the constructor based on whether the field is required or not. Unsure what it would look like, but it would be great to use applicative here to avoid bindings.

hanny24: Implemented even for http4s client.
```
case Ok(resp) =>
        F.map2(parseRequiredHeader(resp, "x-required-header"), parseOptionalHeader(resp, "x-optional-header"))(GetFooResponse.Ok)
```

It uses two helpers methods `parseRequiredHeader` and `parseOptionalHeader` that are defined directly in the generated `Client` class. Perhaps there is a better place for it?

blast-hardcheese: @hanny24 A challenge here is that for more than 22 parameters, even this approach will start failing. I mistakenly assumed it would be as easy as `Ok.apply _ curried`, but it appears that that machinery is limited to 22 parameters _as well_ for some reason.

The only thing I can think of that gets us close to safety here is of the structure:

```scala
case class Foo(
  a1: Int, a2: Int, a3: Int, a4: Int, a5: Int, a6: Int, a7: Int, a8: Int, a9: Int, a10: Int, a11: Int, a12: Int, a13: Int, a14: Int, a15: Int, a16: Int, a17: Int, a18: Int, a19: Int, a20: Int, a21: Int, a22: Int,
  a23: Int, a24: Int, a25: Int, a26: Int, a27: Int, a28: Int, a29: Int
)

Applicative[Option].ap7(
  Applicative[Option].map22(
    Option(1), Option(2), Option(3), Option(4), Option(5), Option(6), Option(7),
    Option(8), Option(9), Option(10), Option(11), Option(12), Option(13),
    Option(14), Option(15), Option(16), Option(17), Option(18), Option(19),
    Option(20), Option(21), Option(22)
  )(
    (
      a1: Int, a2: Int, a3: Int, a4: Int, a5: Int, a6: Int, a7: Int, a8: Int,
      a9: Int, a10: Int, a11: Int, a12: Int, a13: Int, a14: Int, a15: Int,
      a16: Int, a17: Int, a18: Int, a19: Int, a20: Int, a21: Int, a22: Int
    ) =>
      Foo(
        a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13, a14, a15, a16,
        a17, a18, a19, a20, a21, a22,
        _: Int, _: Int, _: Int, _: Int, _: Int, _: Int, _: Int
      )
  )
)(
  Option(23), Option(24), Option(25), Option(26), Option(27), Option(28),
  Option(29)
)
```
but that requires _way_ more juggling than I'd like to take on currently. You're welcome to, if you'd like, otherwise just assert that if there are more than 22 parameters for the resulting `Response` container, fail the codegen and link to this thread (both in the error message and a comment nearby).

hanny24: Ok, the build now succeeds. Once the PR is approved I can squash all changes.

blast-hardcheese: I just pushed a couple more commits, pushing the validation logic into `build` so the validation happens close to the consumption, to prevent it from getting lost.

I'm onboard merging with this, if you have no objections.

Thank you for your patience through this whole process

hanny24: I'm OK with your commits. Feel free to merge it. Thanks!

blast-hardcheese:
Cleaned up PR history

PR #398, opened 2019-09-19
Implemented issue-370

### LINKED ISSUE #370: Anonymous enum support
We should support anonymous enums. Given this definition:

```
  Foo:
    type: object
    properties:
      value:
        type: string
        enum: [X, Y]
```

we should generate `sealed trait Value` in a companion object of `Foo`. Currently `value` has type `String`.

hanny24:
This PR closes #370 in a similar way how nested structures were introduced in https://github.com/twilio/guardrail/pull/356

**Contributing to Twilio**


modules/codegen/src/test/scala/tests/core/DereferencingAliasesSpec.scala
```diff
@@ -74,7 +74,7 @@ class DereferencingAliasesSpec extends FunSuite with Matchers with SwaggerSpecRu
     val clientCmp = companionForStaticDefns(clientStaticDefns)
 
     val definition = q"""
-      case class propRef(param: Option[Long] = None, array: Option[IndexedSeq[Long]] = None, arrayArray: Option[IndexedSeq[IndexedSeq[Long]]] = None)
+      case class propRef(param: Option[Long] = None, array: Option[IndexedSeq[Long]] = Option(IndexedSeq.empty), arrayArray: Option[IndexedSeq[IndexedSeq[Long]]] = Option(IndexedSeq.empty))
```
blast-hardcheese: This one is odd -- it changes the semantics from `None` (not present) to `Some([])` (present, empty sequence). Was this intentional?

hanny24: I somehow managed to unify this behavior across different part of the generator. For example here: https://github.com/twilio/guardrail/blob/a57b3368aefdac17990752450552c7640d75eb07/modules/codegen/src/test/scala/tests/core/TypesTest.scala#L89 it is also `Option(IndexedSeq.empty)`. 

It's caused by the fact the we now look for a default value of referenced schema as well, which was not the case before.

However I agree that `Option(IndexedSeq.empty)` is kind of weird. If I could make a decision, I would consider default values only in case when an attribute is required.

hanny24:
Is there any update on this?

modules/codegen/src/test/scala/tests/core/DereferencingAliasesSpec.scala
blast-hardcheese: My concern is that translating these values in methods with update semantics, this could start clearing values. Additionally, if we unify the behaviour to `None` instead of `Some(Nil)`, then users can choose to clear by passing `Some(Nil)` themselves, instead of that being the default.

This is an edge case, but I believe this matches the semantics of HTTP methods most closely.

blast-hardcheese:
@hanny24 Please pardon the delay in response. Once we agree on a solution to my most recent comment, I don't see any blocker for merging this.

modules/codegen/src/test/scala/tests/core/DereferencingAliasesSpec.scala
hanny24: Unified to use `None` instead of `Option(IndexedSeq.empty)` and `Option(Map.empty)`.

PR #402, opened 2019-09-20
Better error messages

blast-hardcheese:
Problem:

`Missing operationId` is not a useful error message, especially with particularly large specification files.

--

Given:
```scala
class Foo(val a: Bar)
class Bar(val b: Baz)
class Baz(val c: String)
```

the following will compile:
```scala
val result: Tracker[Option[String]] =
  Tracker(foo).downField("b", _.b).flatDownField("c", _.c)
```

~there are a number of combinators borne of necessity, though my intent that it should be possible to write `Traverse` and `Monoid` instances for these structures, which may simplify syntax a bit.~

I've broken out `IndexedFunctor` and `IndexedDistributive`, as well as some syntax around these combinators, but that permitted a significant reduction in bespoke syntax extensions.

The intent of opening this PR is to get feedback on this approach, as well as to publicize the fact that this is coming, in case others plan on doing major refactors.

**Contributing to Twilio**


blast-hardcheese:
OK, ready for review. There are many more warnings around `.get`, but this is to track the process of infecting more functions with `Tracker`. Everywhere there's a `.get` currently, there's an opportunity to provide better error messages, as we're unwrapping the `Tracker` directly without showing its history to the user.

modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
```diff
@@ -321,146 +322,89 @@ object SwaggerUtil {
...
+                  case Resolved(inner, dep, _, _, _) =>
+                    liftMapType(inner).map(Resolved[L](_, dep, None, None, None))
+                  case x: DeferredMap[L]   => embedMap(x)
+                  case x: DeferredArray[L] => embedMap(x)
+                  case x: Deferred[L]      => embedMap(x)
+                }
+              } yield res
+          )
+          .orRefine({ case ref: Schema[_] if Option(ref.get$ref).isDefined => ref })(ref => getSimpleRef(ref.get).map(Deferred[L]))
+          .orRefine({ case b: BooleanSchema => b })(buildResolve(litBoolean))
```
blast-hardcheese: Turns out these were all the same, with only two variants (has default vs no default possible)

modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/package.scala
```diff
@@ -2,7 +2,37 @@ package com.twilio.guardrail.generators
...
+import scala.language.implicitConversions
+
+package syntax {
+  class RichNotNullShower[A](value: A) {
+    def showNotNull: String = showNotNullIndented(0)
+    def showNotNullIndented(indent: Int): String =
+      ("  " * indent) + value.toString().linesIterator.filterNot(_.contains(": null")).mkString("\n" + ("  " * indent))
+  }
+
+  class RichCollection[A, Repr](xs: IterableLike[A, Repr]) {
```
blast-hardcheese: from https://stackoverflow.com/a/3912833. `distinctBy` is a member of 2.13's `View`, so this can go away when we upgrade.

modules/codegen/src/main/scala/com/twilio/guardrail/core/Tracker.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  implicit class EitherSyntax[A, B](tracker: Tracker[Either[A, B]]) {
+    def fold[C](a: Tracker[A] => C, b: Tracker[B] => C): C =
+      tracker.unwrapTracker.fold(x => a(Tracker.cloneHistory(tracker, x)), x => b(Tracker.cloneHistory(tracker, x)))
+  }
+
+  implicit class OptionSyntax[A](tracker: Tracker[Option[A]]) {
+    def orHistory: Either[Vector[String], Tracker[A]]      = tracker.indexedCosequence.toRight(tracker.history)
+    def raiseErrorIfEmpty(err: String): Target[Tracker[A]] = Target.fromOption(tracker.indexedCosequence, s"${err} (${tracker.showHistory})")
+    class FlatDownFieldPartiallyApplied[C](val dummy: Boolean = true) {
```
kelnos: `extends AnyVal`?  (Otherwise, why bother with `dummy`?)

(& other examples of the same in this file)

modules/codegen/src/main/scala/com/twilio/guardrail/core/Tracker.scala
```diff
@@ -0,0 +1,126 @@
...
+        new Tracker(ev(Option(f(tracker.get))), tracker.history :+ s".${label}")
+    }
+    def downField[C]: DownFieldPartiallyApplied[C] = new DownFieldPartiallyApplied()
+
+    def showHistory: String = tracker.history.mkString("")
+
+    @deprecated("Tracker.get will be removed once the migration has been completed. Please use the fold/traverse combinators to get values out.", "0.0.0")
+    def get: A = tracker.get
+
+    def unwrapTracker: A = tracker.get
```
kelnos: Just to test my understanding: `unwrapTracker` is for places where you really do need to just pull the value out and there's no possibility to fold or traverse, whereas `get` is for places where things aren't yet done the right way, and need to be revisited later?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/package.scala
```diff
@@ -2,7 +2,37 @@ package com.twilio.guardrail.generators
...
 import java.util.Locale
+import io.swagger.v3.oas.models.{ Operation, PathItem }
 import io.swagger.v3.oas.models.media.Schema
+import io.swagger.v3.oas.models.parameters.Parameter
+import scala.collection.IterableLike
+import scala.collection.generic.CanBuildFrom
+import scala.language.implicitConversions
+
+package syntax {
+  class RichNotNullShower[A](value: A) {
```
kelnos: Aww, you made a `Shower` too ;)

modules/codegen/src/main/scala/com/twilio/guardrail/core/Tracker.scala
blast-hardcheese: This follows the pattern used in cats to partially apply type arguments. `dummy` is presumably required to avoid syntactic ambiguity, but the big thing is by fixing `C` in the call site, it permits:

`t.downField[Int]("foo", _.getFoo)` instead of `t.downField[Option[Int], Int]("foo", _.getFoo)`

Small improvement in ergonomics for the user, as it frees the unspecified type parameter to be resolved by the compiler.

modules/codegen/src/main/scala/com/twilio/guardrail/core/Tracker.scala
blast-hardcheese: Exactly that. I added `get` initially, but then it was difficult to disambiguate between `Option.get` (which is regrettably present) and `tracker.get`. I marked `get` as deprecated and added this.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/package.scala
blast-hardcheese: 🚿📈

modules/codegen/src/main/scala/com/twilio/guardrail/core/Tracker.scala
```diff
@@ -0,0 +1,126 @@
...
+
+import _root_.io.swagger.v3.oas.models.OpenAPI
+import cats.data._
+import cats.implicits._
+import com.twilio.guardrail.Target
+import scala.collection.JavaConverters._
+
+class Tracker[+A] private[core] (private[core] val get: A, private[core] val history: Vector[String]) {
+  override def toString(): String = s"Tracker($get, $history)"
+  override def hashCode(): Int    = get.hashCode // FIXME: Are two things with different histories identical?
```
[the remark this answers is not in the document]
blast-hardcheese: Pretty sure that `distinctBy` removed the need for this entirely. We have a few places where `distinct` is used with the expectation that order will be preserved, this was an attempt to make it just not matter. Running a test right now with it removed.

modules/codegen/src/main/scala/com/twilio/guardrail/core/Tracker.scala
kelnos: Yeah, I've seen that pattern cats, but they do `extends AnyVal` on those.  I assumed the need for `dummy` was just because you have to have one-and-only-one `val` constructor param when extending `AnyVal`.

blast-hardcheese:
To follow up on "I don't really understand some or it, like `IndexedDistributive`", I found [`Distributive`](https://hackage.haskell.org/package/distributive-0.5.3/docs/Data-Distributive.html) to solve part of the problem I was having, as described here:

given `G[F[A]]`, I want `F[G[A]]`, but I want the typeclass that controls exactly how `G[_]` is applied over the different `F[_]`s to know that that is what's happening, instead of how `traverse`/`sequence` work, in the sense of strictly regimented generic methods that don't know about either structure.

I think the name comes from the distributive property of arithmetic: `A * (B + C)` == (A * B) + (A * C)`, `Tracker * List[B + C]` == `List[Tracker[B] + Tracker[C]]`.

---

The other side of the name, `Indexed`, comes from the fact that `Distributive` in that Hackage link doesn't know anything about the implementation details of the data structure it's walking over. I needed the `key: String` of a `Map` to be in scope so I could appropriately represent that we've indexed into a `Tracker[Map[K, V]]` to get a `Map[K, Tracker[V]]` for all `K`s, similarly with `Tracker[List[A]]` and `List[Tracker[A]]`. This is expressed in the tests by the `history.downField.list` and `history.downField.map` tests.

---

Initially I actually didn't have this abstraction, and I ended up writing almost identical syntax extensions for all the different structures I was trying to represent. There was a lot of duplication and divergent method names, since I wasn't sure what I had already implemented. When I started unifying them under this new banner of `IndexedFunctor` and `IndexedDistributive` things really started cleaning up.

blast-hardcheese:
Rebased!

kelnos:
Welp, I don't really understand the stuff in that hackage link, but your description helped a lot, thanks!

Issue #405, opened 2019-09-22
String form parameters no longer accept empty string in Akka/Scala


efossier:
In the latest version of guardrail, there is a regression in form parameter handling in generated AkkaHttp servers. Previously, sending an empty string value to the endpoint would correctly pass an empty string to the controller method, however this recently changed and the request now treats the param as not being sent at all, if it is required it will be rejected with the error that the value is required.

Sample swagger spec:
```yaml
openapi: 3.0.0
info:
  title: Test Server
  version: "1.0"
servers:
  - url: http://localhost:8000
paths:
  /v1/Foo:
    post:
      summary: Test resource
      operationId: foo
      requestBody:
        content:
          application/x-www-form-urlencoded:
            schema:
              properties:
                Bar:
                  type: string
                  description: A required string parameter that should accept empty string
                Baz:
                  type: string
                  description: An optional string parameter that should accept empty string
              required:
                - Bar
      responses:
        204:
          description: Success
```

efossier:
I am taking a look at fixing this

PR #408, opened 2019-09-26
Issue 405 Fix regression in empty string Form param handling in AkkaHttpServer 

efossier:
This fixes https://github.com/twilio/guardrail/issues/405 by removing the `contentRequiredUnmarshaller` from the generated form param parsing code. This prevented clients from sending empty string values to String form params which was a regression from earlier versions of guardrail. 

The `contentRequiredUnmarshaller` is not required because for other types, the type specific parser should fail on an empty string input, however for String this should be allowed.

**Contributing to Twilio**


blast-hardcheese:
Looks like a formatting issue, could you `sbt format` and commit the result? No other objections

PR #413, opened 2019-10-06
Adding more docs codegen

blast-hardcheese:
Converting akka-http Client code to codegen in the docs.

Getting back into the docs after Tracker and the `circe-java8`/module issues are largely resolved. Ideally, this would not be destructured, it'd just be consuming `Free` so the codegen samples can be reused between the different languages/frameworks/etc, but hopefully that'll get fleshed out as the docs continue to evolve.

**Contributing to Twilio**


blast-hardcheese:
There was also a bug introduced by the `Tracer` stuff that would prevent the release from succeeding because `testSuite` didn't compile the microsite module; I fixed the bug and added microsite compilation to `testSuite`.

Issue #77, opened 2018-07-14
Unify identical objects


blast-hardcheese:
Currently, identical classes are completely incompatible, even if generated from the same OpenAPI specification file. This is due to different JVM package paths being chosen for generated classes, and can be worked around by having identical definitions generated with the same package and name.

This is not scalable, as generating clients and servers in different scopes (`compile` vs `test:compile`, for instance) would be very fragile, if it worked at all.

A better solution would be to keep track of all generated classes (this would additionally help to avoid name conflicts between different OpenAPI specification files), hashing structures to determine if they are identical, then only emitting one (with type aliases or some other technique to maintain the API of the deduplicated classes).

Something else to consider would be conversions between different types, if they are close enough (optional parameters vs required parameters, for instance, would permit going one direction but not necessarily the other direction).

stanislav-chetvertkov:
Do items for `guardrailTasks` task (ScalaClient and ScalaServer definitions) get combined and translated into single CLI call by the SBT plugin?

blast-hardcheese:
> Do items for `guardrailTasks` task (ScalaClient and ScalaServer definitions) get combined and translated into single CLI call by the SBT plugin?

Yes, ideally. CLI and sbt integrations do, and since the Gradle plugin is based on the CLI driver that does as well. Maven unfortunately does not, though perhaps there is a way to convince it to do so.

PR #415, opened 2019-10-10
Fixed handling of nested structures/enums when polymorphism is used

hanny24:
This fixes an issue that resulted into a code that was not possible to compile.

Essentially a wrong name of companion object was used when nested enum/case class was referenced.

**Contributing to Twilio**


blast-hardcheese:
I suspect when #77 is addressed, we'll have real types to put in the `Map[_, _]`. For now, these strings seem to be the Swagger identifiers, which is sufficient for the time being.

blast-hardcheese:
Also, thanks for helping shore up the edge cases around this use case!

Issue #420, opened 2019-10-18
Ordering issue in composed objects


tomasherman:
Hello, it's been a long time! In the meanwhile we started using guardrail at work thanks to your hard work and @hanny24 s contribution. Thank you! We found the following error, though.

Ordering matters in composed objects. This code compiles just fine:

```yaml
                          |swagger: '2.0'
                          |definitions:
                          |  Bar:
                          |    properties:
                          |      id:
                          |        type: string
                          |  Foo:
                          |    allOf:
                          |      - $ref: "#/definitions/Bar"
                          |      - type: object
                          |        properties:
                          |          otherId:
                          |            type: string
```

while this one fails with `UserError(Reference None not found among definitions)`:

```yaml
                          |swagger: '2.0'
                          |definitions:
                          |  Bar:
                          |    properties:
                          |      id:
                          |        type: string
                          |  Foo:
                          |    allOf:
                          |      - type: object
                          |        properties:
                          |          otherId:
                          |            type: string
                          |      - $ref: "#/definitions/Bar"

```

I am creating PR with the issue

PR #421, opened 2019-10-18
Issue 420 - Ordering breaking composed objects

tomasherman:
**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
@@ -381,7 +381,7 @@ object CirceProtocolGenerator {
       case ExtractSuperClass(swagger, definitions) =>
         def allParents: Tracker[Schema[_]] => Target[List[(String, Tracker[Schema[_]], List[Tracker[Schema[_]]])]] =
           _.refine[Target[List[(String, Tracker[Schema[_]], List[Tracker[Schema[_]]])]]]({ case x: ComposedSchema => x })(
-            _.downField("allOf", _.getAllOf().asScala.toList.filter(s => Option(s.get$ref()).isDefined)).indexedDistribute match {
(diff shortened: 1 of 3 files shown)


tomasherman:
failing test for #420 - if anyone is able to fix this quickly feel free :) or i will try to get to it when i have some time

tomasherman:
nvm, fix is actually quite easy

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -381,7 +381,7 @@ object CirceProtocolGenerator {
       case ExtractSuperClass(swagger, definitions) =>
         def allParents: Tracker[Schema[_]] => Target[List[(String, Tracker[Schema[_]], List[Tracker[Schema[_]]])]] =
           _.refine[Target[List[(String, Tracker[Schema[_]], List[Tracker[Schema[_]]])]]]({ case x: ComposedSchema => x })(
-            _.downField("allOf", _.getAllOf()).indexedDistribute match {
+            _.downField("allOf", _.getAllOf().asScala.toList.filter(s => s.get$ref() != null)).indexedDistribute match {
```
```suggestion
            _.downField("allOf", _.getAllOf()).indexedDistribute.filter(_.downField("$ref", _.get$ref()).unwrapTracker.nonEmpty) match {
```

blast-hardcheese: Pretty sure this will work, I'm on mobile right now so I can't test. Leaving the comment in case you get to it before I do.

blast-hardcheese: What's nice about it is that `indexedDistribute` happens before the filter, so array index is preserved instead of being the index into items after filtering

blast-hardcheese: (this is part of the new infrastructure introduced in #402, so I'm not expecting everyone to know how it works yet)

tomasherman: thanks for the comment, i will apply it later today

PR #424, opened 2019-10-21
Only create params for form request bodies

### LINKED ISSUE #313: JSON body param with embedded schema generates FormData
Given the following spec with an embedded schema for the in body, Guardrail generates a client method (see below) that takes `name` and `body` and uses them as form parameters. It also takes a `body: io.circe.Json` which is ignored.

This could be related to #237, but since we're consuming JSON here, the form params seem definitely wrong here. It also clashes with the `name` path parameter.

As soon as I move the schema into a definition and reference that, everything works fine.

Ideally, Guardrail would generate a case class and codecs for the embedded schema (perhaps using the `name` field for the class name?). At least it should not create form params but just use the JSON body instead.

```yaml
swagger: '2.0'
info:
  title: JupyterHub
  description: The REST API for JupyterHub
produces:
  - application/json
consumes:
  - application/json
paths:
  /users/{name}:
    patch:
      operationId: changeUser
      summary: Modify a user
      description: Change a user's name or admin status
      parameters:
        - name: name
          description: username
          in: path
          required: true
          type: string
        - name: data
          in: body
          required: true
          description: Updated user info. At least one key to be updated (name or admin) is required.
          schema:
            type: object
            properties:
              name:
                type: string
                description: the new name (optional, if another key is updated i.e. admin)
              admin:
                type: boolean
                description: update admin (optional, if another key is updated i.e. name)
      responses:
        '200':
          description: The updated user info
```

```scala
  def changeUser(name: String, name: String, admin: Boolean, body: io.circe.Json, headers: List[HttpHeader] = Nil): EitherT[Future, Either[Throwable, HttpResponse], ChangeUserResponse] = {
    val allHeaders = headers ++ scala.collection.immutable.Seq[Option[HttpHeader]]().flatten
    makeRequest(HttpMethods.PATCH, host + basePath + "/users/" + Formatter.addPath(name), allHeaders, FormData(List(List(("name", Formatter.show(name))), List(("admin", Formatter.show(admin)))).flatten: _*), HttpProtocols.`HTTP/1.1`).flatMap(req => EitherT(httpClient(req).flatMap(resp => resp.status match {
      case StatusCodes.OK =>
        resp.discardEntityBytes().future.map(_ => Right(ChangeUserResponse.OK))
      case _ =>
        FastFuture.successful(Left(Right(resp)))
    }).recover({
      case e: Throwable =>
        Left(Left(e))
    })))
  }
```

sbrunk:
A simple attempt to fix #313, without opening the box of trying to properly support multiple mime types in a single request. :)

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/test/scala/core/issues/Issue313.scala
+++ b/modules/codegen/src/test/scala/core/issues/Issue313.scala
@@ -1,0 +1,75 @@
+package core.issues
+
+import com.twilio.guardrail.generators.syntax.Scala.companionForStaticDefns
+import com.twilio.guardrail.generators.AkkaHttp
+import com.twilio.guardrail.{Client, Clients, Context}
+import org.scalatest.{FunSuite, Matchers}
+import support.SwaggerSpecRunner
+
+import scala.meta._
+
+class Issue313 extends FunSuite with Matchers with SwaggerSpecRunner {
+
+  val swagger: String = s"""
+    |openapi: 3.0.1
+    |info:
+    |  title: Whatever
+    |  version: 1.0.0
+    |paths:
+    |  /pets:
+    |    post:
+    |      operationId: addPets
+    |      requestBody:
+    |        required: true
+    |        content:
+    |          application/json:
+    |            schema:
+    |              type: object
+    |              properties:
+    |                id:
+    |                  type: integer
+    |                name:
+    |                  type: string
+    |      responses:
+    |        201:
+    |          description: Created
+    |          content: {}
+    |""".stripMargin
+
+  test("Don't generate params for non-form request bodies") {
+
+    val (
+      _,
+      Clients(Client(_, _, _, staticDefns, cls, _) :: _, Nil),
+      _
+      )       = runSwaggerSpec(swagger)(Context.empty, AkkaHttp)
+
+    val client = q"""
+      class Client(host: String)(implicit httpClient: HttpRequest => Future[HttpResponse], ec: ExecutionContext, mat: Materializer) {
+        val basePath: String = ""
+        private[this] def makeRequest[T: ToEntityMarshaller](method: HttpMethod, uri: Uri, headers: scala.collection.immutable.Seq[HttpHeader], entity: T, protocol: HttpProtocol): EitherT[Future, Either[Throwable, HttpResponse], HttpRequest] = {
+          EitherT(Marshal(entity).to[RequestEntity].map[Either[Either[Throwable, HttpResponse], HttpRequest]] {
+            entity => Right(HttpRequest(method = method, uri = uri, headers = headers, entity = entity, protocol = protocol))
+          }.recover({
+            case t =>
+              Left(Left(t))
+          }))
+        }
+        def addPets(body: io.circe.Json, headers: List[HttpHeader] = Nil): EitherT[Future, Either[Throwable, HttpResponse], AddPetsResponse] = {
+          val allHeaders = headers ++ scala.collection.immutable.Seq[Option[HttpHeader]]().flatten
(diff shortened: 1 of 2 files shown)


modules/codegen/src/test/scala/core/issues/Issue313.scala
```diff
@@ -0,0 +1,75 @@
...
+        val basePath: String = ""
+        private[this] def makeRequest[T: ToEntityMarshaller](method: HttpMethod, uri: Uri, headers: scala.collection.immutable.Seq[HttpHeader], entity: T, protocol: HttpProtocol): EitherT[Future, Either[Throwable, HttpResponse], HttpRequest] = {
+          EitherT(Marshal(entity).to[RequestEntity].map[Either[Either[Throwable, HttpResponse], HttpRequest]] {
+            entity => Right(HttpRequest(method = method, uri = uri, headers = headers, entity = entity, protocol = protocol))
+          }.recover({
+            case t =>
+              Left(Left(t))
+          }))
+        }
+        def addPets(body: io.circe.Json, headers: List[HttpHeader] = Nil): EitherT[Future, Either[Throwable, HttpResponse], AddPetsResponse] = {
```
blast-hardcheese: I think we want to aim in the opposite direction from this, synthesizing `body: (BigInt, String)` instead of throwing away that information. If you wouldn't mind seeing how difficult it would be to go in that direction, that would be very much appreciated.

sbrunk: I completely agree that we should eventually synthesize typed request bodies and I'd be happy to give it a try. This was just meant as a quick fix to improve the current situation for JSON bodies, because right now they are just completely broken (see #313).

sbrunk:
@blast-hardcheese we can have inline object literals inside a `requestBody` which look just like schemas/definitions used as ref. I.e. they might be nested.

So I imagine we could to generate definitions for inline objects inside requests, reusing `ProtocolGenerator`. Example:

```yaml
operationId: addPets
requestBody:
  required: true
  content:
    application/json:
      schema:
      type: object
      properties:
        id:
          type: integer
        name:
          type: string
```
Could generate s.th. roughly like this:
```scala
object AddPetsRequest {
  case class JsonRequestBody(id: Int, name: String)
  // more definitions if we have nested objects
}
// client and server handler work with JsonRequestBody
def addPets(body: AddPetsRequest.JsonRequestBody, ...)
```

Later, this could also be extended to support multiple mime types in one request:

```scala
object AddPetsRequest {
  sealed abstract class RequestBody
  case class JsonRequestBody(id: Int, name: String) extends RequestBody
  case class FormRequestBody() extends RequestBody
}
def addPets(body: RequestBody, ...)
```

Does that make sense?

blast-hardcheese:
@sbrunk I hadn't considered that, I agree with your assessment. I was just thinking of form bodies, but you're right, this could be literals nested deeply.

Out of curiosity, have you been able to work around this by just passing unused parameters in, or is there a name conflict?

sbrunk:
In the case described in #313 there's a name conflict so the generated code doesn't compile.

In cases without a name conflict it generates a client that makes a form request from the parameters, ignoring the JSON body which makes it fail on the (non guardrail) server who expects JSON.

So my current workaround ist to actually comment out properties of schema definition, because in that case it uses the untyped JSON body.
```yaml
requestBody:
  content:
    application/json:
      schema:
        type: object

blast-hardcheese:
Revisiting this, regrettably quite late at this time. I've applied the spirit of the approach you took here in #1078. As a result, closing this out, with an apology for it taking this long to resolve, and appreciation for your contribution.

PR #430, opened 2019-10-31
Update Scalameta. Fixes #429

### LINKED ISSUE #429: Enum values are not enclosed in backticks, preventing correct code generation for certain patterns
Exemple:
```json
	 "returned_anwsers_count": {
          "type": "string",
          "enum": [
            "10"
          ]
        }
```
results in:

```scala
  object ReturnedAnwsersCount {
    object members { case object 10 extends ReturnedAnwsersCount("10") }
    val 10: ReturnedAnwsersCount = members.10
    val values = Vector(10)
```

Which is not valid Scala code

It may be

```scala
  object ReturnedAnwsersCount {
    object members { case object `10` extends ReturnedAnwsersCount("10") }
    val `10`: ReturnedAnwsersCount = members.`10`
    val values = Vector(10)
```

sbrunk:
**Contributing to Twilio**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -210,7 +210,7 @@ lazy val codegen = (project in file("modules/codegen"))
     (name := "guardrail") +:
       codegenSettings,
     libraryDependencies ++= testDependencies ++ Seq(
-      "org.scalameta"               %% "scalameta"                    % "4.1.0",
+      "org.scalameta"               %% "scalameta"                    % "4.2.4",
       "com.github.javaparser"       % "javaparser-symbol-solver-core" % javaparserVersion,
       "org.eclipse.jdt"             % "org.eclipse.jdt.core"          % "3.17.0",
       "org.eclipse.platform"        % "org.eclipse.equinox.app"       % "1.3.600",
--- a/modules/codegen/src/test/scala/core/issues/Issue429.scala
+++ b/modules/codegen/src/test/scala/core/issues/Issue429.scala
@@ -1,0 +1,52 @@
+package core.issues
+
+import com.twilio.guardrail.generators.Http4s
+import com.twilio.guardrail.{ClassDefinition, Context, ProtocolDefinitions}
+import org.scalatest.{FunSuite, Matchers}
+import support.SwaggerSpecRunner
+
+import scala.meta._
+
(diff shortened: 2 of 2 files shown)


blast-hardcheese:
To avoid regression, would you mind adding a small specfile that tests this functionality?

sbrunk:
I actually have a test ready but it passes even with the old version for some reason when comparing the structures (generating files shows the difference though). I'll add the test nevertheless. Perhaps you have an idea what needs to be compared to show the behavior.

blast-hardcheese:
my thought would be that `Lit.Int(10)` should be used anywhere, whereas `Term.Name("10")` should always fail. I presume the scala.meta upgrade would fix this by escaping the `Term.Name("10")`

sbrunk:
```scala
@ import $ivy.`org.scalameta::scalameta:4.1.0`, scala.meta._
@ Term.Name("10").syntax
res1: String = "10"
```

```scala
@ import $ivy.`org.scalameta::scalameta:4.2.4`, scala.meta._
@ Term.Name("10").syntax
res1: String = "`10`"
@ Defn.Object(List(), Term.Name("10"), Template(List(), List(), Self(Name(""), None), List())).syntax
res2: String = "object `10`"
```

So perhaps we could simply add
```scala
Term.Name("10").syntax shouldBe "`10`"
```
as additional test condition.

blast-hardcheese:
@sbrunk I'm onboard with that, though still interested in why https://github.com/twilio/guardrail/pull/430#issuecomment-548574779 is the case

sbrunk:
> my thought would be that `Lit.Int(10)` should be used anywhere, whereas `Term.Name("10")` should always fail. I presume the scala.meta upgrade would fix this by escaping the `Term.Name("10")`

I'm not sure I understand. Why should `Lit.Int(10)` be used anywhere?

blast-hardcheese:
@sbrunk This is excellent, thank you for your investigation! I'll try to get a bugfix release out with this at some point this weekend.

cc @nb-ceffa, @tomasherman

tomasherman:
Thank you!

sbrunk:
If only every bugfix was that simple 😀

PR #434, opened 2019-11-04
Fully quallified name

tomasherman:
Hey guys, 

so this is gonna be a tough one. I inherited this branch that @hanny24 made before he left for his vacation and we released guardrail internally with it and now we would like to merge it to original guardrail so that we don't get out of sync too much. We rely on it in production but im not 100% sure what it solves :)

I think it tries to solve some conflicts in objects created by referencing them all using fully qualified names. It's quite crucial for us.

I will try to figure out more details later.


blast-hardcheese:
Woah, if that's the case, that is awesome! A cursory review doesn't raise any red flags, other than a missing test for the generated code to prevent regressions. Also, if we truly are fully qualifying all DTO packages, prepending `_root_` to ensure no conflicts on the root package (for instance, `io example` vs `scala.io` is disambiguated by `_root_.io.example`)

tomasherman:
So i updated the code to add `_root_` at the beginning of FQN and added a test, not sure if i did the test correctly but i tried my best ;)

tomasherman:
Also, not sure if 'awesome' is the word I would use :D

blast-hardcheese:
"awesome" in the West Coast American usage, meaning "probably very good, and effort is appreciated" :wink:

Looks good! I'll do a thorough review in the next full days, but no objection to merging.

tomasherman:
sure, i also updated the previously merged code to be bound to `_root_` here #435

blast-hardcheese:
Why didn't the failed test cause this build to fail?

blast-hardcheese:
Lol, `# TODO enable javaTestSuite`

tomasherman:
I am really sorry about this :( is it something that this PR introduced? It totally slipped by me

blast-hardcheese:
@tomasherman Not a concern! There were two issues here:
1) Both of your PRs passed PR validation (got the check from Travis and Github Actions), but I didn't realize the second would conflict with the first. When I merged them both, it broke master due to my impatience.
2) The PR actually did break Java codegen, as `_root_` is invalid Java, but that was added to the language-agnostic part of the codegen. Due to this issue, however, it caused me to discover that Java tests have been disabled in `.travis.yml`! #444 resolves this, but thank you for pointing it out!

I'm trying to resolve the build failures currently, but it seems as though there's a _third_ issue, which is that my local compilation environment is not fully cleaned whenever I do `resetSample` -- I presumed that SBT was smart enough to only feed `.class` files _it_ produced during the most recent compilation activity into scalac for linking, but it turns out that previous class files were actually sticking around, giving false positives for future test runs. I'm resolving that now as well.

tomasherman:
ok :) if there is something i can do, let me know ... we are increasingly relying on guardrail more and more here at avast (at least in my little corner of avast universe) so we are happy to help where we can

blast-hardcheese:
@tomasherman If you wouldn't mind reviewing #444, since you've recently seen that part of the code recently, that'd be appreciated 👍

I'll try to merge tomorrow, once some time has passed to get some other eyes on it

PR #439, opened 2019-11-10
Fqns for scala types

### LINKED ISSUE #438: Generating class which name is 'Option'
I have swagger docs which has some conflicting part:
```json
"option": {
          "description": "Some options blabla",
          "type": "object",
          "properties": {
            "inventorySource": {bla bla}
           }
}
```

So by default generated case class for this object looks like this:
`case class Option(...)`
which obviously conflicts with scala Option. Is it a way to generate definiton class with some prefix to avoid collision like this ?

tomasherman:
attempt to fix #438

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -68,6 +68,7 @@ val exampleCases: List[ExampleCase] = List(
   ExampleCase(sampleResource("multipart-form-data.yaml"), "multipartFormData"),
   ExampleCase(sampleResource("petstore.json"), "examples").args("--import", "support.PositiveLong"),
   ExampleCase(sampleResource("plain.json"), "tests.dtos"),
+  ExampleCase(sampleResource("fqns.yaml"), "tests.fqns").frameworks(Set("http4s")),
   ExampleCase(sampleResource("polymorphism.yaml"), "polymorphism"),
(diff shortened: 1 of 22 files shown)


tomasherman:
i think this is dead end

blast-hardcheese:
Understood, thanks for investigating! Are there any learnings that can be applied elsewhere?

tomasherman:
I think the takeaway for me is that the way to go is to fully qualify types
that we generated instead of scala stdlib and not use wild card imports


On 16 November 2019 at 22:04:19, Devon Stewart (notifications@github.com)
wrote:

Understood, thanks for investigating! Are there any learnings that can be
applied elsewhere?

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/twilio/guardrail/pull/439?email_source=notifications&email_token=AAAYMHVHALSJB64FT2RCEMLQUBN5HA5CNFSM4JLLIB22YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEEH2QEI#issuecomment-554674193>,
or unsubscribe

PR #442, opened 2019-11-10
bug fix: use 'camelCase' for Java variable name

sullis:
This fixes a bug in the Java code generator.

I observed a line of code that could not be compiled by the Java compiler:

```
  public static final BadRequest BadRequest = new BadRequest();
```

The Java compiler reports an error because the variable name is the same as the class name.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
@@ -16,6 +16,7 @@ import com.twilio.guardrail.{ ADT, ClassDefinition, EnumDefinition, RandomType, 
 import com.twilio.guardrail.extract.ServerRawResponse
 import com.twilio.guardrail.generators.{ ScalaParameter, ScalaParameters }
 import com.twilio.guardrail.generators.syntax.Java._
+import com.twilio.guardrail.generators.syntax.RichString
 import com.twilio.guardrail.languages.JavaLanguage
 import com.twilio.guardrail.protocol.terms.Response
 import com.twilio.guardrail.protocol.terms.server._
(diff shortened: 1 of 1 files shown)


sullis:
@blast-hardcheese   ready for review

blast-hardcheese:
So, I'm not running into this error -- There's no name collision because field members (as members of `${operationId}Response`) are in a different namespace from nested classes (`${operationId}Response.${statusCodeName}`)

What's the actual compiler error here?

sullis:
This is the error that I am trying to fix:

```

[error] /Users/user123/github/sbt-guardrail-java-codegen-example/target/scala-2.12/src_managed/main/com/example/servers/petstore/user/UserHandler.java:204:40: BadRequest is already defined as class BadRequest
[error]         public static final BadRequest BadRequest = new BadRequest();

```

The Java source file [UserHandler.java] has a syntax error.

sullis:
On [master], the Dropwizard Server Java Generator produces this line of code:

`public static final BadRequest BadRequest = new BadRequest();`

With my change, the Dropwizard Server Java Generator produces this line of code:

`public static final BadRequest badRequest = new BadRequest();`

sullis:
How to reproduce the compile error:

```
$ git clone https://github.com/sullis/sbt-guardrail-java-codegen-example.git
$ cd sbt-guardrail-java-codegen-example
$ sbt clean compile
```

kelnos:
So the intent is for it to be PascalCase, and I haven't seen this error either.  Not sure why you're seeing one, because in Java, the namespaces for variable names and class names (and method names) are different.

blast-hardcheese:
@sullis Ah, so in your example repo, you have

```json
        "responses": {
          "400": {
            "description": "Invalid ID supplied"
          },
          "404": {
            "description": "Pet not found"
          },
          "400": {
            "description": "Validation exception"
          }
        }
```
which highlights a different bug, but one that wouldn't be solved with camel vs Pascal case

sullis:
fyi: file [https://github.com/sullis/sbt-guardrail-java-codegen-example/blob/master/petstore.json] was originally copied from [https://github.com/twilio/guardrail/blob/master/modules/sample/src/main/resources/petstore.json]

blast-hardcheese:
oh jeez. Is this a feature? Arbitrary, possibly conflicting schemas for the same status code?

kelnos:
I don't see what that would be a feature, since you can specify arbitrary, conflicting schemas for the same status code without repeating it.

blast-hardcheese:
@sullis So, just to confirm -- there isn't actually a problem with the codegen, even with the camel case we'd end up with the same bug. Do you agree?

blast-hardcheese:
Closing this to cut down on noise, please feel free to reopen if you feel strongly about it.

PR #444, opened 2019-11-11
Fix build and tests

### LINKED ISSUE #443: Changes in PR #435 to make FQDN for StausCodes breaks test in PR #434
The above two PRs appear to have been merged into master, although he second is breaking a test in the first: 

> Test that fully qualified names are used

`httpClient.fetch(req)({
             case org.http4s.Status.Ok(resp) =>
               F.map(getUserOkDecoder.decode(resp, strict = false).value.flatMap(F.fromEither))(GetUserResponse.Ok.apply)
             case resp =>
               F.raiseError(UnexpectedStatus(resp.status))
})`

`org.http4s.Status.Ok(resp)` should become `_root_.org.http4s.Status.Ok(resp)`, however I'm raising this issue to understand what problem existed by just using the FQDN for `org.http4s.Status.Ok`

blast-hardcheese:
Resolves #443 

So, master started failing to compile after merging #434 due to some Scala-specific changes being applied directly in the "core" interpreter. This wasn't caught earlier because the Java tests haven't been run during the PR validation process.

- [x] Switching from `scalaTestSuite` to `testSuite` in travis and github actions
- [x] Adding `codegen:test` formatting check
- [x] Moving `_root_` package prefix to `ScalaGenerator` (identity in `JavaGenerator`)
- [x] Only add `definitions` imports if there are actually definitions (avoid errors during Java codegen if there are no definitions)

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaGenerator.scala
@@ -29,20 +29,20 @@ object ScalaGenerator {
     def apply[T](term: ScalaTerm[ScalaLanguage, T]): Target[T] = term match {
       case VendorPrefixes() => Target.pure(List("x-scala", "x-jvm"))
 
-      case LitString(value)        => Target.pure(Lit.String(value))
-      case LitFloat(value)         => Target.pure(Lit.Float(value))
-      case LitDouble(value)        => Target.pure(Lit.Double(value))
-      case LitInt(value)           => Target.pure(Lit.Int(value))
-      case LitLong(value)          => Target.pure(Lit.Long(value))
-      case LitBoolean(value)       => Target.pure(Lit.Boolean(value))
-      case LiftOptionalType(value) => Target.pure(t"Option[${value}]")
-      case LiftOptionalTerm(value) => Target.pure(q"Option(${value})")
-      case EmptyOptionalTerm()     => Target.pure(q"None")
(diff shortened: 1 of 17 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaGenerator.scala
```diff
@@ -398,7 +400,7 @@ object ScalaGenerator {
               ..${extraImports}
               import ${buildPkgTerm(List("_root_") ++ pkgName ++ List("Implicits"))}._
               ..${frameworkImplicitName.map(name => q"import ${buildPkgTerm(List("_root_") ++ pkgName)}.${name}._")}
-              import ${buildPkgTerm(List("_root_") ++ dtoComponents)}._
+              ..${dtoComponents.map(x => q"import ${buildPkgTerm(List("_root_") ++ x)}._")}
```
tomasherman: as a side note - i think this is the main reason we are having trouble with swagger definitions named `Option` etc ... if we avoided these imports and used fully qualified imports in the generated code, we would be safe - think

blast-hardcheese: Yeah; After the qualified DTO PR was merged, I actually don't think this explicit import is even required anymore, other than people using `x-scala-type: List[Foo]`, not having a good way to do effectively `x-scala-type: List[#/definitions/Foo]`

PR #448, opened 2019-11-13
Add support for http4s 0.21.x

tonyvelichko:
Starting the version 0.21 http4s requires to explicitly provide a transformation functor from G to F in EntityResponseGenerator. The following chores allow compiling generated code using both 0.20 and 0.21 http4s version. 

**Contributing to Twilio**

- [ √] I acknowledge that all my contributions will be made under the project's license.


tonyvelichko:
Probably we could try to use http4s-directives to simplify generated code, which is already handling the effect conversion for request/response [F] and their bodies [G]. But even with that in 100% cases, F is G for generated code.  

> Could we leverage liftG to provide better better functionality now?
Probably not yet.

blast-hardcheese:
Understood, thank you!

Issue #455, opened 2019-11-17
NPE on encoding/decoding nested recursive data structures 


azhur:
guardrail generates `forProductN` circe encoder/decoder that throws NPE for recursive data structures like:
```yaml
openapi: 3.0.0
info:
  title: Test Server
  version: "1.0"
servers:
  - url: http://localhost:8000
paths:
  /v1/Boo:
    post:
      summary: Test resource
      operationId: boo
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/RecursiveData"
      responses:
        200:
          description: Ok
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/RecursiveData"

components:
  schemas:
    RecursiveData:
      type: object
      description: Represents a recursive nested object
      required: [id, display_name]
      properties:
        id:
          type: integer
          format: int64
        display_name:
          type: string
        nesting:
          $ref: "#/components/schemas/RecursiveData"
```

related circe bug: https://github.com/circe/circe/issues/561

blast-hardcheese:
@azhur Yikes. Thanks for the spec!

blast-hardcheese:
Resolved in #456

PR #456, opened 2019-11-17
Recursive definitions generate NPE due to Circe bug

blast-hardcheese:
Resolving #455 

- Disables the `forProductN` circe encoders/decoders to work around the bug
- Switching the expanded form of object decoder from using the term names themselves internally in the decoder to using `q"v$n"` (v0, v1, v2, ...) to avoid name confusion (Capitalized names has different semantic meaning in Scala than lower cased ones, trying to resolve this could cause name collisions).

**Contributing to Twilio**


blast-hardcheese:
FYI @azhur

blast-hardcheese:
Formatting was required

PR #468, opened 2019-11-29
GitHub CI: enable 'cache' action

sullis:
- SBT 1.3.x uses Coursier for dependency resolution
- we preserve Coursier's cache to achieve faster builds

Docs:
- https://github.com/actions/cache
- https://get-coursier.io/docs/cache

**Contributing to Twilio**


.github/workflows/ci.yml
```diff
@@ -23,5 +23,11 @@ jobs:
         java-version: ${{ matrix.java }}
     - name: print Java version
       run: java -version
+    - uses: actions/cache@v1.0.3
+      with:
+        path: ~/.cache/coursier
+        key: ${{ runner.os }}-sbt-${{ hashFiles('**/ivy.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-sbt-
```
blast-hardcheese: `restore-keys` is a prefix, such that it matches everything except for `${{ hashFiles('**/ivy.xml') }}`?

blast-hardcheese: It doesn't seem like it, how is this expected to be used?

sullis: I just pushed a fix.   I changed it to *.sbt

blast-hardcheese: Question still stands wrt `restore-keys`:

> restore-keys - An ordered list of keys to use for restoring the cache if no cache hit occurred for key

How does the initial one/fallback one get cached?

sullis: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows

```
The cache action will attempt to restore a cache based on the key you provide. When the action finds a cache it restores the cached files to the path you configure.

If no exact match is found, the action creates a new cache entry on the successful completion of the job. The new cache will use the key you provided and contains the files in the path directory.

You can optionally provide a list of restore-keys that are used when the key doesn't match an existing cache. A list of restore-keys is useful when you are restoring a cache from another branch in a repository because restore-keys can partially match cache keys. 

```

sullis: 'restore-keys' is optional.  I'm going to remove it.

blast-hardcheese: > restore-keys can partially match cache keys.
👍

sullis:
LGTM

Screen Shot 2019-11-29 at 4 10 25 PM

PR #477, opened 2019-12-03
Use Optional.ofNullable instead of .of where appropriate

kelnos:
Avoids runtime exceptions with the trade off that if you don't expect something you pass to be `null`, and it is, you might be confused.

**Contributing to Twilio**


kelnos:
Keeping me honest ;) Added a test

sullis:
LGTM

PR #493, opened 2019-12-11
Make WrapToObject optional

kelnos:
Java doesn't need it, as all declarations should end up as static members of their outer class.

**Contributing to Twilio**


kelnos:
Added a test.  It's a little lame as it just makes sure the final object is constructable and the nested classes are in the right place, but it'll do the job for now.

kelnos:
Ok nevermind I made the test a little less lame and a little more useful.

kelnos:
Not sure what's up with the tests.  They did finish and did complete successfully, but GH seems to not know about that.

PR #504, opened 2019-12-18
Trying a little harder to pull out a useful error message from ParseProblemException

blast-hardcheese:
Instead of overloading the user with mountains of stack traces, just print what the problem is

**Contributing to Twilio**


blast-hardcheese:
After a big of time away from the problem, I came back a bit later to give it another shot and write some tests. I think it's better structured now, and has the benefit of not throwing NPEs during the error handler (🎉)

Issue #239, opened 2019-04-16
use built-in `OffsetDateTimeParam` etc. for jsr310 types


kelnos:
Currently we emit an injection binder to allow JSR310 types as query/form/header params in annotated resource methods. [dropwizard-jersey](https://github.com/dropwizard/dropwizard/tree/master/dropwizard-jersey/src/main/java/io/dropwizard/jersey/params) appears to support them natively, but via using `${Type}Param` as the argument type.

I'm not a fan of those, but we could use them, and then "unwrap" them before passing them to the handler so the handler interface remains sane.

kelnos:
Issues:
* The `AbstractParam` base class appears to throw a `WebApplicationException` with a hard-coded JSON response on error.  This precludes the possibility of someone using an `ExceptionMapper` in their service to return an error response that conforms to their openapi spec.
* These `*Param` classes were only merged into the main DW code as of DW 1.0.  DW 0.7-0.9 can use [dropwizard-java8](https://github.com/dropwizard/dropwizard-java8) (which the newer code is based on), but of course they changed the package names when merging into mainline, so any solution will have to use guardrail modules to reflect the different package names.

blast-hardcheese:
@kelnos Was this resolved fully in #506, or is there still work to be done?

kelnos:
Should be done

PR #506, opened 2019-12-19
Replace JSR310 injection providers with DW AbstractParam

kelnos:
We can't use DW's built-in `AbstractParam` subclasses because they always return a JSON error response that can't be overridden in service code without a lot of pain.

This also allows us to avoid needing separate guardrail modules for DW7-9 and DW10+, as the JSR310 `AbstractParam` subclasses are in separate artifacts and have different package names for the two version ranges.

Addresses https://github.com/twilio/guardrail/issues/239

**Contributing to Twilio**


kelnos:
Hmm, it looks like the http4s tests are failing now (compiling the new test spec) due to ambiguous implicits.  @blast-hardcheese any ideas?

kelnos:
Excellent, you did indeed have ideas :D

PR #532, opened 2020-02-02
improve parameter error messages

sullis:
Include the parameter name in the  error messages

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/SwaggerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/SwaggerGenerator.scala
@@ -16,11 +16,13 @@ import scala.collection.JavaConverters._
 import scala.util.Try
 
 object SwaggerGenerator {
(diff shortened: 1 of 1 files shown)


blast-hardcheese:
What was the previous error message, `...parameter[2]` or something?

sullis:
The previous message was

`Parameter has no schema`

This PR adds the parameter name to the message.

blast-hardcheese:
This uses the `raiseErrorIfEmpty` mechanism though, it should provide a path to the element that failed validation immediately after the error message. Would you mind confirming whether it's there or not?

sullis:
Without this PR:

```
Error: Parameter has no schema type (.paths./v2/Apples/Oranges.operations.POST.requestBody[0].content.application/json.schema.type)
```

With this PR:

```
Error: Parameter chef_environments has no schema type (.paths./v2/Apples/Oranges.operations.POST.requestBody[0].content.application/json.schema.type)
```

blast-hardcheese:
Ah, that is better.
We should still give a useful message even if the `name` parameter is `null` though, other than that :+1: from me.

PR #535, opened 2020-02-03
Ignoring default response code

blast-hardcheese:
- Removing `Currently not supported for Java` in favor of just trusting the generated static class
- Removing `Unknown HTTP status code: default` in favor of ignoring it and emitting a warning

Unfortunately, while StructuredLogger is disabled, this effectively removes any capacity to actually emit these logs, while the performance refactoring is going on.

**Contributing to Twilio**


blast-hardcheese:
Closing this as it's stale. If there is a reason to revisit this it can be done in the future, either way it'll need to be rewritten ontop of the current `master`

PR #539, opened 2020-02-07
Issue538

### LINKED ISSUE #538: Double inheritance of same field
Consider following definition:
```yaml
swagger: '2.0'
definitions:
  Bar:
    properties:
      id:
        type: string
  Baz:
    properties:
      id:
        type: string
    required: [id]
  Foo:
    allOf:
      - $ref: "#/definitions/Bar"
      - $ref: "#/definitions/Baz"
```

Code generated is as follows:
`case class Foo(id: String, id: String)`, which is obviously not a valid Scala

I think that when two definitions are included, fields with same name should be deduplicated  if name and types are same - with rules that if either one of them is required, the result is required aswell. If both are optional, result is optional.

Do you guys agree? If yes, I will implement this fix

tomasherman:
fixed #538 

**Contributing to Twilio**

- [ ] I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
@@ -5,16 +5,17 @@ import _root_.io.swagger.v3.oas.models.media._
 import cats.data.NonEmptyList
 import cats.free.Free
 import cats.implicits._
-import com.twilio.guardrail.core.{ Mappish, Tracker }
+import com.twilio.guardrail.core.{Mappish, Tracker}
 import com.twilio.guardrail.core.implicits._
 import com.twilio.guardrail.generators.RawParameterType
 import com.twilio.guardrail.generators.syntax._
 import com.twilio.guardrail.languages.LA
 import com.twilio.guardrail.protocol.terms.protocol._
 import com.twilio.guardrail.terms.framework.FrameworkTerms
-import com.twilio.guardrail.terms.{ ScalaTerms, SwaggerTerms }
+import com.twilio.guardrail.terms.{ScalaTerms, SwaggerTerms}
 import java.util.Locale
 
+import cats.Eq
 import com.twilio.guardrail.extract.Default
 
 import scala.collection.JavaConverters._
@@ -44,7 +45,21 @@ case class ProtocolParameter[L <: LA](
     dataRedaction: RedactionBehaviour,
     defaultValue: Option[L#Term]
 )
-
+object ProtocolParameter {
+  implicit def eqInstance[L <: com.twilio.guardrail.languages.LA](implicit eqTerm: Eq[L#Term], eqMethodParam: Eq[L#MethodParameter]) = new Eq[ProtocolParameter[L]] {
+    override def eqv(x: ProtocolParameter[L],
+                     y: ProtocolParameter[L]): Boolean = {
+      Eq.eqv(x.term, y.term) &&
+        x.name == y.name &&
+        x.dep == y.dep &&
+        x.rawType == y.rawType &&
+        x.readOnlyKey == y.readOnlyKey &&
+        x.emptyToNull == y.emptyToNull &&
+        x.dataRedaction == y.dataRedaction &&
+        Eq.eqv(x.defaultValue, y.defaultValue)
+    }
+  }
+}
 case class Discriminator[L <: LA](propertyName: String, mapping: Map[String, ProtocolElems[L]])
 
 object Discriminator {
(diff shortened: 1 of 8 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
```diff
@@ -173,7 +174,13 @@ object CirceProtocolGenerator {
...
           param => discriminatorNames.contains(param.term.name.value)
         )
-        val terms = params.map(_.term)
+
+        import ScalaLanguage._
+
+        val deduplicatedParams = params.foldLeft(List.empty[ProtocolParameter[ScalaLanguage]]) { (s, a) =>
+          s.find(p => Eq.eqv(a, p)).map(_ => s).getOrElse(a :: s)
+        }
+        val terms = deduplicatedParams.map(_.term).reverse // not needed but without it we break bunch of tests
```
kelnos: Is it possible to do this just in ProtocolGenerator rather than needing to do this for each concrete impl?  JacksonGenerator will need this too.

Also I think we should just bail with an error if there are two params with the same name and they aren't "compatible".

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
```diff
@@ -44,7 +45,20 @@ case class ProtocolParameter[L <: LA](
...
+    new Eq[ProtocolParameter[L]] {
+      override def eqv(x: ProtocolParameter[L], y: ProtocolParameter[L]): Boolean =
+        Eq.eqv(x.term, y.term) &&
+          x.name == y.name &&
+          x.dep == y.dep &&
+          x.rawType == y.rawType &&
+          x.readOnlyKey == y.readOnlyKey &&
+          x.emptyToNull == y.emptyToNull &&
+          x.dataRedaction == y.dataRedaction &&
+          Eq.eqv(x.defaultValue, y.defaultValue)
```
kelnos: This is probably stricter than it needs to be.  We can "widen" things in some situations, like if `readOnlyKey` doesn't match, we can set it to `false`.  If `emptyToNull` or `dataRedaction` doesn't match, we can set them to the `true`-equivalent.  I'm torn on `defaultValue`; probably should require them to be the same to avoid unexpected behavior (though maybe we allow it if one has no default value and the other does?).  Required vs. optional is also a good question... I could see the argument for collapsing that to either required or optional if they differ.

Overall I would say if two parameters have the same name, we can de-dupe them as long as they have the same types and same default value.  If they don't, then we should terminate with a hard error.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
tomasherman: well it would be possible to move this logic to ProtocolGenerator, but this would require larger refactoring - as of now, there is already a lot of specific logic regarding how extensions and parents are handled - i don't feel qualified for that refactoring

do you feel that this is required before for merge to be considered? if so, i can take a stab at it but honestly i would rather just solve my own problem

regarding the error, that is a fair point

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
tomasherman: good points

modules/codegen/src/main/scala/com/twilio/guardrail/generators/CirceProtocolGenerator.scala
kelnos: That's fair.  I've been eyeing some of the redundant protocolgenerator code for a refactor for a while now, so I might look into it after this PR lands.  No need to tackle that before getting this in, though.

tomasherman: ok, i will attack your other concern then

hanny24:
So I took over this PR. Summary of the changes I have done:
- implementation should be generic, so it is deduplicated for each framework
- it takes into account parameter name, type and default value
- if a parameter was marked as redacted at least once, the merged parameter is treated as redacted as well
- similarly we prefer to preserve `x-empty-is-null: true`
- there is still one thing that is not done properly: https://github.com/twilio/guardrail/pull/539/files#diff-f0ed213f3160ee8faee416537f0c544eR470 I'm not really sure how to raise an error cleanly there

blast-hardcheese:
@hanny24 Would you please permit outside collaborators on this branch? I've got a fix for the `throw` that I would like to push, but it seems disallowed.

blast-hardcheese:
Just gonna open a separate PR actually.

blast-hardcheese:
Merged

PR #542, opened 2020-02-10
Handle byte string format according to specification

hanny24:
**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
@@ -281,7 +281,7 @@ object SwaggerUtil {
             case (Some("string"), Some("email"))        => stringType(None)
             case (Some("string"), Some("date"))         => dateType()
             case (Some("string"), Some("date-time"))    => dateTimeType()
-            case (Some("string"), Some("byte"))         => pureTypeName("Base64String").flatMap(widenTypeName)
+            case (Some("string"), Some("byte"))         => bytesType()
             case (Some("string"), fmt @ Some("binary")) => fileType(None).map(log(fmt, _))
             case (Some("string"), fmt)                  => stringType(fmt).map(log(fmt, _))
             case (Some("number"), Some("float"))        => floatType()
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
@@ -281,6 +281,7 @@ object SwaggerUtil {
             case (Some("string"), Some("email"))        => stringType(None)
             case (Some("string"), Some("date"))         => dateType()
             case (Some("string"), Some("date-time"))    => dateTimeType()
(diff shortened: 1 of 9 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala
```diff
@@ -281,6 +281,7 @@ object SwaggerUtil {
             case (Some("string"), Some("email"))        => stringType(None)
             case (Some("string"), Some("date"))         => dateType()
             case (Some("string"), Some("date-time"))    => dateTimeType()
+            case (Some("string"), Some("byte"))         => stringType(None)
```
blast-hardcheese: https://swagger.io/docs/specification/data-models/data-types/#format seems to suggest this needs to be a base64-encoded string. There was https://github.com/OAI/OpenAPI-Specification/issues/50 as well, discussing the semantics.

Perhaps we could generate a `Base64` in order to explicitly control how the data is decoded?

hanny24: What class do you mean? AFAIK java's `Base64` class contains only encoder and decoders from `Array[Byte]` and `String`

blast-hardcheese: I'm suggesting similar to how [`Part`](https://github.com/twilio/guardrail/blob/1f61ff1ca69b01500710dbbbde5e448c6ee455ec/modules/codegen/src/test/scala/tests/generators/akkaHttp/server/FormFieldsTest.scala#L86-L90) works for form data parts, where we'd have a

```scala
class Base64String(val value: String) extends AnyVal
object Base64String {
  implicit val encode: Encoder[Base64String] =
    Encoder[String].contramap[Base64String](a =>
      new String(java.util.Base64.getEncoder().encode(a.value.getBytes()))
    )
  implicit val decode: Decoder[Base64String] =
    Decoder[String].map[Base64String](a =>
      new Base64String(new String(java.util.Base64.getDecoder().decode(a)))
    )
}
```
that would make it extremely clear what the expectation of the `String` is as it is passed through the encoding implicit infrastructure.

The motivation here is to explicitly state the set of expectations internally, if not even explicitly exposed to consumers of guardrail, to communicate that `type: string, format: byte` actually does enforce some potentially unexpected encoding semantics instead of being surprising. The amount of discussion around this particular feature seems to suggest that this is a source of confusion for many, and communicating the semantics via types seems to be useful here.

Open to discussion, just putting ideas forward.

hanny24: So I added `Base64String` inside `Implicits` object. Not sure if it's the best place for it.

hanny24:
I can see you did some changes. How can I help you to move this forward?

blast-hardcheese:
@hanny24 The nudge was appreciated, just rebased and force-pushed the reformatted branch -- let's see how far it gets.

tomasherman:
hey @blast-hardcheese, would it be possible to release guardrail with this code?

blast-hardcheese:
Can do, thanks for your patience!

tomasherman:
no worries

PR #543, opened 2020-02-10
Better parameter names handling in http4s server

hanny24:
This PR fixes a use case (http4s server) where parameter name was also scala keyword. This currently generates pattern matching code with backticks. Unfortunately, backticks are used for referencing existing variable inside pattern matching. So they can't be used for variables such as `type`. This PR aims for at least partial fix. Generator now tries to detect it and "fix" it by prefix parameter name with underscore. Not perfect, nor terrible. 

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
@@ -4,6 +4,7 @@ package generators
 import cats.arrow.FunctionK
 import cats.data.NonEmptyList
 import cats.implicits._
+import cats.Traverse
 import com.twilio.guardrail.core.Tracker
 import com.twilio.guardrail.extract.{ ServerRawResponse, TracingLabel }
 import com.twilio.guardrail.generators.syntax._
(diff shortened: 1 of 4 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Http4sServerGenerator.scala
```diff
@@ -713,6 +714,24 @@ object Http4sServerGenerator {
...
+    /**
+      * It's not possible to use backticks inside pattern matching as it has different semantics: backticks inside match
+      * are just references to an already existing bindings.
+      */
+    def prepareParameters[F[_]: Traverse](parameters: F[ScalaParameter[ScalaLanguage]]): Target[F[ScalaParameter[ScalaLanguage]]] =
+      Traverse[F].traverse(parameters) { param =>
+        if (param.paramName.syntax == param.paramName.value) {
+          Target.pure(param)
+        } else {
+          // let's try to prefix it with underscore and see if it helps
```
blast-hardcheese: This'll potentially generate conflicts as well. I'm going to rewrite this to prefix _all_ parameters if _any_ need to be prefixed.

blast-hardcheese: updated

PR #544, opened 2020-02-12
Upgrade http4s, cats, cats-effect and circe

jgogstad:
Upgrade http4s and dependencies:

http4s 0.21.0
cats 2.1.0
cats-effect 2.1.1
circe 0.13

* scalac had troubles finding the correct lub in the generated code for the http4s client. Solved with explicit type annotations. I don't know why this suddenly became an issue. Also, quasiquotes is not my strong suite, so please review and suggest edits.
* The changes in `CirceVersion` is just cargo culting, not sure if it's necessary

**Contributing to Twilio**

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -27,11 +27,11 @@ guardrail version | akka-http | akka-stream | circe  | cats  | Notes
 
 ### http4s
 
-guardrail version | http4s | circe-core | cats  | Notes
------------------ | ------ | ---------- | ----- | -----
-0.34.0 ⚠          | 0.18.0 | 0.9.0      | 1.0.1 | Only clients are supported
-0.39.0            | 0.18.0 |  "         |  "    | First server release
-0.41.1            | 0.19.0 | 0.10.0     | 1.4.0 |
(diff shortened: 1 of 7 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/circe/CirceVersion.scala
```diff
@@ -13,7 +13,7 @@ object CirceVersion {
     def encoderObjectCompanion = q"ObjectEncoder"
     def print                  = q"pretty"
   }
-  case object V012 extends CirceVersion {
+  case object V013 extends CirceVersion {
```
blast-hardcheese: This should be left at V012, otherwise we'll be bumping this needlessly every minor version. The intent is really to separate the new source incompatibility from the old variant, Travis has otherwise been very good about keeping source compatibility so I don't expect we'll need a tremendous number of these. If another is created, we'll have to figure out the naming scheme at that point.

Thank you for working through this idea though!

blast-hardcheese: Resolved this so it can be merged and released

Issue #251, opened 2019-04-25
Use InputStream instead of File for DW server file type


kelnos:
`File` requires that Jersey write out the data to a temp file on disk, while `InputStream` allows Jersey to stream the data to us.

I'm not sure how well it'll work client-side.  Users will have to create a `FileInputStream` to pass instead of a `File` (which isn't a big deal), but unclear how to make AsyncHttpClient work with it.  AHC has `InputStreamBodyGenerator` for turning an IS into a request body, but doesn't have something analogous for multipart file parts.

blast-hardcheese:
Something similar could be used in scala land as well. `File` was a stopgap to get _anything_ working. The helper method that generates `File`s from known parameters per handler route could be altered to produce `InputStream`s

PR #571, opened 2020-03-26
MIME type overhaul

kelnos:
From the most important commit message:

    This adds generic `TextContent` and `BinaryContent` classes that represent
    arbitrary content-types outside of our main directly-supported types
    (like application/json, text/plain, etc.).  The goal here is to
    transparently support arbitrary text and binary types.
    
    For example, imagine an API that handles document storage.  The API
    would likely make use of JSON for handling metadata about documents, but
    there could be endpoints that return the documents themselves.  It would
    be nice if we could support, say, application/pdf documents without
    having to special case every possible document type.
    
    The classification of text vs. binary when it comes to the application/
    registry is a bit fuzzy, since there's no encoding information attached
    to the content-type string itself.  There are a few things we can be
    sure about: if the type ends in "+json" or "+xml", then it's going to be
    text.  I've also hard-coded a few possibly-useful content-types that are
    known to be text.  Otherwise, things in application/ are assumed to be
    binary.  If there are other types that should be text, and people ask
    for them, adding support is a one-line addition.
    
    One wart: I've explicitly listed application/xml in
    `ContentType.unapply()` as unsupported (so `unapply()` will return `None`),
    because accepting it will break codegen for existing specs that have
    (for example) both application/json and application/xml listed as
    content types for a particular response.  At some point if we decide to
    actually support XML (whether as a first-class citizen, or just passing
    it through as a `String`, or using scala-xml's `NodeSeq`, etc.), we can add
    it back.

In addition, I've fixed up a bunch of issues with Java AHC client and DW server issues around content types and request/response bodies, and made it work properly with this new code.  The other frameworks should continue to output what they used to.

One interesting change is that was required to get this all to work: I've implemented https://github.com/twilio/guardrail/issues/251 finally (which supersedes https://github.com/twilio/guardrail/pull/318).  I decided to take the compromise position and use `InputStream` for request and response bodies that are solely binary data, but continue to use `File` on the server side for multipart/form-data.  This should give us much better usability and performance for the cases where it's safe to do so, but avoid unbounded memory buffering for binary data in the server multipart case.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
@@ -17,16 +17,14 @@ import com.twilio.guardrail.generators.Java.AsyncHttpClientHelpers._
 import com.twilio.guardrail.generators.{ ScalaParameter, ScalaParameters }
 import com.twilio.guardrail.generators.syntax.Java._
 import com.twilio.guardrail.languages.JavaLanguage
-import com.twilio.guardrail.protocol.terms.Response
+import com.twilio.guardrail.protocol.terms.{ ApplicationJson, ContentType, MultipartFormData, OctetStream, Response, TextPlain, UrlencodedFormData }
 import com.twilio.guardrail.protocol.terms.client._
 import com.twilio.guardrail.shims._
 import com.twilio.guardrail.terms.RouteMeta
-import com.twilio.guardrail.terms.RouteMeta.ContentType
 import com.twilio.guardrail.{ RenderedClientOperation, StaticDefns, SupportDefinition, SwaggerUtil, Target }
 import io.swagger.v3.oas.models.Operation
 import io.swagger.v3.oas.models.media.MediaType
 import java.net.URI
-
 import scala.collection.JavaConverters._
 
 object AsyncHttpClientClientGenerator {
@@ -47,17 +45,17 @@ object AsyncHttpClientClientGenerator {
   def getBestConsumes(operation: Operation, parameters: ScalaParameters[JavaLanguage]): Option[ContentType] =
     if (parameters.formParams.nonEmpty) {
       if (parameters.formParams.exists(_.isFile)) {
-        Option(RouteMeta.MultipartFormData)
+        Option(MultipartFormData)
       } else {
-        Option(RouteMeta.UrlencodedFormData)
+        Option(UrlencodedFormData)
       }
     } else if (parameters.bodyParams.isDefined) {
-      val validTypes = Seq(RouteMeta.ApplicationJson, RouteMeta.TextPlain)
+      val validTypes = Seq(ApplicationJson, TextPlain)
       operation.consumes
         .collectFirst({ case ContentType(value) if validTypes.contains(value) => value })
         .orElse({
           println(s"WARNING: no supported body param type for operation '${operation.getOperationId}'; falling back to application/json")
-          Option(RouteMeta.ApplicationJson)
+          Option(ApplicationJson)
         })
     } else {
       Option.empty
@@ -65,9 +63,9 @@ object AsyncHttpClientClientGenerator {
 
   def getBestProduces(operation: Operation, response: Response[JavaLanguage]): Option[ContentType] = {
     val priorityOrder = NonEmptyList.of(
-      RouteMeta.ApplicationJson,
-      RouteMeta.TextPlain,
-      RouteMeta.OctetStream
+      ApplicationJson,
+      TextPlain,
+      OctetStream
     )
 
     response.value
@@ -82,7 +80,7 @@ object AsyncHttpClientClientGenerator {
         priorityOrder
           .collectFirstSome(ct => allProduces.find(_ == ct))
           .orElse({
-            val fallback = if (valueType.isNamed("String")) RouteMeta.TextPlain else RouteMeta.ApplicationJson
+            val fallback = if (valueType.isNamed("String")) TextPlain else ApplicationJson
             println(s"WARNING: no supported body param type for operation '${operation.getOperationId}'; falling back to ${fallback.value}")
             Option(fallback)
           })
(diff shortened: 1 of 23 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/languages/JavaLanguage.scala
```diff
@@ -1,5 +1,14 @@
 package com.twilio.guardrail.languages
 
+import com.twilio.guardrail.languages.JavaLanguage.JavaTypeName
+
+object JavaLanguage {
+  case class JavaTypeName(tpe: com.github.javaparser.ast.`type`.Type) {
+    def asString: String          = toString
+    override def toString: String = tpe.asString
```
blast-hardcheese, line 8: IMO,

```suggestion
    def asString: String          = tpe.asString
    override def toString: String = asString
```

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
```diff
@@ -751,9 +691,9 @@ object AsyncHttpClientClientGenerator {
                                               println(
                                                 s"WARNING: no supported content type specified for ${operation.get.getOperationId}'s ${response.statusCode} response; falling back to application/json"
```
blast-hardcheese, line 692: Suggest including `operation.history` in the error message here

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
```diff
@@ -767,26 +707,40 @@ object AsyncHttpClientClientGenerator {
...
-                                                AssignExpr.Operator.ASSIGN
+                                            case contentType =>
+                                              val bodyGetter = valueType match {
+                                                case _ if valueType.isNamed("InputStream") => "getResponseBodyAsStream"
+                                                case _ if valueType.isNamed("byte[]")      => "getResponseBodyAsBytes"
+                                                case _ if valueType.isNamed("ByteBuffer")  => "getResponseBodyAsByteBuffer"
+                                                case _ if valueType.isNamed("String")      => "getResponseBody"
+                                                case _ =>
+                                                  println(
+                                                    s"WARNING: Don't know how to handle response of type ${valueType.asString} for content type $contentType for operation ${operation.get.getOperationId}; falling back to String"
```
blast-hardcheese, line 718: Suggest threading through `Tracker` for `valueType` (in whatever form that takes). If that's not practical, at least bounding with `operation.history` would be better for context.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardHelpers.scala
```diff
@@ -0,0 +1,58 @@
...
+            .orElse(contentTypes.collectFirst({ case bc: BinaryContent => bc }))
+            .getOrElse({
+              val fallback =
+                if (bodyParam.argType.isPrimitiveType || bodyParam.argType.isNamed("String")) TextPlain
+                else ApplicationJson
+              println(s"WARNING: no supported body param type for operation '$operationId'; falling back to $fallback")
+              fallback
+            })
+        })
+      )
```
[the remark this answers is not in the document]
kelnos: Yeah, I'm not really sure why I changed that in the first place.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
```diff
@@ -42,32 +52,35 @@ object AsyncHttpClientClientGenerator {
...
+      )
+      .orElse(
+        parameters.bodyParams.map({ bodyParam =>
+          val allConsumes = operation.consumes.flatMap(ContentType.unapply)
+          CONSUMES_PRIORITY
+            .collectFirstSome(ct => allConsumes.find(_ == ct))
+            .orElse(allConsumes.collectFirst({ case tc: TextContent => tc }))
+            .orElse(allConsumes.collectFirst({ case bc: BinaryContent => bc }))
+            .getOrElse({
+              val fallback = if (bodyParam.rawType.format.contains("string")) TextPlain else ApplicationJson
```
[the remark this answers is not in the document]
kelnos: Hmm I guess you looked at some intermediate state, as this code doesn't exist anymore.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardHelpers.scala
```diff
@@ -0,0 +1,58 @@
...
+        parameters.bodyParams.map({ bodyParam =>
+          CONSUMES_PRIORITY
+            .collectFirstSome(ct => contentTypes.find(_ == ct))
+            .orElse(contentTypes.collectFirst({ case tc: TextContent => tc }))
+            .orElse(contentTypes.collectFirst({ case bc: BinaryContent => bc }))
+            .getOrElse({
+              val fallback =
+                if (bodyParam.argType.isPrimitiveType || bodyParam.argType.isNamed("String")) TextPlain
+                else ApplicationJson
+              println(s"WARNING: no supported body param type for operation '$operationId'; falling back to $fallback")
```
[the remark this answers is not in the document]
kelnos: Good call

modules/codegen/src/main/scala/com/twilio/guardrail/languages/JavaLanguage.scala
kelnos: Yeah, that feels cleaner.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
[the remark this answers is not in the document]
blast-hardcheese: Yep, just reviewed each of the commits individually and spotted this. Calling it out for consistency.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/AsyncHttpClientClientGenerator.scala
kelnos: Hmm, looking at where `valueType` comes from, not sure that's practical (ultimately it's from the `Responses[LA]` param passed through the term case class.  But yeah, `operation.history` at least is a good idea.

PR #579, opened 2020-04-03
Add Java POJO support to form data

kelnos:
OpenAPI v3 allows you to use `$ref`s when specifying form data, which means you end up with a model object as your body parameter, instead of a parameter for each form field.  The latter has always been supported, but until now I lacked a decent way to use the model objects for
anything but JSON.

On the server side, I've implemented a MessageBodyReader and MessageBodyWriter which can be registered with Jersey and will handle ser/deser.

On the client side, I've inserted generated code for marshalling requests and unmarshalling responses to do the ser/deser.

Underlying all this is some (fortunately mostly common) code that uses Jackson's introspection capabilities to figure out the properties in the POJO and map to and from form data fields.

**Contributing to Twilio**


blast-hardcheese:
Closing this out due to merge conflicts. This does improve usability though

PR #585, opened 2020-04-06
Cleanup

blast-hardcheese:
Effectively a no-op. The only potentially questionable commit here is ffa5ce0, though if it's actually required a test will fail.

**Contributing to Twilio**


kelnos:
> if it's actually required a test will fail.

looks like a test failed 😂 ... not sure if it's due to the commit you referenced.  a butt-ton of:
```
[error] /home/runner/work/guardrail/guardrail/modules/sample-dropwizard/target/generated/polymorphismMapped/server/dropwizard/definitions/C.java:21:1: polytype has private access in polymorphismMapped.server.dropwizard.definitions.Base
[error] polytype
```
so like... something has made something that used to generate as (i assume) `protected` is now going `private`?  odd.

blast-hardcheese:
I'll bisect in a sec to figure out what's up

PR #591, opened 2020-04-09
Converting Free to Tagless final (pt1)

blast-hardcheese:
Converting `CoreTerm` completely, converting ClientTerm with a hacky Free adapter for now that will entirely go away once the rest of the refactors have been completed.

Merge conflicts should be fairly minimal, save for adding new parameters to `Free` methods.

Checklist
---

- [ ] `ProtocolSupportTerm`
- [ ] `ModelProtocolTerm`
- [ ] `EnumProtocolTerm`
- [ ] `ArrayProtocolTerm`
- [ ] `PolyProtocolTerm`
- [x] `ClientTerm`
- [ ] `ServerTerm`
- [ ] `FrameworkTerm`
- [ ] `SwaggerTerm`
- [x] `ScalaTerm`

**Contributing to Twilio**


blast-hardcheese:
@kelnos Gonna call this PR good at this point -- I wrote a scalafix rule that almost completely automated the migration of fc5e6d9, so it should be fairly straightforward to migrate everything else. I'll do the next few in Part 2.

blast-hardcheese:
One thing I don't like about this process is that there are a lot more conflicts generated for these one-line terms like the ones in ScalaTerms.

This can't be helped, and will be much less of an issue in larger interpreters, but scalafix seems to apply its own ideas of what good formatting looks like, only partially undone by scalafmt. I'll do what I can to keep the conflicts minimal, but the automated approach definitely does look like it will generate many more conflicts than I had hoped for.

The silver lining there is that with the scalafix rule, these migrations can be almost completely regenerated automatically in the case of some PR that should be merged first.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
```diff
@@ -17,7 +17,7 @@ import scala.reflect.ClassTag
 import scala.util.{ Failure, Success, Try }
 
 object Java {
-  implicit class RichType(val tpe: Type) extends AnyVal {
+  implicit class RichType(private val tpe: Type) extends AnyVal {
```
kelnos, line 20: Out of curiosity: why these changes?

modules/codegen/src/main/scala/com/twilio/guardrail/protocol/terms/client/ClientTerms.scala
```diff
@@ -10,31 +10,25 @@ import com.twilio.guardrail.terms.{ RouteMeta, SecurityScheme }
 import com.twilio.guardrail.{ RenderedClientOperation, StaticDefns, StrictProtocolElems, SupportDefinition }
 import java.net.URI
 
-class ClientTerms[L <: LA, F[_]](implicit I: InjectK[ClientTerm[L, ?], F]) {
+abstract class ClientTerms[L <: LA, F[_]] {
```
kelnos, line 13: Any particular reason for using abstract classes rather than traits?

modules/codegen/src/test/scala/support/SwaggerSpecRunner.scala
```diff
@@ -31,12 +32,13 @@ trait SwaggerSpecRunner extends EitherValues {
   ): (Context, FunctionK[CodegenApplication[L, ?], Target]) => (ProtocolDefinitions[L], Clients[L], Servers[L]) = {
     val parseOpts = new ParseOptions
     parseOpts.setResolve(true)
+    implicitly[ScalaTerms[L, Free[CodegenApplication[L, ?], ?]]]
```
kelnos, line 35: Debug code, or is this necessary?

modules/codegen/src/test/scala/support/SwaggerSpecRunner.scala
```diff
@@ -31,12 +32,13 @@ trait SwaggerSpecRunner extends EitherValues {
...
     runSwagger(new OpenAPIParser().readContents(spec, new java.util.LinkedList(), parseOpts).getOpenAPI, dtoPackage) _
   }
 
   def runSwagger[L <: LA](swagger: OpenAPI, dtoPackage: List[String] = List.empty)(context: Context, framework: FunctionK[CodegenApplication[L, ?], Target])(
       implicit Fw: FrameworkTerms[L, CodegenApplication[L, ?]],
-      Sc: ScalaTerms[L, CodegenApplication[L, ?]],
+      Sc: ScalaTerms[L, Free[CodegenApplication[L, ?], ?]],
```
kelnos, line 41: `Free` is being _added_ here?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
blast-hardcheese: #### `scalafmt: PreferCurlyFors` (1547e1d)
Rewrites `for (a <- fa; b <- fb) yield (a, b)` (bafflingly, the syntax scalafix prefers) into the long-form
```scala
for {
  a <- fa
  b <- fb
} yield (a, b)
```

#### `scalafix: RemoveUnused` (2fe53d9)
Stripped out some unused imports. I still think there are more, I'll try running this again a few more times.

#### `scalafix: ExplicitResultTypes` (86f2554)
There were a few public functions that didn't have an explicit result type, which this rule fixed.

#### `scalafix: LeakingImplicitClassVal` (31d1162)
This one was somewhat surprising:

```scala
@ implicit class RichString(val foo: String) { def bang = s"$foo!" }
defined class RichString

@ "blep".bang
res3: String = "blep!"

@ "blep".foo
res4: String = "blep"
```

Resolved if you don't have the `val`,
```scala
@ implicit class RichString(foo: String) { def bang = s"$foo!" }
defined class RichString

@ "blep".bang
res1: String = "blep!"

@ "blep".foo
cmd2.sc:1: value foo is not a member of String
val res2 = "blep".foo
                  ^
Compilation Failed
```

but more completely resolved by just marking the field private.

modules/codegen/src/main/scala/com/twilio/guardrail/protocol/terms/client/ClientTerms.scala
blast-hardcheese: I thought it would be fewer moving pieces to go from class to abstract class, but I guess it does prevent implementing multiple different of these interfaces via multiple mixin, even if we don't do it here, so I can change it.

modules/codegen/src/test/scala/support/SwaggerSpecRunner.scala
blast-hardcheese: so, this was the really exciting part of this PR. Take a look at the diff of the type of `ScalaTerms`:
```diff
-class ScalaTerms[L <: LA, F[_]](implicit I: InjectK[ScalaTerm[L, ?], F]) {
+abstract class ScalaTerms[L <: LA, G[_]] {
  // Changed F[_] to G[_] to disambiguate this example
```
... as well as the diff of the definition of one of the members...
```diff
-  def litString(value: String): Free[F, L#Term]                                 = Free.inject[ScalaTerm[L, ?], F](LitString(value))
+  def litString(value: String): G[L#Term]
```
... as well as the syntax injector, relying on `InjectK` to lift the terms of this interface into the target program:
```diff
-  implicit def scalaTerm[L <: LA, F[_]](implicit I: InjectK[ScalaTerm[L, ?], F]): ScalaTerms[L, F] = new ScalaTerms[L, F]
 // No bespoke implementation of functions needed, since ScalaTerms wasn't abstract before
+  implicit def scalaTerm[L <: LA, F[_]](implicit I: InjectK[ScalaTerm[L, ?], F]): ScalaTerms[L, Free[F, ?]] = new ScalaTerms[L, Free[F, ?]] {
+    def litString(value: String): Free[F, L#Term] = Free.inject[ScalaTerm[L, ?], F](LitString(value))
```

From the consumer's perspective, `implicit def scalaTerm` actually has the same structure, and can be used in the same way, because if we pass `Free[F, ?]` into `ScalaTerm`'s constructor in place of `G[_]`, it turns out all the types just line right back up again.

```scala
def litString(value: String): G[L#Term] // ... becomes...
def litString(value: String): Free[F, L#Term] // which was the exact same type as the first thing.
```

At the _call sites,_ however, anywhere we're still depending on `Free` to do our job still actually need to be explicit that they're using the `ScalaTerms[L, Free[CodegenApplication[L, ?], ?]]` implementation during this transitional period, only until there are implementations of all depended interfaces for the desired `F[_]` (`Target[_]` in this case) for any given function.

You can see that here: https://github.com/twilio/guardrail/blob/fc5e6d921b6a6ad3152a6a4821a9e0d1a8c69bce/modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala#L676-L678

Even though it doesn't explicitly mention whether it uses `Free[F, ?]` or `Target[_]` or whatever, all the implicit terms match (since there's only one), so _whatever_ `F[_]` gets subbed in will match, since we started providing both definitions.

In this way, the whole codebase will get converted over to this syntax without changing anything, really. The `Free` overrides we're doing on our internal guardrail extensions will continue to work until the very end, when only `Free` is deleted, but no other changes are made, since all the syntax changes have already been made gradually through these PRs. (That's another reason for prioritizing the scalafix rules to automate these changes)

PR #597, opened 2020-04-12
Tagless final pt3

blast-hardcheese:
- Bubbling `FunctionK` up into each interface, to reduce churn in the future
- Adding `copy(...)` methods to TF interpreters
- Keeping TF -> Free transitional adapters for the next release

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttp.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/AkkaHttp.scala
@@ -12,23 +12,15 @@ import CirceProtocolGenerator._
 import ScalaGenerator._
 import AkkaHttpGenerator._
 
-object AkkaHttp extends FunctionK[CodegenApplication[ScalaLanguage, ?], Target] {
-  val interpDefinitionPM: DefinitionPM[ScalaLanguage, ?] ~> Target       = ProtocolSupportTermInterp or new ModelProtocolTermInterp(CirceVersion.V012)
-  val interpDefinitionPME: DefinitionPME[ScalaLanguage, ?] ~> Target     = EnumProtocolTermInterp or interpDefinitionPM
-  val interpDefinitionPMEA: DefinitionPMEA[ScalaLanguage, ?] ~> Target   = ArrayProtocolTermInterp or interpDefinitionPME
-  val interpDefinitionPMEAP: DefinitionPMEAP[ScalaLanguage, ?] ~> Target = PolyProtocolTermInterp or interpDefinitionPMEA
-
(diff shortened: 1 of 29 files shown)


blast-hardcheese:
To build the `copy` methods, I used this in the REPL:
```scala
def buildCopy: Defn.Class => Unit = { case q"abstract class $ifaceName[..$_] { ..${stats} }" =>
  val (parms, defns) = stats.collect({ case Decl.Def(_, name, _, paramss, ret) =>
    val extract = "^(.)(.*)$".r
    val newName = name match { case Term.Name(extract(head, tail)) =>
      Term.Name(s"new${head.toUpperCase}${tail}")
    }
    val (argss, tpess) = paramss.map(_.map({ case Term.Param(_, name@Term.Name(_), Some(tpe), default) =>
      (name, tpe)
    }).unzip).unzip
    val parmdefault = name match {
      case Term.Name("MonadF") => name
      case _ => q"${name} _"
    }
    val parm = param"${newName}: ${tpess.foldRight[Type](ret)(Type.Function(_, _))} = ${parmdefault}"
    val defn = q"def ${name}(...${paramss}) = ${newName}(...${argss})"

    (parm, defn)
  }).unzip

  val copyDefn = q"""
  def copy(..${parms}) = new ${Init(ifaceName, ifaceName, Nil)}[L, F] {
    ..${defns}
  }
  """

  println(copyDefn.syntax)
}

val aclass = q"""
abstract class ArrayProtocolTerms[L <: LA, F[_]] extends FunctionK[ArrayProtocolTerm[L, ?], F] {
  def MonadF: Monad[F]
  def extractArrayType(arr: SwaggerUtil.ResolvedType[L], concreteTypes: List[PropMeta[L]]): F[L#Type]
}
"""

buildCopy(aclass)
```

where I swapped in each of the different interfaceas in for aclass and re-ran

blast-hardcheese:
@kelnos Yeah... sorry about that. I did my best to minimize the churn, but I hit diminishing returns. Even the last commit, `Finishing the migration by flattening all stub aliases`, is really gnawing at me. If I've got the energy tomorrow, I'll see if I can split it up and move those changes into their respective commits.

These three PRs have been effectively another big bang, any pending patch sets will likely need to be completely redone after this. Sorry to all blindsided by this, I don't anticipate any other major changes like this any time in the foreseeable future.

blast-hardcheese:
@kelnos OK, as promised I was able to golf the commits in this PR a bit, and the individual commits should be more comprehensible and are almost completely independent of each other (the `FunctionK` bubble-up and `copy()` commits are unable to be reordered past each other since they both append to the end of the interface)

PR #605, opened 2020-04-20
Finishing converting to Tagless Final

blast-hardcheese:
- [x] Hackily adapting the rest of the interpreters
- [x] Moving `FunctionK` implementation to the term interfaces
- [x] Removing all direct references to `FunctionK` in favor of `Framework`

**Contributing to Twilio**


blast-hardcheese:
@kelnos OK. Split this one out of the last PR, and painstakingly reverted the needless syntactic changes wrought by scalafix, the diff is much more tractable. I think this PR is significantly improved here, so thanks for pushing back.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
```diff
@@ -511,18 +510,5 @@ object EndpointsClientGenerator {
...
-        generateClientOperation(className, tracing, securitySchemes, parameters)(route, methodName, responses)
-      case GetImports(tracing)                                                      => getImports(tracing)
-      case GetExtraImports(tracing)                                                 => getExtraImports(tracing)
-      case ClientClsArgs(tracingName, serverUrls, tracing)                          => clientClsArgs(tracingName, serverUrls, tracing)
-      case GenerateResponseDefinitions(operationId, responses, protocolElems)       => generateResponseDefinitions(operationId, responses, protocolElems)
-      case GenerateSupportDefinitions(tracing, securitySchemes)                     => generateSupportDefinitions(tracing, securitySchemes)
-      case BuildStaticDefns(clientName, tracingName, serverUrls, ctorArgs, tracing) => buildStaticDefns(clientName, tracingName, serverUrls, ctorArgs, tracing)
-      case BuildClient(clientName, tracingName, serverUrls, basePath, ctorArgs, clientCalls, supportDefinitions, tracing) =>
-        buildClient(clientName, tracingName, serverUrls, basePath, ctorArgs, clientCalls, supportDefinitions, tracing)
-    }
```
kelnos, line 526: This is just delightful validation of the conversion.  The old interpreter just called out to methods that already existed, with the correct names and signatures, so the only change is to remove the old interpreter!

modules/codegen/src/main/scala/com/twilio/guardrail/terms/SwaggerTerms.scala
```diff
@@ -24,7 +25,7 @@ abstract class SwaggerLogAdapter[F[_]] {
   def error(message: String): F[Unit]
 }
 
-abstract class SwaggerTerms[L <: LA, F[_]] {
+abstract class SwaggerTerms[L <: LA, F[_]] extends FunctionK[SwaggerTerm[L, ?], F] {
```
kelnos, line 28: Do you plan to de-`FunctionK` this (and `FrameworkTerm` below) as well?

blast-hardcheese: Yep. Going to do one release with the hybrid TF/Free stuff, then kill FunctionK after that. Note, this PR _is still using `Free[F, ?]`,_ as the edges are still asking for Free, but now the thing that controls whether we're using Free or TF is exactly one function.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/EndpointsClientGenerator.scala
blast-hardcheese: Yeah, I was _really_ happy to see that 😄

blast-hardcheese:
Good call wrt `ScalaTerm`, I'll do that now. Also `ScalaParameter`.

kelnos:
Hah, didn't expect you to want to do the rename in this PR, but +1 :D

blast-hardcheese:
:tada:

Issue #476, opened 2019-12-03
Please make `using tags as package name` as a optional feature


UkonnRa:
## Background

Not all swagger authors use `tags` to mark the jvm packages, in fact, some authors just use `tags` as a ignorable mark.

## Expected Feature

Make `using tags as package name` as a optional feature, so that it will generate a large route instead of generating several packages

blast-hardcheese:
This is actually something I intended to remove long ago. It has already been deprecated for some time, and should just be deleted.

For the time being, you can use `x-jvm-package` to explicitly set the package, which will override the `tags`

The work required to remove this feature is to delete [this block](https://github.com/twilio/guardrail/blob/6bcb5a61b94299a66a781a0e8a4b5bcf2f927bec/modules/codegen/src/main/scala/com/twilio/guardrail/generators/SwaggerGenerator.scala#L122-L132), and then ensure that any tests that were using this feature are either updated to work correctly (if they used it incidentally) or just deleted outright.

Thank you!

blast-hardcheese:
For those interested in tackling this, the work required to remove this feature is to delete [this block](https://github.com/twilio/guardrail/blob/6bcb5a61b94299a66a781a0e8a4b5bcf2f927bec/modules/codegen/src/main/scala/com/twilio/guardrail/generators/SwaggerGenerator.scala#L122-L132), and then ensure that any tests that were using this feature are either updated to work correctly (if they used it incidentally) or just deleted outright.

UkonnRa:
Although I know `x-jvm-package` is a solution, but unfortunately, sometimes I have to use the swagger file written by others, who may not want to add such fields for some reasons... So please release the feature, THX a lot

blast-hardcheese:
This has been resolved in #1403, thanks!

PR #611, opened 2020-04-21
Removing tags shim

blast-hardcheese:
Resolving #476

Removing `tags` auto-mapping to package structure, as this has been deprecated for years.

**Contributing to Twilio**


kelnos:
The thing that I don't like about this deprecation or removal is that if you're consuming a third-party spec that you have no ability to edit or influence, then you just get non-namespaced `Client` or `Routes` or `Handler` classes, with no way to override that.

blast-hardcheese:
Hmm. We should make it opt-in then. I'll think a bit more about this.

PR #620, opened 2020-04-28
 Fix missing Produces annotation when no 2xx response with body

kelnos:
If an error response has a body type, even if the success response does not, we should still generate a Produces annotation for the method.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
@@ -6,7 +6,7 @@ import cats.implicits._
 import com.github.javaparser.StaticJavaParser
 import com.github.javaparser.ast.Modifier._
 import com.github.javaparser.ast.Modifier.Keyword._
-import com.github.javaparser.ast.{ Node, NodeList }
+import com.github.javaparser.ast.{ ImportDeclaration, Node, NodeList }
 import com.github.javaparser.ast.`type`.{ ClassOrInterfaceType, PrimitiveType, Type, VoidType }
 import com.github.javaparser.ast.body._
 import com.github.javaparser.ast.expr._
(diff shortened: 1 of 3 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
```diff
@@ -287,21 +287,33 @@ object DropwizardServerGenerator {
...
+
+              val allProduces = operation.downField("produces", _.produces).map(_.flatMap(ContentType.unapply)).unwrapTracker
+              NonEmptyList
+                .fromList(
+                  responses.value
+                    .flatMap(DropwizardHelpers.getBestProduces(operationId, allProduces, _))
+                    .distinct
+                    .map(_.toJaxRsAnnotationName)
+                )
+                .foreach(
```
blast-hardcheese, line 304: What does it mean to have more than one `@Produces`? Is that legit?

kelnos: Yeah, so I sorta mentioned this in DM on slack, but it is indeed allowed -- well, rather, you have one @Produces, but it can hold a `String[]` rather than just `String`.

Unfortunately, this doesn't currently work because the old swagger 2-to-3 `.produces` shim causes us to lose which content type is associated with which response.  That'll require some extra plumbing to fix, but I'll handle that in another PR.

modules/codegen/src/test/scala/tests/generators/dropwizard/server/DropwizardContentTypesTest.scala
```diff
@@ -0,0 +1,95 @@
...
+    ) = runSwaggerSpec(swagger)(Context.empty, Dropwizard)
+
+    genResource
+      .asClassOrInterfaceDeclaration()
+      .getMembers
+      .asScala
+      .collectFirst({
+        case method: MethodDeclaration if method.getNameAsString == "foo" => method
+      })
+      .get
```
blast-hardcheese, line 56: I think SwaggerSpecRunner brings in `OptionValues` from scalatest, so you can say `.value` and get a test failure instead of an exception.

blast-hardcheese: I'll actually take this in the PR I'm opening shortly.

Issue #231, opened 2019-04-09
Support for validation


tomasherman:
Hey,

are there any plans/interest in generating validation code for DTO objects? For example arrays that contain certain number of elements, strings with length constraints etc?

Thanks!

blast-hardcheese:
Yep, just hasn't been a priority for us as of yet. I've been thinking of translating validation to https://github.com/fthomas/refined, though I've heard some folk are already using that through `x-scala-type` and some `imports`.

I suspect what will need to happen to support different libraries which can be emailed to provide different functionality is to decompose new and constrained algebras to represent _just_ validation, for instance, then enable or disable the `refined` interpreter. If others want different behavior, they can swap in new interpreters that use different libraries, or the noop one that just passes values through.

tomasherman:
I'm thinking on about taking a stab at this...for our use case, we don't get to change the swagger definitions so we must handle it differently (we are unable to add `x-scala-type`)

I'm not really sure what you mean by this:

> I suspect what will need to happen to support different libraries which can be emailed to provide different functionality is to decompose new and constrained algebras to represent just validation, for instance, then enable or disable the refined interpreter. If others want different behavior, they can swap in new interpreters that use different libraries, or the noop one that just passes values through.

What i had in mind was to have two modes - `no validation` or `refined` validation which can be switched using cli arguments ... and for no validation, everything would work as is now but for `refined`, you would get types enriched with `refined` features

blast-hardcheese:
whoops, `s/emailed/enabled/`:

> I suspect what will need to happen to support different libraries which can be enabled to provide different functionality is to...
> decompose new and constrained algebras to represent just validation, for instance...
> then enable or disable the refined interpreter.

This is effectively what you suggested 😄

> If others want different behavior, they can swap in new interpreters that use different libraries, or the noop one that just passes values through.

The noop one would just be the current behaviour.

The big point is coming up with a new `Free` algebra that represents the set of validation features, transforming the generated types on their way through the pipeline.

Fortunately, there's almost no reflection once generated (apart from some `case t"Iterable[$tpe]"` stuff during parameter transform), so once you construct the types, they won't be messed with.

tomasherman:
Hey, it's me again ... so i spent a little time going through the code and i have some questions and insights.

[...]

PR #623, opened 2020-04-30
First naive pass at using Bean Validation in Jackson objects

andrewryno:
A naive implementation allowing for validation (#231) in Dropwizard projects by using Bean Validation on the Jackson protocol objects. Right now only implemented `@Min`/`@Max` to support `minimum`/`maximum`. 

I say naive as this is missing a few pieces for "full" support (apart from supporting all available constraints OpenAPI allows). Including testing, of which there is none.

The other key piece is determining how this validation actually works alongside Dropwizard. Dropwizard Validation will validate beans before the request ever makes it to the handler, meaning we'll be returning responses to the client that aren't necessarily documented or even follow a documented error response format.

Couple thoughts I've had:
- Create a new ExceptionMapper for `JerseyViolationMapper` which can map constraint violations to response objects generated by Guardrail -- would likely need to generate an abstract class with a method per operation. Developer would need to create an implementation of that class and register it using `environment.jersey().register()` (see: https://www.dropwizard.io/en/latest/manual/validation.html#extending)
- Attempt to bypass any validation by Dropwizard and amend the handler code to include built-in validation by calling Bean Validation directly (possibly through abstract class so a handler implementation can call `super` to use it or write their own)
- Something else I haven't though of

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
@@ -38,6 +38,19 @@ import com.twilio.guardrail.protocol.terms.protocol._
 import com.twilio.guardrail.terms.CollectionsLibTerms
 
 import scala.collection.JavaConverters._
+import com.github.javaparser.StaticJavaParser
+import com.github.javaparser.ast.{ Node, NodeList }
+import com.github.javaparser.ast.stmt._
+import com.github.javaparser.ast.Modifier.Keyword.{ FINAL, PRIVATE, PROTECTED, PUBLIC }
+import com.github.javaparser.ast.Modifier._
+import com.github.javaparser.ast.body._
+import com.github.javaparser.ast.expr._
+import java.math.BigInteger
+import java.util.Locale
+
+import com.twilio.guardrail.protocol.terms.protocol.PropertyConstraint.{ Maximum, Minimum }
+
+import scala.util.Try
 
 object JacksonGenerator {
   private val BUILDER_TYPE        = StaticJavaParser.parseClassOrInterfaceType("Builder")
@@ -51,7 +64,8 @@ object JacksonGenerator {
       parameterType: Type,
       rawType: RawParameterType,
       defaultValue: Option[Expression],
-      dataRedacted: RedactionBehaviour
+      dataRedacted: RedactionBehaviour,
+      constraints: Set[PropertyConstraint]
   )
 
   // returns a tuple of (requiredTerms, optionalTerms)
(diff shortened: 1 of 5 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -429,17 +434,22 @@ object JacksonGenerator {
...
 
         terms.foreach({
-          case ParameterTerm(propertyName, parameterName, fieldType, _, _, _, _) =>
+          case ParameterTerm(propertyName, parameterName, fieldType, _, _, _, _, constraints) =>
             val field: FieldDeclaration = dtoClass.addField(fieldType, parameterName, PRIVATE, FINAL)
             field.addSingleMemberAnnotation("JsonProperty", new StringLiteralExpr(propertyName))
+            for (constraint <- constraints) yield (constraint) match {
+              case Minimum(value) => field.addSingleMemberAnnotation("Min", new IntegerLiteralExpr(value.toString))
+              case Maximum(value) => field.addSingleMemberAnnotation("Max", new IntegerLiteralExpr(value.toString))
+              case _ => Unit
```
andrewryno: I don't _think_ this is needed?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -429,17 +434,22 @@ object JacksonGenerator {
           terms.filterNot(term => discriminatorNames.contains(term.propertyName))
 
         terms.foreach({
-          case ParameterTerm(propertyName, parameterName, fieldType, _, _, _, _) =>
+          case ParameterTerm(propertyName, parameterName, fieldType, _, _, _, _, constraints) =>
             val field: FieldDeclaration = dtoClass.addField(fieldType, parameterName, PRIVATE, FINAL)
             field.addSingleMemberAnnotation("JsonProperty", new StringLiteralExpr(propertyName))
+            for (constraint <- constraints) yield (constraint) match {
+              case Minimum(value) => field.addSingleMemberAnnotation("Min", new IntegerLiteralExpr(value.toString))
+              case Maximum(value) => field.addSingleMemberAnnotation("Max", new IntegerLiteralExpr(value.toString))
```
andrewryno, line 451: Thoughts on just using the fully qualified class name here instead of prematurely adding it to the imports?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
andrewryno: Ah, I think if I leave this out then if I add more `PropertyConstraint` classes the compiler will warn that it might not include every possible match?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
kelnos: I'm torn on this -- fully qualified is nice to prevent conflicts with names that end up getting generated from the spec file (like maybe there's a DTO called `Max` in `components/schemas`), but it also makes the generated code much harder to read.  I think the scala generators have been pushing in the direction of FQNs, but for java I haven't gone that way yet.  So for now I think this is fine.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
kelnos: As long as the parent trait of Minimum & Maximum is `sealed`, and those are the only two implementing classes, nope, not needed.  (And if it _was_ needed, you'd just have a blank case body there.)

kpavlov:
run pr validation

PR #625, opened 2020-05-02
Emitting OpenAPI parser warnings

blast-hardcheese:
Documenting initial investigatory work for integrating better error messages from OpenAPIParser

**Contributing to Twilio**


rtfpessoa:
As mentioned in https://github.com/twilio/guardrail/issues/741 would be really great if we could fail when some messages happen.
Not sure if there are messages which do not matter.

blast-hardcheese:
This is likely better than nothing, as you mention.

kelnos (approved):

kelnos:
It would be nice to be able to pass these out to the plugins so they can use their own logging facility, but that'd require a bit more work and for now it's good to just get this in.

Issue #315, opened 2019-05-29
Handing of required/nullable is in schemas not correct


kelnos:
After looking at [a SO post](https://stackoverflow.com/a/45577763/1626395), I've realized we're not quite handling `required` correctly.  The `nullable` property (which defaults to false) was added in OpenAPI v3 (there is no analog in Swagger v2).

(Note in the below tables, when I say `required: true` I mean "the property is included in the `required` array".)

## Serialization

| `required` | `nullable` | Output generation |
|:------------:|:------------:|:------------------------:|
| false | false | Property may be missing or present, but, if present, must not be `null` |
| false | true | Property may be missing or present, and, if present, may be `null` |
| true | false | Property must be present and must not be `null` |
| true | true | Property must be present but may be `null` |

Currently we do not do anything with `nullable`.  We treat `required: true` as `(required: true, nullable: false)`, and `required: false` as `(required: true, nullable: true)`.

It's not entirely clear what we should do for Swagger v2, which does not support `nullable`.  I would suggest that, for backward compatibility, we retain the current behavior there.  Since OpenAPI v3 support/use in Guardrail is still relatively new, we can correct the behavior without too much collateral damage.

## Deserialization

The spec is (I think) a little underspecified here; I think we should be as permissive as possible:

| `required` | `nullable` | Input parsing |
|:------------:|:------------:|:-----------------:|
| false | false | Allow `null` and treat it as not-present |
| false | true | Anything goes |
| true | false | Property must be present and non-`null` |
| true | true | Allow not-present and treat it as `null` |

I'm not 100% sure, but I _think_ we currently treat `required: true` as `(required: true, nullable: false)` and `required: false` as `(required: false, nullable: true)`.

## Other Stuff

The next question: how do we represent the tri-state of (absent, present-but-null, present) in our generated APIs?  (I have some thoughts here but will add as a comment later.)

As a somewhat-off-topic rant, I'm a little annoyed that OpenAPI didn't just follow JSON schema here and allow an array for `type:`, so you'd do `type: [ "string", "null" ]` instead of using `nullable: true`.

kelnos:
Only considering Java here.

Strawman value type:
```java
public abstract class PropertyValue<T> {
    public static class Absent<T> extends PropertyValue<T> {}

    public static class Null<T> extends PropertyValue<T> {}

    public static class Value<T> extends PropertyValue<T> {
        private final T value;

        private Value(final T value) {
            this.value = value;
        }

        public T getValue() {
[...]

PR #627, opened 2020-05-04
fix issue315 for scala

hanny24:
This is an initial take on #315 for scala generated backends. It turned out to be rather simple. Only things that might not be very obvious is `com.twilio.guardrail.protocol.terms.protocol.PropertyRequirement.Configured`. Essentially we want to be able to use different optional handling strategy for encoders and decoders. This will allow us to use legacy decoding on the input, while being strict on the output. Use cases of the users might be different thought, so that's why it is all configurable (default is to use legacy `PropertyRequirement` which means full backwards compatibility is maintained). 

Thanks in advance for your feedback!

**Contributing to Twilio**


modules/codegen/src/test/scala/tests/core/TypesTest.scala
```diff
@@ -320,4 +321,60 @@ class TypesTest extends AnyFunSuite with Matchers with SwaggerSpecRunner {
...
+      q"""
+        object TestObject {
+          implicit val encodeTestObject: Encoder.AsObject[TestObject] = {
+            val readOnlyKeys = Set[String]()
+            Encoder.AsObject.instance[TestObject](a => JsonObject.fromIterable(Vector(("required", a.required.asJson), ("required-nullable", a.requiredNullable.asJson), ("legacy", a.legacy.asJson)) ++ a.optional.map(value => ("optional", value.asJson)) ++ a.optionalNullable.fold(ifAbsent = None, ifPresent = value => Some("optional-nullable" -> value.asJson)))).mapJsonObject(_.filterKeys(key => !(readOnlyKeys contains key)))
+          }
+          implicit val decodeTestObject: Decoder[TestObject] = new Decoder[TestObject] { final def apply(c: HCursor): Decoder.Result[TestObject] = for (v0 <- c.downField("required").as[String]; v1 <- c.downField("required-nullable").as[Json].flatMap(_.as[Option[String]]); v2 <- c.downField("optional").as[Json].map(_.as[String].map(Some(_))).getOrElse(Right(None)); v3 <- c.downField("optional-nullable").as[Json].map(_.as[Option[String]].map(Property.Present(_))).getOrElse(Right(Property.Absent)); v4 <- c.downField("legacy").as[Option[String]]) yield TestObject(v0, v1, v2, v3, v4) }
```
blast-hardcheese, line 369: Question, should `required-nullable` be `Property[String]` instead of `Option[String]`? This seems asymmetric, compared to `optionalNullable: Property[Option[String]]`

hanny24: Not sure if I understood correctly, but...are you suggesting to add a third case to `Property`? So we have cases such as `Absent`, `Present` and `Value`?

This was actually my very initial implementation, but I ended up with this instead. Current implementaiton seemed to be easier to use, but now I'm having some second thoughts. Especially when I consider that `T` in `Property is actually always expected to be `Option`. So I would actually prefer to change this.

Let me know what you think.

blast-hardcheese: Well not even a third case, just `Property[String]` with the possible values of `Absent` and `Present(value)`. This is consistent with how `optional-nullable` represents `Present(Some(value))`/`Present(None)`/`Absent`.

If we think in terms of an imaginary world where `sealed trait Property[F[_], +A]` would be fun to work with,

| optional-nullable   | required-nullable |
| ------------------- | ----------------- |
| Property[Option, A] | Property[Id, A]   |

we get a consistent model.

hanny24: Ok, I think I understand.  You essentially propose to change `Present` case into `final case class Present[+T](value: Option[T]) extends Property[T]`. Is that correct?

blast-hardcheese: Sorry, must have miscommunicated -- I'm thinking along the lines of:
```diff
 implicit val decodeTestObject: Decoder[TestObject] = new Decoder[TestObject] {
   final def apply(c: HCursor): Decoder.Result[TestObject] = for {
     v0 <- c.downField("required").as[String]
-    v1 <- c.downField("required-nullable").as[Json].flatMap(_.as[Option[String]])
+    v1 <- c.downField("required-nullable").as[Json].flatMap(_.as[Option[String]].map(_.map(Property.Present(_)).getOrElse(Property.Absent)))
     v2 <- c.downField("optional").as[Json].map(_.as[String].map(Some(_))).getOrElse(Right(None))
     v3 <- c.downField("optional-nullable").as[Json].map(_.as[Option[String]].map(Property.Present(_))).getOrElse(Right(Property.Absent))
     v4 <- c.downField("legacy").as[Option[String]]
   } yield TestObject(v0, v1, v2, v3, v4)
 }
```

We're already using `Property[Option[T]]` for `optional-nullable`, so it's not too much to generate another case with this type.

blast-hardcheese: I'm realizing that I may be confusing the intent of `Property` to describe the `Nullability` of a value, which may not have been how you intended it to be used when writing this

hanny24: However it makes sense to use `Property` for `optional` case. It would make it somehow consistent. `Property` is used to describe presence/absent of the field in the json, while `Option` is used to `Nullability`.

modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala
```diff
@@ -169,14 +171,20 @@ object ProtocolGenerator {
...
     (for {
       isNullable <- schema.downField("nullable", _.getNullable)
-    } yield (isRequired, isNullable.getOrElse(false)) match {
-      case (true, false)  => PropertyRequirement.Required
-      case (true, true)   => PropertyRequirement.RequiredNullable
-      case (false, false) => PropertyRequirement.Optional
-      case (false, true)  => PropertyRequirement.OptionalNullable
+    } yield (isRequired, isNullable) match {
+      case (true, None)         => PropertyRequirement.Required
+      case (false, None)        => defaultPropertyRequirement
```
tomasherman, line 183: im nitpicking here but it might be good to order this somehow to make reading easier ... id suggest to sort by isRequired

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/CirceProtocolGenerator.scala
```diff
@@ -461,7 +522,28 @@ object CirceProtocolGenerator {
           q"implicit val guardrailEncodeLocalDateTime: Encoder[LocalDateTime] = Encoder[LocalDateTime]",
           q"implicit val guardrailEncodeLocalTime: Encoder[LocalTime] = Encoder[LocalTime]",
           q"implicit val guardrailEncodeOffsetDateTime: Encoder[OffsetDateTime] = Encoder[OffsetDateTime]",
-          q"implicit val guardrailEncodeZonedDateTime: Encoder[ZonedDateTime] = Encoder[ZonedDateTime]"
+          q"implicit val guardrailEncodeZonedDateTime: Encoder[ZonedDateTime] = Encoder[ZonedDateTime]",
+          q"""sealed trait Property[+T] {
+               def fold[R](ifAbsent: => R,
```
tomasherman, line 527: i think having to `toOption` would be a useful method here - i suspect his is what people will often want to do

modules/codegen/src/test/scala/tests/core/TypesTest.scala
hanny24: Changed!

tomasherman:
one unfortunate thing about this, if i understand this correctly is that for field that is nullable and optional, when you are returning data in server - is that you need to create it like this, correct?
```
case class User(name: Property[
User(

File: modules/codegen/src/test/scala/tests/core/TypesTest.scala (hunk 1 above)
Comment (tomasherman) [reply modules/codegen/src/test/scala/tests/core/TypesTest.scala]: i think name `Property` is a bit confusing and too generic ... could we name it something like `Nullable`? i think that is much more descriptive

**Comment by hanny24** [comment]
@tomasherman yes, you do.

File: modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala (hunk 2 above)
Comment (hanny24) [reply modules/codegen/src/main/scala/com/twilio/guardrail/ProtocolGenerator.scala]: reordered

File: modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/CirceProtocolGenerator.scala (hunk 3 above)
Comment (hanny24) [reply modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/CirceProtocolGenerator.scala:535]: Replaced `flatten` with `toOption`. Also added `Property.fromOption`.

File: modules/codegen/src/test/scala/tests/core/TypesTest.scala (hunk 1 above)
Comment (hanny24) [reply modules/codegen/src/test/scala/tests/core/TypesTest.scala]: Nullability is handled by `Option` data type. `Property` refers to a presence of a key in a json.

Comment (tomasherman) [reply modules/codegen/src/test/scala/tests/core/TypesTest.scala]: ah, i think i understand now ... i still think that `Property` name is too generic - for trait `Property` I would expect that the inheriting case classes would be things like `StringProperty, ObjectProperty, IntProperty etc...`

i would propose to rename it to something like `PropertyPresence`

Comment (blast-hardcheese) [reply modules/codegen/src/test/scala/tests/core/TypesTest.scala]: Could even go for `Presence.Present` vs `Presence.Absent`, in your other comment you indicated that this will end up leaking into user code, a good type name is important here.

Additionally, I think we establish this pattern in a few other places, but we will want to partially or fully quality the places we use this type. Additions like this into the common namespace of what can be generated in a specification reduce the set of specifications that don't need to be modified to work out of the box with guardrail.

Comment (hanny24) [reply modules/codegen/src/test/scala/tests/core/TypesTest.scala]: Renamed to `Presence` and it is always referenced using FQN from DTO classes (even though I belieive there might still be issues with name clashes)

**Comment by blast-hardcheese** [comment]
I aim to review this fully and explore the ergonomics this weekend. In the meantime, if there are any merge conflicts that come up, I will triage. Thank you for this, hanny 🎉

**Comment by hanny24** [comment]
No worries. Thanks!

**Comment by tomasherman** [comment]
hey @blast-hardcheese, did you get a chance to play around with this yet?

**Comment by blast-hardcheese** [comment]
@thomasherman Unfortunately not, this last weekend was our first Mother's Day, ended up not getting as much downtime as I expected. I've not forgotten about it, just trying to carve out some time for review. Thanks for your patience

**Comment by tomasherman** [comment]
no worries ;)

**Comment by blast-hardcheese** [comment]
OK, so I've got a few commits that I'll push to this branch in a bit. Just leaving some notes here to follow up on where I am with the review.

### Stuff I thought would improve the utility for users:
- Added `Presence.present[A](a: A)`/`Presence.absent[A]` to avoid invariance when using `Option.fold` from parameters
- Added `Presence.map`:

```scala
pa.fold(Presence.absent[Option[String]])(x => Presence.present(Some(x)))
```
became
```scala
Presence.fromOption(pa).map(Option(_))
```

(to permit easy translating of `formData` parameters into `Presence`-modeled `components`)

### Future work:
- `formData` + `x-empty-is-null` + `x-nullable` would permit `application/x-form-urlencoded` and `multipart/form-data` to have a useful implementation of `Presence` as well, currently this only works for circe and jackson generated components

### An issue I did find...
... seems to be `--encoder-optional-property optional --decoder-optional-property optional` generates uncompilable code, unfortunately. I'm going to flesh out the testing a bit to ensure that the future generated code doesn't violate any assumptions we're establishing in this PR.

### I'm also trying to...
... find some better names for `--[encoder|decoder]-optional-property [optional|requred-optional|legacy]` since doesn't (at least for me) immediately convey the expected behaviour, I find I have to read and reread in order to get it.

blast-hardcheese:
Settled on
```
--optional-[encode|decode]-as [legacy|optional|required-nullable]
```

If you've got time to review the changes I've made to your PR to ensure it still conforms to the expectations you have around behaviour, that would be very much appreciated. I still haven't had a chance to test the full runtime semantics across the 4x CLI options, but it should be possible to verify the semantics between clients and servers generated with the different options pretty easily.

If you get to the tests before me, great -- I'll announce here when I plan to start writing them though, so we don't duplicate effort.

blast-hardcheese:
OK, I'm giving a first shot at some tests. For now, I'm primarily interested in getting explicit clarity around semantics before merging and releasing. One other thing I noticed is after enabling `-deprecated`, I'm getting the warning about defining `Presence` inside the definitions package object. I'm going to move that out into a normal class definition.

tomasherman:
maybe a slightly unrelated question - have you considered creating a `guardrail-runtime` library? Seems like this `Presence` ADT would be a good candidate for it - having one type shared across all implementations might make it easier to write reusable use code ... right now there will be `Presence` generated for each client, right?

tomasherman:
Also, thanks for helping us with this - if there is something that we can help with let us know

hanny24:
It looks good from my point of view. Is there anything you need help with? In other words, is there anything else that you need to do/implement before we merge this?

blast-hardcheese:
>  `guardrail-runtime` library?
the ideal solution to this would be a `Common` configuration target, then a way to tell the other configurations to use it, then this becomes a configuration problem instead of trying to find which version of guardrail works with which version of the `runtime` library.

> is there anything else that you need to do/implement before we merge this?
I was looking at what would be involved with splitting `Presence` into a `support` package.

Something that irked me about the current strategy is that by putting this into `definitions`, it reduces the set of valid specs that can be run unmodified through guardrail.

If you have `Presence` in your `components`, even if you don't use the nullable stuff, you'll not be able to upgrade.

So, either this whole feature needs to be able to be turned off (I'd rather not introduce that complexity if we can help it) or we move the definition into a sibling package to `definitions`.

This may have other benefits in the future, as `support` would effectively be `guardrail-runtime`, as breaking it out may end up being easier.

Sorry for not bringing this up sooner, it really only occurred to me as I tried to start actually using `Presence`

hanny24:
So, I think I managed to move generated `Presence` into its own package (`support`).

blast-hardcheese:
I'll take a look ASAP! I did a once-over and this looks like exactly what I was hoping for. Making room for more utility infrastructure without stepping on users' toes, I really like it.

tomasherman:
hey, really sorry to nag again, i know you're busy man :) but any chance you were able to take a look? :)

blast-hardcheese:
I'm aiming to get this done today, thanks for your understanding Tomas :smile:

tomasherman:
no worries ;) i just had a sync with other team that is using this and promised to ask O:-)

blast-hardcheese:
Changes I've made:
- Decomposing/deduplicating the decoder generator into function composition
- Altering the generated `Decoder` to maintain the history in the case of decoding failure. Previously,
```scala
    import _root_.issues.issue315.optionaloptional.client.akkaHttp.Client
    import _root_.issues.issue315.optionaloptional.client.akkaHttp.definitions._
    import _root_.issues.issue315.optionaloptional.client.akkaHttp.support.Presence

/*  case class TestObject(
        required: String,
        requiredNullable: Option[String],
        optional: Presence[String],
        optionalNullable: Presence[Option[String]],
        legacy: Presence[String]
    ) */

    println(Json.obj(
        "required" -> Json.fromString("foo"),
        "required-nullable" -> Json.Null,
        "optional-nullable" -> Json.fromInt(5)
    ).as[TestObject])

// Exiting paste mode, now interpreting.

Left(DecodingFailure(String, List()))
```

Because `Int` was not compatible with `String`, we get a failure, but due to

```scala
c.downField("optional-nullable")
  .as[Json]
  .map(_.as[String]...)
```
in the Decoder, we lose the association between the `downField` and the error itself.

It's not super pretty, but this works:
```scala
((c: HCursor) =>
  c.value
    .asObject
    .filter(!_.contains("optional-nullable"))
    .fold(c.downField("optional-nullable").as[Option[String]].map(x => Presence.present(x))) { _ =>
      Right(Presence.absent)
    })(c)
```

tomasherman:
🎉

PR #629, opened 2020-05-06
Adding List and Vector to form encoder lifters, AddArg

### LINKED ISSUE #622: Missing AddArg.addArgVector?
Hi,
[v0.55.0](https://github.com/twilio/guardrail/blob/master/MIGRATING.md#0550-may-contain-breaking-changes-for-scala-clients-and-servers) changed the Scala type for OpenAPI's `type: array` from `IndexedSeq` to `Vector`, but I noticed [AddArg](https://github.com/twilio/guardrail/blob/master/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaGenerator.scala#L209) doesn't provide an implicit `addArgVector`: is this a bug or the intended behavior to have clients migrate to `Iterable`?

Note that query params defined as array of strings are generated as `Iterable[String]`, not `Vector[String]`!

Thank you so much for the amazing library!

blast-hardcheese:
Resolves #622 

Verified in the console:

```
scala> implicitly[AddArg[Option[Vector[Long]]]]
res0: issues.issue622.Implicits.AddArg[Option[Vector[Long]]] = issues.issue622.Implicits$AddArg$$anon$1@269e559f

scala> res0.addArg
   def addArg(key: String,v: Option[Vector[Long]]): String

scala> res0.addArg("foo", Some(Vector(1, 2, 3, 4)))
res1: String = foo=1&foo=2&foo=3&foo=4
```

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/Common.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/Common.scala
@@ -122,8 +122,14 @@ object Common {
     val CodegenDefinitions(clients, servers, supportDefinitions, frameworkImplicits)                     = codegen
     val frameworkImplicitName                                                                            = frameworkImplicits.map(_._1)
 
-    val dtoComponents: List[String]                         = definitions ++ dtoPackage
-    val filteredDtoComponents: Option[NonEmptyList[String]] = if (protocolElems.nonEmpty) NonEmptyList.fromList(dtoComponents) else None
+    val dtoComponents: List[String] = definitions ++ dtoPackage
+
+    // Only presume ...definitions._ import is available if we have
+    // protocolElems which are not just all type aliases.
+    val filteredDtoComponents: Option[NonEmptyList[String]] =
+      NonEmptyList
+        .fromList(dtoComponents)
+        .filter(_ => protocolElems.exists({ case _: RandomType[_] => false; case _ => true }))
 
     for {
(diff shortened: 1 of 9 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/Common.scala
```diff
@@ -122,8 +122,14 @@ object Common {
...
-    val dtoComponents: List[String]                         = definitions ++ dtoPackage
-    val filteredDtoComponents: Option[NonEmptyList[String]] = if (protocolElems.nonEmpty) NonEmptyList.fromList(dtoComponents) else None
+    val dtoComponents: List[String] = definitions ++ dtoPackage
+
+    // Only presume ...definitions._ import is available if we have
+    // protocolElems which are not just all type aliases.
+    val filteredDtoComponents: Option[NonEmptyList[String]] =
+      NonEmptyList
+        .fromList(dtoComponents)
+        .filter(_ => protocolElems.exists({ case _: RandomType[_] => false; case _ => true }))
```
kelnos, line 132: Are there any implications for Java here?

blast-hardcheese: This was explicitly to handle the java case 😅 `package-info.java` never seemed/seems to work for me, so this correctly does not emit the `definitons.*` import if the only `protocolElems` we have are just aliases

kelnos: Ah, neat.  We can get rid of `package-info.java` entirely then.

PR #655, opened 2020-06-04
Quoting issue

blast-hardcheese:
Turns out the casing normalization code attempts to interpolate groups. `$1`/`$2`... etc turn out not to be escaped, causing unexpected oddities.

The generated names are stable, if not odd, so the underlying protocol is not impacted, but the generated interface is unpleasant.

This also addresses an edge case where `[A-Z][^A-Z]`, _intended_ to catch `ABCFoo`, turning it into `abcFoo`, ends up with `ABCFOO$` into `AbcfoO$`, which makes sense, but is odd.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/package.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/package.scala
@@ -63,7 +63,7 @@ package object syntax {
   private val SPLIT_DELIMITERS = "[-_\\s\\.]+".r
   private val BOUNDARY_SPLITTERS = List(
     "([^A-Z])([A-Z])".r,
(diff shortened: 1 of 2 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/package.scala
```diff
@@ -62,14 +63,14 @@ package object syntax {
   private val SPLIT_DELIMITERS = "[-_\\s\\.]+".r
   private val BOUNDARY_SPLITTERS = List(
     "([^A-Z])([A-Z])".r,
-    "([A-Z]+)([A-Z][^A-Z]+)".r
+    "([A-Z]+)([A-Z][a-z]+)".r
```
kelnos, line 66: Well that's embarrassing. :-/

Issue #645, opened 2020-05-26
Allow for injecting Directive0 before complete in Akka HTTP?


DWvanGeest:
Hello again!

In some circumstances, it may be very useful if the Guardrail user can inject a `Directive0` into the generated Akka HTTP routes. The user could pass this directive as an argument to `FooResource.routes` with a default argument of `pass`, which does nothing. Then this directive can be included in the generated routes immediately prior to `complete`.

My use case is that I have a `Directive0` which checks that the authenticated user has the correct OAuth scope to access a certain path. I don't want to run this check, or give an `AuthorizationFailedRejection`, unless everything else (path, HTTP method, etc.) matches. 

Currently, I do wrap the generated route in my directive (similar to [this suggestion](https://github.com/twilio/guardrail/issues/90#issuecomment-413899877), which is OK-ish, but gives rise to confusing response codes. If my scope-protected route is `GET /hello` and someone tries `GET /bonjour` without the required scope, they'll get `403 Unauthorized` back. It's technically correct, but can lead the client to conclude that the `GET /bonjour` path actually exists. Essentially this is very similar to my issue #643 which gets at the precedence of HTTP response codes, and 404 having higher precedence than 403.

Another (theoretical) example of how this could be used is for metrics. You could inject a `Directive0` which increments a request counter for a specific path. I'm sure there are other ways to use such an escape hatch.

I'd be willing to contribute this. Do you think it would be a good idea?

kelnos:
For your specific use case, we should implement the OAS security goo (https://swagger.io/docs/specification/authentication/), which should support everything you need.  It specifically supports OAuth2 and specifying scopes.

I did some work last year to expose the security stuff in the generators.  Handling the semantics of the various combinations they can be expressed (it's honestly a little _too_ flexible for my taste) turns out to be kinda difficult; I did some work on a Dropwizard implementation, but never finished it.  It's possible that akka-http would be easier to do.

DWvanGeest:
Hey @kelnos, thanks for your response. 

[...]

PR #656, opened 2020-06-07
Add afterPathMethodMatch hook to generated Akka HTTP routes.

DWvanGeest:
This solves #645, by allowing the Guardrail user to inject behavior that is run immediately before the handler is called, via a `Directive0`. I realize @kelnos wasn't quite sure about this approach, but figured it would be good to have concrete code to discuss.

I followed the approach of #259, which perhaps is a somewhat similar idea on the Http4s side. Like that PR, this has broad applicability for doing auth, monitoring, response transformation, etc., but only after request preconditions specified in the OpenAPI spec are met (e.g. path, method, param matching).

@blast-hardcheese let me know what you think.

**Contributing to Twilio**


DWvanGeest:
Maybe one thing I should have taken from #259 is making `String => Directive0` the type of `beforeComplete`, so that the `operationId` is available. Am happy to add this if desired.

DWvanGeest:
Going to try a more general approach in a different PR.

PR #659, opened 2020-06-11
Escape invalid characters when used in Java identifiers

kelnos:
I ran into this when I wanted an enum value called `integer?`, and figured I'd add code for escaping all the various non-identifier characters, at least the ones found on a US keyboard (note that `$` is actually legal in Java identifiers, so it's not included here).

This turned out to be a huge rabbit hole, because, despite the work we did last year to make naming better, there was still a lot of inconsistency and hackery around naming things.  In order to make this a reasonable problem to solve, I added (to `LanguageTerms`) some new methods:

* `formatPackageName`
* `formatTypeName`
* `formatMethodName`
* `formatMethodArgName`
* (we already had) `formatEnumName`

The bonus here is that as we add other languages with different naming conventions, it becomes trivial to have guardrail follow those naming conventions.

In addition, I set things up so some names (for example, the name of a response class) are generated by the core (by calling the above methods) and threading them into the interpreter methods so the interpreters don't keep rebuilding names in an error-prone way.

This doesn't fix everything.  There are still some instances of `.capitalize` and `.uncapitalized` here and there.  There are *fewer* instances than before, which is nice.  And the core no longer does `.toCamelCase`, `.toSnakeCase`, or `.toPascalCase` anywhere.  The one thing I am still not able to eradicate is `ProtocolGenerator`'s `transformProperty`'s `needsCamelSnakeConversion`, because doing so still breaks the conflicting-name fixer-upper, and making that continue to work would require more surgery.  So, sadly, I will yet again punt on that.

This will unfortunately require some user code to change to continue to compile, which is detailed in MIGRATING.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/ClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/ClientGenerator.scala
@@ -34,28 +34,26 @@ object ClientGenerator {
   )(implicit C: ClientTerms[L, F], Fw: FrameworkTerms[L, F], Sc: LanguageTerms[L, F], Sw: SwaggerTerms[L, F]): F[Clients[L]] = {
     import C._
     import Sw._
+    import Sc._
     for {
       clientImports      <- getImports(context.tracing)
       clientExtraImports <- getExtraImports(context.tracing)
       supportDefinitions <- generateSupportDefinitions(context.tracing, securitySchemes)
       clients <- groupedRoutes.traverse({
         case (className, unsortedRoutes) =>
-          val routes     = unsortedRoutes.sortBy(r => (r.path.unwrapTracker, r.method))
-          val clientName = s"${className.lastOption.getOrElse("").capitalize}Client"
-          def splitOperationParts(operationId: String): (List[String], String) = {
-            val parts = operationId.split('.')
-            (parts.drop(1).toList, parts.last)
-          }
-
+          val routes = unsortedRoutes.sortBy(r => (r.path.unwrapTracker, r.method))
           for {
+            clientName <- formatTypeName(className.lastOption.getOrElse(""), Some("Client"))
             responseClientPair <- routes.traverse {
               case route @ RouteMeta(path, method, operation, securityRequirements) =>
                 for {
                   operationId         <- getOperationId(operation)
                   responses           <- Responses.getResponses[L, F](operationId, operation, protocolElems)
-                  responseDefinitions <- generateResponseDefinitions(operationId, responses, protocolElems)
+                  responseClsName     <- formatTypeName(operationId, Some("Response"))
+                  responseDefinitions <- generateResponseDefinitions(responseClsName, responses, protocolElems)
                   parameters          <- route.getParameters[L, F](protocolElems)
-                  clientOp            <- generateClientOperation(className, context.tracing, securitySchemes, parameters)(route, operationId, responses)
+                  methodName          <- formatMethodName(operationId)
+                  clientOp            <- generateClientOperation(className, responseClsName, context.tracing, securitySchemes, parameters)(route, methodName, responses)
                 } yield (responseDefinitions, clientOp)
             }
             (responseDefinitions, clientOperations) = responseClientPair.unzip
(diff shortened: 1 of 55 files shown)


kelnos:
Oh, also interesting, perhaps: `BacktickTest` is nearly obsolete, as almost all the names generated no longer need backticks.  Strangely `dashy-class` is still there as-is, which I should perhaps dig into and fix before we merge this.

kelnos:
Ok, two things:

1.  Fixed up protocol definition class names not getting formatted.  I'm not sure I did it in the best way, so I left it in its own commit so you can take a look at it separately.
2.  FINALLY I got rid of `needCamelSnakeConversion`.  Super happy this is gone, finally finally finally.

kelnos:
Oh -- now `BacktickTest` doesn't actually have any backticks in the generated code anymore, heh.  Should we remove it?

Also, apologies for the size of this PR... but I guess most of it is test changes/fixes.

modules/codegen/src/test/scala/core/issues/Issue126.scala
```diff
@@ -31,30 +31,30 @@ class Issue126 extends AnyFunSuite with Matchers with SwaggerSpecRunner {
 
     val handler  = q"""
       trait StoreHandler {
-        def getRoot(respond: StoreResource.getRootResponse.type)(): scala.concurrent.Future[StoreResource.getRootResponse]
+        def getRoot(respond: StoreResource.GetRootResponse.type)(): scala.concurrent.Future[StoreResource.GetRootResponse]
```
blast-hardcheese, line 34: _ah._ This is what you meant. Yes, this is quite unfortunate.

modules/sample-dropwizard/src/test/scala/core/Dropwizard/JavaInvalidCharacterEscapingTest.scala
```diff
@@ -0,0 +1,10 @@
+package core.Dropwizard
+
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers
+
+class JavaInvalidCharacterEscapingTest extends AnyFreeSpec with Matchers {
+  "Invalid characters should be escaped to reasonable things" in {
+
```
blast-hardcheese, line 8: the best kind of tests are the ones that can't be written, well done.

modules/codegen/src/main/scala/com/twilio/guardrail/ClientGenerator.scala
```diff
@@ -34,28 +34,26 @@ object ClientGenerator {
...
-          val routes     = unsortedRoutes.sortBy(r => (r.path.unwrapTracker, r.method))
-          val clientName = s"${className.lastOption.getOrElse("").capitalize}Client"
-          def splitOperationParts(operationId: String): (List[String], String) = {
-            val parts = operationId.split('.')
-            (parts.drop(1).toList, parts.last)
-          }
-
+          val routes = unsortedRoutes.sortBy(r => (r.path.unwrapTracker, r.method))
           for {
+            clientName <- formatTypeName(className.lastOption.getOrElse(""), Some("Client"))
```
blast-hardcheese, line 46: I think this is a great opportunity to push the string `"Client"` into each generator, further reducing the impedance mismatch when moving to support non-pascal-case languages.

modules/sample-dropwizard/src/test/scala/core/Dropwizard/JavaInvalidCharacterEscapingTest.scala
kelnos: Lol oops.  Too much multitasking.

modules/codegen/src/main/scala/com/twilio/guardrail/ClientGenerator.scala
kelnos: This is fine, actually, and i even could have made the param value "client" or "CLIENT" or whatever; the formatter is responsible for doing the right thing for the language.  I wanted the generators to _not_ be responsible for naming things, as much as is reasonably possible.

kelnos:
I'll try to take a stab at the scalafix rules, but might not get to it before you have time.

modules/codegen/src/main/scala/com/twilio/guardrail/ClientGenerator.scala
blast-hardcheese: makes sense, though hopefully my reticence towards hard-coded strings in core is understandable

kelnos:
I'm sure this will need a couple iterations to get right, but here is the scalafix rule I whipped up: https://github.com/blast-hardcheese/guardrail-scalafix-rules/pull/1

PR #689, opened 2020-06-17
Move Shower class to framework def instead of server/client def

kelnos:
When generating models only, the framework definitions of the server and client are not rendered.  However, the Shower class is still required for models, so move it into the framework support definitions list.

**Contributing to Twilio**


kelnos:
No changes are required.  All this does is remove the same something from both the client and server generator's support defs, and add it to the framework's support defs.  The framework support defs are always rendered, regardless of whether it's a client or server, so there's no user-visible change (except that generating just models will now work, which did not work before).

blast-hardcheese:
Was wondering about package membership, but if these aren't static classes then ok

kelnos:
Ah, nah, they're toplevels, and the package ends up being the same.

PR #700, opened 2020-06-25
Remove custom GuardrailJerseySupport

head-thrash:
Use bundled io.dropwizard.jersey classes instead of custom-written

+ fix tests

**Contributing to Twilio**


kelnos:
There was a specific reason why I wasn't using the built-in ones, but I can't remember offhand what that was.  Let's hold off on this until I can figure out what the reasoning was.

kelnos:
Figured it out: the built-in ones throw a `WebApplicationException` with a hard-coded JSON error payload that hides any information about what caused the error, so unless you happened to specify their exact error payload format in your spec file, your responses will be broken.  Even if you have a custom exception mapper that catches it, you can't extract any useful error information that would allow you to construct your own error payload.

head-thrash:
So, the only reason you'd overriden the default dw params is to output bad value in response?

Current behaviour -> it's gonna do the same, but without user input. 

```java
 /**
     * Given an input value from a client, creates a parameter wrapping its parsed value.
     *
     * @param input an input value from a client request
     */
    @SuppressWarnings({"AbstractMethodCallInConstructor", "OverriddenMethodCallDuringObjectConstruction"})
    protected AbstractParam(@Nullable String input, String parameterName) {
        this.parameterName = parameterName;
        try {
            this.value = parse(input);
        } catch (Exception e) {
            throw new WebApplicationException(error(input, e));
        }
    }

    /**
     * Given a string representation which was unable to be parsed and the exception thrown, produce
     * a {@link Response} to be sent to the client.
     *
     * By default, generates a {@code 400 Bad Request} with a plain text entity generated by
     * {@link #errorMessage(Exception)}.
     *
     * @param input the raw input value
     * @param e the exception thrown while parsing {@code input}
     * @return the {@link Response} to be sent to the client
     */
    protected Response error(@Nullable String input, Exception e) {
        LOGGER.debug("Invalid input received: {}", input);
        String errorMessage = errorMessage(e);
        if (errorMessage.contains("%s")) {
            errorMessage = String.format(errorMessage, parameterName);
        }
        return Response.status(getErrorStatus())
                       .entity(new ErrorMessage(getErrorStatus().getStatusCode(),
                               errorMessage))
                       .type(mediaType())
                       .build();
    }
```

idk... should guardrail respect it's framework choices? i mean, they'd decided to hide the user input after all. maybe some reason here? security, etc

kelnos:
The framework is _wrong_ in this case.  It will return an error payload that does not match what you have in your OpenAPI spec, and there is no way for the user to override that.

With the classes I wrote, the user can map the validation failures to the appropriate response payload in an exception mapper of their own devising.  It's not perfect, but it at least makes it _possible_ to do the right thing.

head-thrash:
>With the classes I wrote, the user can map the validation failures to the appropriate response payload in an exception mapper of their own devising. It's not perfect, but it at least makes it possible to do the right thing.

Did I get you correctly, by switching from WAE to BadRequestException you're removing WAE's specific payload that is provided by dw-jersey? that's the only change?

kelnos:
Pretty much, yes.  It's a shame that it doesn't seem like there's a simpler way to override this.

head-thrash:
yeah... sucks.

```java
...
                    if (throwable instanceof WebApplicationException) {
                        final WebApplicationException webApplicationException = (WebApplicationException) throwable;

                        // set mapped throwable
                        processingContext.routingContext().setMappedThrowable(throwable);

                        waeResponse = webApplicationException.getResponse();
                        if (waeResponse.hasEntity()) {
                            LOGGER.log(Level.FINE, LocalizationMessages
                                    .EXCEPTION_MAPPING_WAE_ENTITY(waeResponse.getStatus()), throwable);
                            return waeResponse;
                        }
                    }
...
```

basically you can't override WAE exception mapper if the WAE response has entity.

head-thrash:
pity jersey latest master has this unchanged too.

blast-hardcheese:
Is this an opportunity to clarify this with the upstream? Even if we can't delete some of our hacks currently, could we start a conversation surfacing this errant behaviour?

head-thrash:
not likely: 

https://github.com/eclipse-ee4j/jersey/issues/3716

looks like the owner of the origin change is not there. we may try to change it with PR and see what tests will it break, see if there any explanation in the test descriptions. or maybe get a reaction

blast-hardcheese:
Seems like it could help clarify things upstream. Do you have time and would you be willing to take this on?

head-thrash:
I'll check it out, try out the blunt PR, see the outcome. Will get back with results.

kelnos:
Sounds good.  Thanks for looking into this @head-thrash

PR #712, opened 2020-07-02
Separate out collections library operations and add support for Java Vavr

kelnos:
This pulls the things related to the collections library (optional, array, map) from `LanguageTerms`, and puts them in a new Terms class, `CollectionsLibTerms`.  It then threads through an implicit value of that class to all the protocol, client, and server generators.

I did a very very very minimal amount of work to convert a small part of the circe generator to use it, but there are still a lot of hard-coded uses of `Option`, `Vector`, and `Map`.  Given that there isn't really a popular alternative Scala collections library, I think this is ok for now.  If at some point there is one (or in the unlikely event there are drastic enough changes in Scala 3), we can go through and fix all that up.

I did a reasonably thorough job of converting the jackson, asynchttpclient, and dropwizard generators to use the new agnosticism.  There are still some warts:

1. There are places where we need to know if a particular `Type` is an optional or array type; previously it was done by looking for `java.util.Optional` and `java.util.List`, respectively.  Now it also looks for `io.vavr.control.Option` and `io.vavr.collection.Vector` as well.  This maybe should be factored out into a method on `CollectionsLibTerms`, but having the return value be wrapped in `Target` (`Target[Boolean]`, that is) will make it a complete bear to use (how do we feel about having some methods on a Terms class that has a return value _not_ wrapped in `Target`?).  A better solution might be to thread through that information from the core (if that's even possible), but that would probably require more surgery than I want to sign up for right now.
2. There are some places where generated code needs to manipulate optional values, specifically to do something side-effect-ful if the optional has a value in it.  The API on `java.util.Optional` and `io.vavr.control.Option` for doing this differs (`.ifPresent()`, and `.forEach()`, respectively), and this seemed a little too specialized an operation to put into `CollectionLibTerms` (`performOptionalSideEffect(Term, List[Statement])`? Gross).  So for now I'm doing something a little ugly and instead am using the `.map()` method (which exists on both classes), even though the body of the lambda doesn't transform anything.
3. AsyncHttpClient's Builder class takes in a HTTP client function and a jackson `ObjectMapper`, and it's stored internally as a `java.util.Optional`.  I decided to leave this alone because it complicates things to change it to be collections-lib-agnostic (wart 2 comes into play).  It's a minor wart because the use of `Optional` here does not at all leak into the builder's API.

Thoughts on how to resolve warts 1 and 2 are welcome.

Overall I'm reasonably happy with the result, though there is a lot of mechanical change to move things that are in "normal" code into the body of a `for` comprehension, so there's a lot of `val` removal in assignments, and a lot of `_ =` addition for side-effect-ful statements (which we have a lot of, because that's how javaparser works).  The generated code is still nice (aside from wart 2 above), and more places use fully qualified names, at least for collections classes.

I did not touch the Spring generators; they could stand to get the same treatment.

The Vavr `CollectionsLibTerms` impl appears to work and all the generated code compiles, at least.

I'm leaving this as a draft PR for now because I want to do some more testing, both actually adding tests for the Vavr stuff to make sure it actually works properly, and comparing old generated code with new generated code for the existing stuff to make sure there's nothing surprising.  Also need to add docs.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/ClientGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/ClientGenerator.scala
@@ -3,10 +3,10 @@ package com.twilio.guardrail
 import cats.data.NonEmptyList
 import cats.implicits._
 import com.twilio.guardrail.languages.LA
-import com.twilio.guardrail.protocol.terms.client.ClientTerms
 import com.twilio.guardrail.protocol.terms.Responses
+import com.twilio.guardrail.protocol.terms.client.ClientTerms
 import com.twilio.guardrail.terms.framework.FrameworkTerms
-import com.twilio.guardrail.terms.{ LanguageTerms, RouteMeta, SecurityScheme, SwaggerTerms }
+import com.twilio.guardrail.terms._
 import java.net.URI
 
 case class Clients[L <: LA](clients: List[Client[L]], supportDefinitions: List[SupportDefinition[L]])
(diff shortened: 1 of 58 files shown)


kelnos:
redid and split this in half, will open a new pr

PR #713, opened 2020-07-02
fix: fallback to string type for unsupported format types

ingarabr:
Falling back to string type that we know works rather than the raw format type specified in the specification. This will ensure
that the unsupported format type will compile, keeping the log message to guide the user on how it can be changed.

**Contributing to Twilio**

### DIFF
--- a/modules/sample/src/main/resources/raw-response.yaml
+++ b/modules/sample/src/main/resources/raw-response.yaml
@@ -24,3 +24,6 @@ definitions:
       id:
         type: integer
         format: int64
+      url:
+        type: string
+        format: url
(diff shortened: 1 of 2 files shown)


modules/sample/src/main/resources/raw-response.yaml
```diff
@@ -24,3 +24,6 @@ definitions:
       id:
         type: integer
         format: int64
+      url:
+        type: string
+        format: url
```
eli-jordan, line 29:
```suggestion
        format: uri
```

ingarabr:
@kelnos, sorry I don't follow how this change will not work in general. My proposal here is to fallback to a `String` and not the raw value provided in the specification. I find it better to generate code that compiles and guide the user to provide a more precise type.

I agree that specifying the preferred type is the best way but that's troublesome when you're not the owner/author of the specification. A way around this is to use type aliases but that again can cause name conflicts and other problems.

kelnos:
Argh, sorry, that's my mistake; I read the PR backward... I thought you were changing it *from* passing `None` *to* passing `fmt`, but it's actually the opposite.

ingarabr:
@kelnos, The failing test seemed to me to be unrelated to my changes. Anything else needed to be addressed or do you feel it's ready to be merged?

kelnos:
yeah, looks like codecov is being flaky.

Issue #616, opened 2020-04-25
possible to configure something like x-server-raw-request ?


willgu-gdm:
I see that `x-server-raw-response` is available to expose the underlying to users, allowing deeper customization of response.

I was wondering if it's possible to expose a similar interface for request as well? So that we can utilize some predefined functions by `akka-http`. For example, `extractClientIP`.

Not a huge deal and is blocking anything, since we can simply replicate all these functions, but would be a helpful feature to have :)

blast-hardcheese:
I guess this is already somewhat possible, as if you get a status code that's not in your spec, you just get the object as a `Left`.

That being said, for your use case, looking at [the docs] (https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/misc-directives/extractClientIP.html), it looks like this could just be exposed as a normal header parameter, it would just be a pain to expose all three.

There may be a valid way to write a `oneOf` to cover all three cases, though, which would be a good use case for exploring better support for `oneOf`.

blast-hardcheese:
Ah, just realized the motivation behind this, considering #572. Yeah, until that gets fixed the only workaround would be to only specify response codes that you won't actually receive, then handle the success responses manually in the `Left` that comes from the client call.

willgu-gdm:
thanks. Just in case it's not that clear, I was looking for `HttpRequest` to be added to the interface. For example, with that option `x-server-raw-request` enabled, instead of getting

```
def getControl(respond: ControlResource.getUserResponse.type)(userId: String): Future[ControlResource.getUserResponse]
```

the generated code looks like. 

```
def getControl(respond: ControlResource.getUserResponse.type)(userId: String, request: HttpRequest): Future[ControlResource.getUserResponse]
```

even though `userId` has been extract from this request, keeping the raw request gives us more room for customization.

ssboisen:
I could use this feature as well as it might allow me to circumvent the problem I'm facing in https://github.com/twilio/guardrail/issues/552#issuecomment-634460237

Issue #552, opened 2020-02-19
OAS 3.0 file upload generation fails


ollyw:
Trying to upgrade existing Guardrails project from 0.41.1 to 0.55.4 and Swagger 2 to OpenAPI 3, and having issues with file uploads. Guardrails is generating Akka Http endpoints.

When trying to attempt either recommended solution on the OAS website (https://swagger.io/docs/specification/describing-request-body/file-upload/) to posting a file upload, guardrails is generating errors. 

Attempt 1:
```
openapi: 3.0.2 
  ...
 /uploadendpoint:
    post:
      ...
      requestBody:
        required: true
        content:
          application/octet-stream:
            schema:
              type: string
              format: binary
```
This fails to generate with `Error:Unable to generate decoder for application/octet-streams`

Attempt 2:
```
openapi: 3.0.2 
  ...
  /uploadendpoint:
    post:
      ...
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                file:
                  type: string
                  format: binary
              required:
                - file
```
This fails to generate with `Unable to generate decoder for multipart/form-data`

It would be good not to have to post application/json and post a base64 encoded field (if that works, but not tested). The second attempt above is compatible with the old API and Swagger 2 (not the same way of declaring, but clients at runtime are compatible).

Perhaps I am doing something dumb, in which case I apologise in advance. Any help would be appreciated, and a confirmation what is possible would be great.

blast-hardcheese:
We tried to put a lot of care to ensure that Swagger 2.x specifications work as well as OAI3. Was there a possibility of just upgrading guardrail itself, checking to ensure there wasn't a regression?

Full file upload is an interesting edge case we have been struggling with across the different frameworks, due to some nuance involved with multiple file support for a single multipart form-data endpoint, as we either need to buffer the whole file in memory, write the file to the filesystem, limit the maximum number of upload fields to one, or accepting that we may need to explicitly make the parameters ordered. None of those answers are really good, which is why it's been particularly difficult to solve.

If you've got any insight into exactly what behavior you would expect would be much appreciated, especially considering you've used it before. Did it work as you expected? Notice any edge cases, etc?

ssboisen:
I'm also having trouble with using file upload with openapi 3.0.3 and akka-http, I'm getting the following error:

> Error:No decoders available (.paths./packageslollerscate.operations.POST.consumes)

I'm using the following configuration for the requestBody

```
      requestBody:
[...]

Issue #381, opened 2019-09-02
Question: Generalize tracing


eli-jordan:
The pattern used when `tracing=true` is defined, is just an instance of a more general pattern. Extracting some context data from the request, and passing it down to the handler.

Would you consider generalizing this, so that the user could pass in a function that generates the context.

e.g. in http4s it would look something like this

```scala

class MyResource[F[_], C](context: String => Request[F] => F[C] ...) {
   def routes(handler: MyHandler[F]): HttpRoutes[F] = HttpRoutes.of {
      case req @ POST -> Root / "test" =>  {
          val operationId = "myOperation"
          context(operationId).flatMap { ctx =>
             mapRoute(operationId, req, {
             req.decodeWith(myDecoder, strict = false) { body => 
                   handler.createEnvironment(MyResponse)(body)(ctx) flatMap {
                   ...
            }
          }
        })
      }    
   }
}
```

This generalised mechanism could be used for tracing, authentication, correlation ids etc

DWvanGeest:
I would love to see this too. Currently I am abusing the tracing mechanism to smuggle an auth directive into the generated Akka HTTP code.

Issue #298, opened 2019-05-21
Support for cookie parameter


kimxogus:
https://swagger.io/docs/specification/describing-parameters/#cookie-parameters

I generated scala `akka-http` routes with cookie parameter
```yaml
      - name: _session
        in: cookie
        description: bearer token
        required: false
        schema:
          type: string
```

But cookie parameter is not shown in generated code.

blast-hardcheese:
Hey @kimxogus, thanks for the report!

It's somewhat alarming that this doesn't work, as https://github.com/twilio/guardrail/blob/3066dfd6b456ed2c9964831407cc743a7965b8af/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaParameter.scala#L98-L99 would otherwise suggest that the functionality should be supported. Would you be available to dig in a bit and figure out why this isn't working as designed?

As a workaround, I'd expect
```
- name: Cookie
  in: header
  schema:
    type: string
```
and putting `s"_session=${cookieEscapedValue}"` as the header value would at the very least unblock you

kelnos:
@blast-hardcheese My guess is this doesn't work because we just don't do anything with the cookie params in the actual code generators.  https://github.com/twilio/guardrail/blob/master/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaParameter.scala#L18-L22 suggests that we don't even look at them, and I don't see any specific cookie handling in the codegen.

kimxogus:
@kelnos Yes, it seems that cookie params are not handled in https://github.com/twilio/guardrail/blob/master/modules/codegen/src/main/scala/com/twilio/guardrail/generators/ScalaParameter.scala#L18-L22 and all generators(akka, https in both java and scala)

TilakPatel:
Temporary fix: 

```
        - name: cookie
          in: header
          description: Cookie (generated from JS script)
          required: true
          schema:
            type: string
```

Apostrophe-seperated for multiple cookies. Ex. "x=y;test=key"

PR #747, opened 2020-08-09
Akka HTTP: Add option to extract arbitrary data from request and pass to handler

DWvanGeest:
This PR allows the user to extract arbitrary data from an Akka HTTP request, and have it passed to the route handler. It's inspired by the `tracing` code, and could theoretically replace it entirely since it's a more general version of the same thing.

There are a few examples of where this would be helpful already. I had a look for relevant open issues:

- @eli-jordan asking for this feature directly: https://github.com/twilio/guardrail/issues/381
- Me, wanting to inject a `Directive0`: https://github.com/twilio/guardrail/issues/645. I realized later that while a `Directive0` would be useful, I'm already extracting an Authorization header by means of a header parameter in the OpenAPI spec. This PR would solve both issues for me.
- @edmondo1984 wants to access Authorization data in the handler: https://github.com/twilio/guardrail/issues/117
- Providing a better place to inject the `oauth` directive discussed in https://github.com/twilio/guardrail/issues/90, so authorization can now be checked _after_ path matching. Wondering if this would address @Stinger911's comment about it being "enough".
- @willgu-gdm wants to extract the whole request, which would now be possible: https://github.com/twilio/guardrail/issues/616
- @ssboisen could extract the body: https://github.com/twilio/guardrail/issues/552#issuecomment-634460237
- @kimxogus would have a workaround for extracting cookies: https://github.com/twilio/guardrail/issues/298

Because of the composability of Akka Directives, you can really do a lot of different things with this! Some of those things may warrant a specific solution as well, but until those more specific solutions are developed, this may be a good escape hatch.

Notes:
- Am very new to Scala Meta, so let me know if I did things strangely
- I'm not very familiar with Http4s, so didn't want to go too far down the road of implementing it there. Plus, I think there's already some similar functionality with `mapRoute` there?
- I chose to make this feature entirely optional, like `tracing`. But, I could also see it being always-on with a default no-op `customExtract` (again similar to `mapRoute` which has a no-op default, looks like). Am happy to make that happen if desired.
- I did add one test, but can add more if needed.
- I didn't look into adding the new option to any of the build plugins yet, but can do if this seems like a good idea

Thanks for having a look!

**Contributing to Twilio**


modules/codegen/src/test/scala/tests/generators/akkaHttp/AkkaHttpServerTest.scala
```diff
@@ -397,4 +397,140 @@ class AkkaHttpServerTest extends AnyFunSuite with Matchers with SwaggerSpecRunne
...
+        def getRoot(respond: StoreResource.GetRootResponse.type)()(extracted: E): scala.concurrent.Future[StoreResource.GetRootResponse]
+        def putBar(respond: StoreResource.PutBarResponse.type)(bar: Long)(extracted: E): scala.concurrent.Future[HttpResponse]
+        def getFoo(respond: StoreResource.GetFooResponse.type)()(extracted: E): scala.concurrent.Future[StoreResource.GetFooResponse]
+        def getFooBar(respond: StoreResource.GetFooBarResponse.type)(bar: Long)(extracted: E): scala.concurrent.Future[StoreResource.GetFooBarResponse]
+        def getOrderById(respond: StoreResource.GetOrderByIdResponse.type)(orderId: Long, status: OrderStatus = OrderStatus.Placed)(extracted: E): scala.concurrent.Future[StoreResource.GetOrderByIdResponse]
+      }
+    """
+    val resource = q"""
+      object StoreResource {
+        def routes[E](handler: StoreHandler[E], customExtract: String => Directive1[E])(implicit mat: akka.stream.Materializer): Route = {
```
DWvanGeest, line 419: The TL;DR for this PR: enabled `customExtract` here, and `extracted` in the above handler trait.

DWvanGeest:
Thanks @kelnos! I can definitely write some docs. Seems like a new section on https://guardrail.dev/scala/akka-http/generating-a-server might be the best place for it? Probably a runtime test would help too.

Here's a short example of how it works with `customExtraction = true`, just extracting the whole Akka `HttpRequest`:

``` scala
// Guardrail-generated code
trait UserHandler[E] {
  def createUser(respond: UserResource.CreateUserResponse.type)(body: definitions.User)(extracted: E): scala.concurrent.Future[UserResource.CreateUserResponse]
}

object UserResource {
  def routes[E](handler: UserHandler[E], customExtract: String => Directive1[E])(implicit mat: akka.stream.Materializer): Route = // .... for details see the test
}

// user-written code
class UserApi extends UserHandler[HttpRequest] {
  override def createUser(respond: UserResource.CreateUserResponse.type)(body: definitions.User)(req: HttpRequest): scala.concurrent.Future[UserResource.CreateUserResponse] = {
    println(s"here's the whole Akka request: $req")
    ???
  }
}

val extractReq = (operationId: String) => extractRequest // Directive from Akka HTTP

val userRoutes = UserResource.routes(new UserApi, extractReq)

val binding = Http().bindAndHandle(userRoutes, "localhost", 1234).futureValue 
```

I didn't compile this, but you get the idea hopefully. And am realizing the handler should be contravariant, so that you could extract a subclass of what the handler expects. Will change that as well.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/AkkaHttpServerGenerator.scala
```diff
@@ -22,6 +22,11 @@ import _root_.io.swagger.v3.oas.models.Operation
 
 object AkkaHttpServerGenerator {
   object ServerTermInterp extends ServerTerms[ScalaLanguage, Target] {
+    val customExtractionTypeParamStr: String         = "E"
+    val customExtractionTypeParamName: Type.Name     = Type.Name(customExtractionTypeParamStr)
+    val customExtractionTypeParam: Type.Param        = Type.Param(List(), customExtractionTypeParamName, List(), Type.Bounds(None, None), List(), List())
```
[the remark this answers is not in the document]
DWvanGeest: Nice, I'll try that.

kelnos:
Oh wow, that's really cool.  Is there a way to extract multiple things?  Like say I wanted to provide a custom extractor that composes two akka-http directives, and ends up returning a `Directive2`?

Yes, `generating-a-server` is probably a good place for docs.

I'm not sure overall about the proposed mechanism for enabling/disabling this; would like @blast-hardcheese to weigh in on that.

DWvanGeest:
> Is there a way to extract multiple things?

You could do this by extracting a tuple. Akka already using this mechanism internally (there's no such thing as a `Directive2`, only `Directive1[Tuple2[A, B]]`. They do a bit of magic to make the extraction nicer, see https://doc.akka.io/docs/akka-http/current/routing-dsl/directives/index.html#automatic-tuple-extraction-flattening-.

We don't have any magic to make the handler flatten the tuple into separate params, but I think that's OK.

> I'm not sure overall about the proposed mechanism for enabling/disabling this

I'm not sure either. It's not a thing that everyone will use, certainly, and it does add a bit of noise to the generated handler trait if you're not using it. Hence the initial thought of making it optional via the build plugin. But, definitely open to other ideas.

DWvanGeest:
OK, I rebased on latest master, made the handler contravariant, cleaned up the type param declaration, and added some docs.

I looked at adding a runtime test, but it looks like that might require getting further into how this feature will be enabled, so held off for now.

anagrius:
Just FYI we are awaiting this with great anticipation :)
Anything we can do to get this moving? We need it for supporting POST of application/zip

blast-hardcheese:
@anagrius Hey, very sorry to keep you all waiting on this! It has been on my list, but I recognize that it has taken some time.

I've blocked off some time on my calendar to review the open question here; if I don't have an answer for you tomorrow, I'll at least provide an update around where I got.

Thank you for your contribution so far!

blast-hardcheese:
> I'm not very familiar with Http4s [...] I think there's already some similar functionality with mapRoute there?
This is actually a good call-out; `mapRoute` may warrant being added to akka-http as well, in the form of `(String, HttpRequest, Future[HttpResponse]) => Future[HttpResponse]`, though anything that could be done with that would be possible via `customExtraction`, so this is more of a call-out in case others recognize the need for their use case.

> I chose to make this feature entirely optional [but] I could also see it being always-on with a default no-op [...].
I used a strategy [sort of like this](https://github.com/twilio/guardrail/blob/1550480/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/AkkaHttpServerGenerator.scala#L442) to control when the user desired hashing file upload contents in-flight:

```scala
        val Unmarshalfile2Part = handler.uploadFileUnmarshalToFile[Option](None, handler.uploadFileMapFileField(_, _, _))...
        val Unmarshalfile3Part = handler.uploadFileUnmarshalToFile[Id]("SHA-256", handler.uploadFileMapFileField(_, _, _))...
```
which removed the possibility of losing the desired hashing strategy (`"SHA-256"`) along the way and ending up with a `None.get`. Under the hood this uses `Functor[Option]` and `Functor[Id]` to do the work, which could be employed here as well to have this sort of "enabled by default but stubbed out" behaviour, kind of like:

```scala
def addPet(respond: PetResource.AddPetResponse.type)(): Unit => scala.concurrent.Future[PetResource.AddPetResponse]
def addPet(respond: PetResource.AddPetResponse.type)(): RemoteAddress => scala.concurrent.Future[PetResource.AddPetResponse]
```
or something, but to the best of my knowledge there's no way to implement this without burdening the user with additional syntactic noise.

To that end, I think enabling it via simple CLI flag and options in the build tooling makes the most sense, at the expense of broadening the different varieties of generated `Handler` interfaces, increasing the cognitive burden when configuring the plugin.

modules/codegen/src/main/scala/com/twilio/guardrail/protocol/terms/server/ServerTerms.scala
```diff
@@ -63,7 +68,7 @@ abstract class ServerTerms[L <: LA, F[_]] {
           List[StrictProtocolElems[L]],
           Map[String, SecurityScheme[L]]
       ) => F[RenderedRoutes[L]] = generateRoutes _,
-      newGetExtraRouteParams: Boolean => F[List[L#MethodParameter]] = getExtraRouteParams _,
+      newGetExtraRouteParams: (Boolean, Boolean) => F[List[L#MethodParameter]] = getExtraRouteParams _,
```
blast-hardcheese, line 71: this highlights the danger of permitting primitive types in these `copy` helpers

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/AkkaHttpServerGenerator.scala
```diff
@@ -166,42 +193,58 @@ object AkkaHttpServerGenerator {
...
+            Type.Apply(Type.Name(handlerName), List(customExtractionTypeParamName))
+          } else {
+            Type.Name(handlerName)
+          }
+        val typeParams     = if (customExtraction) customExtractionTypeParams else List()
+        val routesParams   = List(param"handler: $handlerType") ++ extraRouteParams
         val routeImplicits = List(param"implicit mat: akka.stream.Materializer") ++ AkkaHttpHelper.protocolImplicits(modelGeneratorType)
+        val routesDef      = q"""def routes(..${routesParams})(..$routeImplicits): Route = {
+                             ..${combinedRouteTerms}
+                           }""".copy(tparams = typeParams)
```
blast-hardcheese, line 243: oh dang, https://github.com/scalameta/scalameta/issues/519 got merged! Gotta go back and clean up these `copy(tparams =...`

blast-hardcheese: Rebased this change up

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/AkkaHttpServerGenerator.scala
```diff
@@ -166,40 +194,57 @@ object AkkaHttpServerGenerator {
...
           }
         """
-    def getExtraRouteParams(tracing: Boolean) =
+      }
+    def getExtraRouteParams(customExtraction: Boolean, tracing: Boolean) =
       for {
         _ <- Target.log.debug(s"getExtraRouteParams(${tracing})")
-        res <- if (tracing) {
+        extractParam <- if (customExtraction) {
+          Target.pure(List(param"""customExtract: String => Directive1[$customExtractionTypeParamName]"""))
```
blast-hardcheese, line 214: There's one concern here -- If I supply `customExtract` with a `Directive1[Long] & Directive2[Boolean, String]` I'll end up with a `Directive3[Long, Boolean, String]`, which is not compatible with this type (expecting `Directive1`), effectively requiring `.tmap( (a,b,c) => Tuple1((a,b,c)) )` in order to align the types.

Attempting to reconcile this seems to require a little more thought, possibly `customExtract: String => Directive[E]`, which would force users to `new MyHandler[Tuple1[E]] { ... }`, but optimizes a bit more for composing `Directive`s.

What do you think?

blast-hardcheese (commented):
One question, but other than that this is excellent. The observation about `mapRoute` is a good one.

blast-hardcheese:
also rebased to resolve merge conflicts

blast-hardcheese:
also addressed some build warts from the recent Scala Dropwizard addition.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/AkkaHttpServerGenerator.scala
DWvanGeest: Hmm yeah very good point! Somehow I thought that your composition example would result in a `Directive1[(Long, Boolean, String)]`, but I just tried it, and you get back `Directive[(String, Int, Boolean)]`:

It does seem like `customExtract: String => Directive[E]` would be the way to go to enable that composition, which I think is important. It would be unexpected if `Directive1[Long] & Directive2[Boolean, String]` couldn't be passed in. The `new MyHandler[Tuple1[E]]` or is a little weird, but with some docs shouldn't be too bad?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/AkkaHttpServerGenerator.scala
DWvanGeest: Oh nice, I even found that PR while trying to figure this out, but didn't realize support had been merged :+1: .

DWvanGeest:
> To that end, I think enabling it via simple CLI flag and options in the build tooling makes the most sense, at the expense of broadening the different varieties of generated `Handler` interfaces, increasing the cognitive burden when configuring the plugin.

Agreed, thanks for plumbing that all the way through.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Scala/AkkaHttpServerGenerator.scala
blast-hardcheese: Yeah. Thinking about it a bit more, I really don't have a good way to work around this:
- **Option 1:** we could generate the unit special case, `new MyHandler_[Long]` where `abstract class MyHandler_[A] extends MyHandler[Tuple1[A]] { ... }`. **This is bad because** if someone tries to switch from `Directive1` to `Directive2`, they'll get an error message that drives them _away_ from simply removing the `_`.
- **Option 2:** we could provide a magnet "receiver" that lifts any `DirectiveN` to `Directive1[TupleN` via `.tmap(Tuple1(_))`, which would actually give us exactly the syntax we want, **at the expense of** magic that is more difficult to introspect, as well as introducing yet another thing that could go wrong.

blast-hardcheese: I think the easiest thing here is probably the best, just documenting the `Tuple1` requirement, as if a better solution presents itself in the future, it'll be backwards compatible with `Tuple1`, not forcing users to change anything in that case. If you need help with wording or examples, please feel free to reach out.

blast-hardcheese: Ah I see you already updated the documentation 😄 reviewing now

modules/microsite/docs/scala/akka-http/generating-a-server.md
```diff
@@ -57,20 +57,20 @@ trait UserHandler[-E] {
...
 
 For example, to extract an `X-User-Id` header value from an incoming request, your code might look like this:
 
 ```scala
-class UserApi extends UserHandler[String] {
-  override def createUser(respond: UserResource.CreateUserResponse.type)(body: definitions.User)(userIdHeader: String): scala.concurrent.Future[UserResource.CreateUserResponse] = {
-    println(s"The supplied X-User-Id header is: $userIdHeader")
+class UserApi extends UserHandler[Tuple1[String]] {
+  override def createUser(respond: UserResource.CreateUserResponse.type)(body: definitions.User)(extracted: Tuple1[String]): scala.concurrent.Future[UserResource.CreateUserResponse] = {
+    println(s"The supplied X-User-Id header is: ${extracted._1}")
```
blast-hardcheese, line 73: this is really unpleasant, thanks for pointing it out.

blast-hardcheese: So, despite my best efforts, the only thing I could come up with was:
```scala
object pkg {
  sealed class DirectiveAdapter[A](val value: Directive1[A])

  sealed trait DirectiveAdapterLowPriorityImplicits {
    implicit def fromDirective[A]: Directive[A] => DirectiveAdapter[A] = d => new DirectiveAdapter(d.tmap(Tuple1(_)))
  }
  object DirectiveAdapter extends DirectiveAdapterLowPriorityImplicits {
    implicit def fromDirective1[A]: Directive1[A] => DirectiveAdapter[A] = d => new DirectiveAdapter(d)
  }
  def routes[A](extract: DirectiveAdapter[A]): Directive1[A] = extract.value
}
```
which permits
```scala
pkg.routes[Unit](Directives.get)
pkg.routes[String](Directives.headerValueByName("foo"))
val foobar = (Directives.headerValueByName("foo") & Directives.headerValueByName("bar"))
pkg.routes[(String, String)](foobar)
```

*This is bad because* it does implicit conversion of values, and is broadly undiscoverable.

```scala
object pkg {
  import scala.language.higherKinds
  sealed trait DirectiveAdapter[F[_]] { def convert[A]: F[A] => Directive1[A] }
  sealed trait DirectiveAdapterLowPriorityImplicits {
    implicit object DirectiveAdapterN extends DirectiveAdapter[Directive] { def convert[A] = _.tmap(Tuple1(_)) }
  }
  object DirectiveAdapter extends DirectiveAdapterLowPriorityImplicits {
    implicit object DirectiveAdapter1 extends DirectiveAdapter[Directive1] { def convert[A] = identity _ }
  }
  def routes[F[_], A](extract: F[A])(implicit ev: DirectiveAdapter[F]): Directive1[A] = ev.convert(extract)
}
```
is _technically_ better, though `def routes[F[_], A]` is 👎 from a complexity perspective.

this is quite the puzzler.

#### To clarify

I'm now very much against just introducing `Tuple1` wholesale because propagating `._1` everywhere is really a terrible solution. The amplification of bad code is expressly against the core philosophy of why guardrail exists, we gotta defend this every chance we can get, since the internals of guardrail is much easier to change than the interfaces consumed by users.

DWvanGeest: Yeah as soon as I saw those tuple shenanigans I began to dislike that solution a bit more too.

Another thought: what if we actually had the user configure how many params to extract? If they want to extract 2 params, then we could generate signatures like:

```
def routes[E1, E2](handler: UserHandler, customExtract: String => Directive[(E1, E2)])

trait UserHandler[-E1, -E2] {
  def createUser(respond: UserResource.CreateUserResponse.type)(body: definitions.User)(extracted1: E1, extracted2: E2): scala.concurrent.Future[UserResource.CreateUserResponse]
  // ...
}
```

I didn't actually try this, but seems like it would work. I'm now headed into the woods for a week so can't work on it for a bit, but what do you think?

Alternately, maybe the original `Directive1` solution isn't too bad. I do think most people will be extracting a single thing, and we could document how to use `tmap` to compose multiple extractions.

blast-hardcheese: IMO the `[E1, E2]` strategy:
- requires configuring guardrail to explicitly align with individual use cases, precluding multiple invocations of `Handler` with different parameters in the same compilation unit
- ends up seizing all type parameters to `routes` for itself for complex services

To this end, the `[F[_], A]` approach is preferable, though the lack of symmetry in `routes[Directive1, String]` and `routes[Directive, (String, String)]` is not excellent.

While it's still not ideal, I think leaving `Directive1` and moving forward with `tmap(Tuple1(_))` for the time being, then adding this complexity down the line if there's significant motivation to do so.

blast-hardcheese: I'll revert back to your previous strategy, merge, and release.

DWvanGeest:
Sounds good, and thanks!

blast-hardcheese (approved):

DWvanGeest:
Tried this out today in our codebase, and it's working nicely! Thanks again for getting it merged.

blast-hardcheese:
Great to hear, and thank you for both your contribution and patience through the process! If you're comfortable being listed on our [Adopters](https://github.com/twilio/guardrail/blob/master/README.md#adopters) list, would you mind opening a PR for that?

DWvanGeest:
@blast-hardcheese I would very much like to, but will have to decline for now. If anything changes I'll open up a PR.

blast-hardcheese:
Not a problem, thank you again for the contribution 🙂

PR #752, opened 2020-08-13
Format generated source with scalafmt before writing to disk

kelnos:
Scalameta's code formatter generated unreadable code.  Scalafmt's (ironically, written by the same people) is much better.

I **do not** recommend merging this, as it makes the generation run take a **lot** longer.  But posting this here for posterity, or if there's a way to speed up the scalafmt run considerably, someone can take it.

**Contributing to Twilio**


blast-hardcheese:
Gonna close this, but I agree with the sentiment. Perhaps a better thing is to collect explicit examples of where scalameta emits unreadable code and contribute fixes upstream.

PR #756, opened 2020-08-20
Fix custom classname tests

kelnos:
This is why I shouldn't merge branches that aren't up to date with master...

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/test/scala/core/issues/Issue314.scala
+++ b/modules/codegen/src/test/scala/core/issues/Issue314.scala
@@ -118,7 +118,7 @@ class Issue314 extends AnyFunSpec with Matchers with SwaggerSpecRunner {
          |  sealed abstract class GetUserResponse(val statusCode: StatusCode)
          |  case object GetUserResponseOK extends GetUserResponse(StatusCodes.OK)
          |  object GetUserResponse {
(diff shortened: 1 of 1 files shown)


modules/codegen/src/test/scala/core/issues/Issue314.scala
```diff
@@ -118,7 +118,7 @@ class Issue314 extends AnyFunSpec with Matchers with SwaggerSpecRunner {
          |  sealed abstract class GetUserResponse(val statusCode: StatusCode)
          |  case object GetUserResponseOK extends GetUserResponse(StatusCodes.OK)
          |  object GetUserResponse {
-         |    implicit val getUserResponseTRM: ToResponseMarshaller[GetUserResponse] = Marshaller { implicit ec =>
+         |    implicit def getUserResponseTRM: ToResponseMarshaller[GetUserResponse] = Marshaller { implicit ec =>
```
blast-hardcheese, line 121: Why'd it need to change?

kelnos: For the jackson support in a previous PR this needed to be a `def` (the jackson stuff needs this to take an implicit param); I ended up just changing it for both rather than special-casing it, since def/val are source compatible for this purpose.

PR #768, opened 2020-09-04
Vavr collections lib for Java

kelnos:
Adds Vavr as a collections lib option for the dropwizard framework.  This changes usage in DW and AsyncHttpClient method arguments, response objects, and in Jackson DTOs, as well as using Vavr's `Future` instead of `CompletionStage`.

This does not modify the spring-mvc framework to be compatible with this.

**Contributing to Twilio**


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
```diff
@@ -334,6 +333,11 @@ object DropwizardServerGenerator {
                   // the @NotNull annotation *must* come before the @FormDataParam annotation.  See:
                   // https://github.com/eclipse-ee4j/jersey/issues/3632
                   parameter.getAnnotations.add(0, new MarkerAnnotationExpr("NotNull"))
+
+                  // Vavr's validation support for some reason requires this.
+                  if (param.param.getTypeAsString.startsWith("io.vavr.collection.")) {
```
blast-hardcheese, line 338: Does this impact folks that use `x-java-type: ...` to specify a shorthand vavr type?

modules/codegen/src/main/scala/com/twilio/guardrail/generators/JavaModule.scala
```diff
@@ -43,12 +38,17 @@ object JavaModule extends AbstractModule[JavaLanguage] {
...
-  def extract(modules: NonEmptyList[String]): Target[Framework[JavaLanguage, Target]] = {
-    implicit val collections = new JavaCollectionsGenerator.JavaCollectionsInterp with JavaStdLibCollections
+  def extract(modules: NonEmptyList[String]): Target[Framework[JavaLanguage, Target]] =
     (for {
-      (protocol, model, enum, array, poly) <- popModule("json", ("jackson", jackson))
-      client                               <- popModule("client", ("async-http-client", asyncHttpClient))
-      (server, framework)                  <- popModule("server", ("dropwizard", dropwizard), ("spring-mvc", spring))
+      collections <- popModule(
+        "collections",
+        Some(new JavaCollectionsGenerator.JavaCollectionsInterp with JavaStdLibCollections),
```
blast-hardcheese, line 45: The reason `popModule` hasn't included a "default" option up until now is to force people to choose in configuration, in order to keep the configurations from turning into a convoluted mess. `framework` exists for this, but if somebody opts in to `modules` they should be forced to specify all of them explicitly.

I'd be in favor of a `helpMessage: String` or `releaseNotes: String` though

modules/codegen/src/main/scala/com/twilio/guardrail/generators/collections/JavaVavrCollectionsGenerator.scala
```diff
@@ -0,0 +1,85 @@
...
+          Target.raiseUserError("FIXME: Got an Array of Maps, currently not supported")
+      }
+
+    override def liftMapType(value: Type, customTpe: Option[Type]): Target[Type] =
+      customTpe
+        .fold[Target[ClassOrInterfaceType]](safeParseClassOrInterfaceType("io.vavr.collection.Map").map(identity))({
+          case t: ClassOrInterfaceType =>
+            Target.pure(t)
+          case x =>
+            Target.raiseUserError(s"Unsure how to map $x")
```
blast-hardcheese, line 68: It'd be good to thread `Tracker` to this point (I suspect because interfaces with `SwaggerUtil` it'll be particularly unpleasant to do), but it would be a huge win for improving error messages.

modules/microsite/docs/scala/dropwizard/index.md
```diff
@@ -15,7 +15,7 @@ In addition to standard Dropwizard dependencies, you'll need:
 * `io.dropwizard:dropwizard-forms`
 * `com.datasift.dropwizard.scala:dropwizard-scala-core`
 * `com.fasterxml.jackson.datatype:jackson-datatype-jsr310`
-* `com.fasterxml.jackson.module:jackson-module-scala`
+* `com.fasterxml.jackson.module:jackson-module-scala_${scala_binary_version}`
```
blast-hardcheese, line 18: Should probably be [`scala.compat.version`](https://davidb.github.io/scala-maven-plugin/compile-mojo.html#scalaCompatVersion) as users that have to worry about this will likely be using scala-maven-plugin

blast-hardcheese (changes requested):
Looks good overall, one major consideration wrt `modules` though

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
kelnos: Yes and no.  If they override it to a `java.util` type, then this check will not match, and all will be well.  If they use a different `io.vavr` type, this check should still match, and all still should be well.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/JavaModule.scala
kelnos: The issue here is that this will break existing users. That is, they will have a working setup, then will upgrade guardrail to a new version, and then suddenly it will fail, saying they haven't specified all the modules they need to.

If you're ok with that, I'm fine trying to add a help message that makes it easier to figure out what's missing.

modules/microsite/docs/scala/dropwizard/index.md
kelnos: This is just for documentation purposes, but I guess it can't hurt.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/collections/JavaVavrCollectionsGenerator.scala
kelnos: Yeah, looking at the call trees, there are 12 places that call this, and most of them are deep through `SwaggerUtil` :-/

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
kelnos: To answer the question I maybe didn't realize you are maybe asking: if people are _already_ overriding to an `io.vavr` type, I suspect it does not work, because you get a runtime failure during validation without this extra annotation.

modules/codegen/src/main/scala/com/twilio/guardrail/generators/JavaModule.scala
kelnos: Fixed this up to remove the default after all, and will add a note to MIGRATING after I merge a different PR that will undoubtedly conflict.

kelnos:
Ok, so I removed the default-modules thing I made and set it up so it passes the list of possibilities in the `MissingModule` error object.  Threading `Tracker` through to more things would be nice, but is too much for this PR.  Gonna merge.

PR #777, opened 2020-09-11
fix Java 'equals' method

sullis:
The Java code generator was producing an 'equals' method that could not pass the
verification logic in the [equalsverifier] library.

I have added an EqualsVerifier unit test and adjusted the Java code generator
so that Guardrail's Java classes comply with [equalsverifier]

Reference:
https://github.com/jqno/equalsverifier

**Contributing to Twilio**


blast-hardcheese (approved):
Looks like it makes sense. In your investigation did you find why the previous implementation was not correct?

blast-hardcheese:
Ah -- interactively reviewed, makes more sense now.

PR #778, opened 2020-09-12
Scala deprecation updates

blast-hardcheese:
Addressing deprecation warnings for akka-http and http4s:
- akka-http: `Route.asyncHandler` -> `Route.toFunction`
- akka-http: `StatusCode.RequestEntityTooLarge` -> `PayloadTooLarge`
- akka-http: Remove `ActorMaterializer.apply()`
- http4s: `decodeString` -> `decodeText`

**Contributing to Twilio**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -92,7 +92,22 @@ lazy val root = modules.root.project
   .customDependsOn(microsite)
   .customDependsOn(cli)
   .aggregate(allDeps, microsite)
-  .aggregate(allModules: _*)
+  .aggregate(
+    cli,
+    core,
+    guardrail,
+
+    javaSupport,
+    javaAsyncHttp,
+    javaDropwizard,
+    javaSpringMvc,
+
+    scalaSupport,
+    scalaAkkaHttp,
+    scalaEndpoints,
+    scalaHttp4s,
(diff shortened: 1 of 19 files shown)


kelnos:
Will this narrow the version range of akka-http that we support?

blast-hardcheese:
Yeah:
- akka-http 10.0.11 is the first release that uses `PayloadTooLarge`
- akka-http 10.2.0 is the first release that exposes `Route.toFunction`
- akka 2.6.0 is the first release that uses the global `ActorMaterializer`
- http4s 0.21.5 is the first release with `decodeText`

I guess this PR really should just be a draft, but the http4s upgrade is actually nontrivial as it resolves "[the decoder] may loop infinitely for charsets other than UTF-8."

kelnos:
Circling back on this, I'm a bit worried about the akka 2.6 requirement, as I know there are things that are still using akka 2.5.

The `PayloadTooLarge` bit should be ok because only specs that use that status code should hit this, and those specs would already not work, right?

Maybe can split the http4s change into a separate PR that can get merged on its own?

blast-hardcheese:
These changes apparently have gradually made it in anyway, so closing this.

PR #783, opened 2020-09-17
ci: add JDK 15

sullis:


blast-hardcheese (approved):
Is there a reason to build on all the different JDK versions, or is it typical to just do a representative set?

sullis:
> Is there a reason to build on all the different JDK versions, or is it typical to just do a representative set?

Open source projects typically build with JDK 8, JDK 11, and the latest JDK version.    I'm keeping JDK 13 and JDK 14 here because I know of engineering teams who run Guardrail on these versions.

PR #793, opened 2020-09-27
Add @javax.annotations.Generated to Java generators

kpavlov:
**Contributing to Twilio**

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -36,6 +36,13 @@ There is one exception: if generating names in these formats causes a conflict, 
 
 If you find a case where a name is generated in a different way, please file an issue.
 
+Migrating to 0.62.0
+===================
+
+Generated code is annotated with `@javax.annotations.Generated` annotation to exclude generated code from test coverage analysis. 
+
+For JDK9+ it may be required to add `javax.annotation:javax.annotation-api:1.3.2` dependency to your project. For JDK 8 and below no new dependency is needed.
+
 Migrating to 0.55.0
 ===================
 
--- a/build.sbt
+++ b/build.sbt
@@ -29,6 +29,7 @@ val dropwizardScalaVersion = "1.3.7-1"
 val jerseyVersion          = "2.25.1"
 val kindProjectorVersion   = "0.10.3"
 val jaxbApiVersion         = "2.3.1"
+val javaxAnnotationVersion = "1.3.2"
 val springBootVersion      = "2.3.6.RELEASE"
 val jacksonVersion         = "2.11.3"
 val hibernateVersion       = "6.1.6.Final"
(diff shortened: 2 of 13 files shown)


kelnos:
Do IDEs/linters/other tools do anything useful with this?  I like the idea of doing this for completeness, but since it requires users to add yet another dependency, I'd rather not unless there's a clear user-facing benefit.

andrewryno:
Most code coverage analyzers will ignore anything with that annotation. SonarQube for example won't factor in any generated code for coverage metrics.

kelnos:
Ah, that's really useful then.

kelnos (changes requested):
Looks good, but please add a note to MIGRATING.md about the new required dependency.

kpavlov:
> Most code coverage analyzers will ignore anything with that annotation. SonarQube for example won't factor in any generated code for coverage metrics.
That was exactly the reason of this PR. Now we have to manually exclude Guardrail-generated code to jacoco excludes.

_javax.annotations:javax.annotation-api_ dependency is required for JDK9+ only.

BTW, the Guardrail's missing feature is a target JVM version flag or understanding "maven.compiler.source" property.
If the flag was be present then no dependency would be needed. Depending on the java version either `@javax.annotations.Generated` (<=JDK8) or `@javax.annotation.processing.Generated` annotations should be used.

kelnos:
> BTW, the Guardrail's missing feature is a target JVM version flag or understanding "maven.compiler.source" property.

It's possible to handle this using guardrail's module system, but I don't think it's worth it just for <=8, especially since it'll still work with 8 and below by adding the extra dependency.

kelnos:
I'm a little confused, though.  You're using `@javax.annotations.Generated`, but that will only work on JDK 8 and below?  And for JDK 9 and above you need to use `@javax.annotation.processing.Generated`?  If so, that's a blocker; we can't stop supporting 9 and above.

kpavlov:
It's not a blocker. For 9 and above customers have to add extra dependency as the annotation has moved. For 8 and below - no need for javax.annotations api dependency.

kpavlov:
> It's possible to handle this using guardrail's module system, but I don't think it's worth it just for <=8, especially since it'll still work with 8 and below by adding the extra dependency.

No, we need a dependency for 9+

head-thrash:
i guess the idea here is to use java8 annotation. _if_ you're building for jdk >8 you have to import a lib which has old 1.8 annotations, and yes, in new java they moved in to separate namespace

i think this is a stuff that we have to be diligent about since 
a) odd library from old jdk8 makes guardrail jdk8 aware/specific.
b) jdk8 lib may interfere with some existing code that may use jdk9+ namespace

kelnos (approved):
Apologies for not getting back to this.  Fortunately this is going to be a work-on-guardrail week for me.  Thank you for clarifying when the new dep is needed, that seems fine to me.  And a big thanks also for updating (and writing new) documentation!

kelnos:
Stupid merge conflicts...

PR #870, opened 2020-12-07
Add Coproduct support

markandrus:
This PR adds Coproduct support to Guardrail for Scala projects. What this means is that, if I have a Swagger/Open API schema which defines responses for some operation `foo` like the following:

```yaml
responses:
  responses:
    '200':
      schema:
        $ref: '#/definitions/Success'
    '400':
      schema:
        $ref: '#/definitions/Failure'
    '500':
      schema:
        $ref: '#/definitions/Failure'
```

Then Guardrail will generate a FooResponse like the following:

```scala
sealed abstract class FooResponse {
  type CoproductType = Success :+: Failure :+: CNil
  def toCoproduct: CoproductType
}
```

Then, any time I have a value `fooResponse` of type FooResponse, I can `select`, `map`, etc., the various response values:

```scala
val successO: Option[Success] = fooResponse.toCoproduct.select[Success]
val failureO: Option[Failure] = fooResponse.toCoproduct.select[Failure]

object handler extends Poly1 {
  implicit def caseSuccess = at[Success](success => log(s"Succeeded with $success"))
  implicit def caseFailure = at[Failure](failure => log(s"Failed with $failure"))
}

fooResponse.toCoproduct.map(handler)
```

**Pros**

- Guardrail responses are effectively sum types. Shapeless offers [Coproducts](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#coproducts-and-discriminated-unions), a powerful type for working generically with sum types. By providing a mapping from Guardrail responses to Coproducts, we enable engineers to leverage features in Shapeless while saving them the trouble of writing error-prone conversions themselves.
- Coproducts offer `map`, `flatMap`, etc., operations that ensure each case is covered. Unless you enable fatal warnings (not always feasible), this is something `match` won't give you. `fold` is better but, due to the way handlers are often written in-line, it's possible that, when updating Open API schemas, cases get mixed up.
- Coproducts offer a `select` function which projects/selects the Coproduct to an Option of the desired type. In other words, if a Coproduct contains a value `foo` of type `Foo`, you can `select[Foo]` to get `Some(foo)`; if the Coproduct does not contain a value of type `Foo`, you get `None`.
- The way Coproduct is used here allows distinct responses to be projected to a common value. For example, some HTTP applications return a uniform error response, irrespective of the actual HTTP status code. If your client doesn't care that it got a 400, 401, etc., just that it got an error response, using `select`, `map`, etc., with Coproduct reduces the number of cases you have to handle. Same for success cases.

**Cons**

- All Scala projects incur a dependency on Shapeless 2.x, even if they'll never use Coproduct. This could perhaps be solved with a configuration option to include/exclude Coproduct support, not unlike tracing.
- We could provide many of the same "Pros" listed above by implementing our own `select`, our own traits for consuming responses, without depending on Shapeless.
- The way Coproduct is used here is (intentionally) lossy. For example, if two Guardrail responses are empty, they'll both map to Unit. Now, when calling `map` on the Coproduct, we can no longer distinguish the two responses: we only have a single Unit case to deal with. This could be solved using [discriminated unions](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#coproducts-and-discriminated-unions), where elements of the Coproduct are labeled with their HTTP status code, the type of the response they were projected from, etc. I plan to explore this next.
- Coproduct performance may be slower (at both compile- time and run-time).

**Other Notes/Questions**

- In this PR, in order to access the Coproduct representation, you call a function `toCoproduct` on the Guardrail response. Should we instead provide an implicit conversion to `CoproductType`, such that, if the implicit is in scope, calling `toCoproduct` is not necessary?
- In this PR, Coproduct support is directly added to the Guardrail response, right below the `fold` function. Should we instead generate this in its own file? Would that be easier to include/exclude if we wanted to allow toggling Coproduct?

**Contributing to Twilio**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -33,6 +33,7 @@ val springBootVersion      = "2.3.4.RELEASE"
 val jacksonVersion         = "2.11.3"
 val hibernateVersion       = "6.1.6.Final"
 val javaxElVersion         = "3.0.0"
+val shapelessVersion       = "2.3.3"
 
 mainClass in assembly := Some("com.twilio.guardrail.CLI")
 assemblyMergeStrategy in assembly := {
@@ -344,7 +345,8 @@ val akkaProjectDependencies = Seq(
   "io.circe"          %% "circe-jawn"        % circeVersion,
   "io.circe"          %% "circe-parser"      % circeVersion,
   "org.scalatest"     %% "scalatest"         % scalatestVersion % Test,
-  "org.typelevel"     %% "cats-core"         % catsVersion
+  "org.typelevel"     %% "cats-core"         % catsVersion,
+  "com.chuusai"       %% "shapeless"         % shapelessVersion
 )
 
 val akkaJacksonProjectDependencies = Seq(
@@ -362,6 +364,7 @@ val akkaJacksonProjectDependencies = Seq(
   "org.glassfish"                  %  "javax.el"                % javaxElVersion,
   "org.typelevel"                  %% "cats-core"               % catsVersion,
   "org.scalatest"                  %% "scalatest"               % scalatestVersion % Test,
+  "com.chuusai"                    %% "shapeless"               % shapelessVersion
 )
 
 val http4sProjectDependencies = Seq(
@@ -374,7 +377,8 @@ val http4sProjectDependencies = Seq(
   "org.http4s"    %% "http4s-dsl"          % http4sVersion,
   "org.scalatest" %% "scalatest"           % scalatestVersion % Test,
   "org.typelevel" %% "cats-core"           % catsVersion,
-  "org.typelevel" %% "cats-effect"         % catsEffectVersion
+  "org.typelevel" %% "cats-effect"         % catsEffectVersion,
+  "com.chuusai"   %% "shapeless"           % shapelessVersion
 )
 
 val dropwizardProjectDependencies = Seq(
@@ -481,7 +485,8 @@ lazy val endpointsSample = (project in file("modules/sample-endpoints"))
       "org.julienrf"      %%% "endpoints-xhr-client-circe"    % endpointsVersion,
       "org.julienrf"      %%% "endpoints-xhr-client-faithful" % endpointsVersion,
       "org.scalatest"     %%% "scalatest"                     % scalatestVersion % Test,
-      "org.typelevel"     %%% "cats-core"                     % catsVersion
+      "org.typelevel"     %%% "cats-core"                     % catsVersion,
+      "com.chuusai"       %%% "shapeless"                     % shapelessVersion
     ),
     unmanagedSourceDirectories in Compile += baseDirectory.value / "target" / "generated",
     skip in publish := true,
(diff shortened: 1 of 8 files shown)


kelnos:
I love this!  It's also delightful how tiny the diff is!

> In this PR, in order to access the Coproduct representation, you call a function toCoproduct on the Guardrail response. Should we instead provide an implicit conversion to CoproductType, such that, if the implicit is in scope, calling toCoproduct is not necessary?

Probably not.  Implicit conversions are considered an anti-pattern these days and I think might even be slated for removal in Scala 3.

> All Scala projects incur a dependency on Shapeless 2.x, even if they'll never use Coproduct. This could perhaps be solved with a configuration option to include/exclude Coproduct support, not unlike tracing.

I'm a little worried about this, especially given:

> Coproduct performance may be slower (at both compile- time and run-time).

But:

> We could provide many of the same "Pros" listed above by implementing our own select, our own traits for consuming responses, without depending on Shapeless.

I'd rather we don't do this.  People who are interested in using our coproduct support are likely going to expect it to be shapeless, and rolling our own even if it has an identical API, would be weird.  Not to mention it's more generated stuff for us to maintain.

> n this PR, Coproduct support is directly added to the Guardrail response, right below the fold function. Should we instead generate this in its own file? Would that be easier to include/exclude if we wanted to allow toggling Coproduct?

I would go with whichever makes more readable generated code.  My instinct is that means putting it in the same file, but also I think if we put it in a separate file, that might make it easier for consumers to just _leave out_ the shapeless dependency at runtime (maven scope `provided`; not sure if there's an equivalent sbt concept) and still use the code as-is, without needing an enable/disable switch on guardrail itself.  IIRC the JVM won't try to resolve classes/imports until first use, so that might work?

> The way Coproduct is used here is (intentionally) lossy [...] This could be solved [...] I plan to explore this next.

I actually really like the simplicity of the lossiness here.  But I do imagine there are some APIs out there where the response code is significant but can't be divined from the contents of the entity body...

markandrus:
@kelnos 

> I love this! It's also delightful how tiny the diff is!

Thanks! Although I have an update adding support for discriminated unions which will increase the code size a bit (partially due to a bunch of scaladoc I've added&hellip;).

> Probably not. Implicit conversions are considered an anti-pattern these days and I think might even be slated for removal in Scala 3.

Thanks, good to know. I took a look at some other Scala 3 changes which informed some other things I was considering (e.g., symbol literal deprecation in Scala 3 informed my choice to use strings instead of symbols in extensible records — will share this experiment soon).

> I'd rather we don't do this. People who are interested in using our coproduct support are likely going to expect it to be shapeless, and rolling our own, even if it has an identical API, would be weird. Not to mention it's more generated stuff for us to maintain.

Agree.

> I would go with whichever makes more readable generated code. My instinct is that means putting it in the same file, but also I think if we put it in a separate file, that might make it easier for consumers to just leave out the shapeless dependency at runtime (maven scope `provided`; not sure if there's an equivalent sbt concept) and still use the code as-is, without needing an enable/disable switch on guardrail itself. IIRC the JVM won't try to resolve classes/imports until first use, so that might work?

Still need to explore this&hellip;

> I actually really like the simplicity of the lossiness here. But I do imagine there are some APIs out there where the response code is significant but can't be divined from the contents of the entity body...

I will share an update which adds discriminated union support (`toUnion`) and re-implements `toCoproduct` in terms of `toUnion`. This ideally brings "the best of both worlds". That experiment will also examine extensible records, which may identify one potential path forward on #77, although I'm not looking at re-implementing DTOs yet.

PR #876, opened 2020-12-10
More collections lib helpers

kelnos:
Discovered that I need a few more things...

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/collectionslib/CollectionsLibType.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/collectionslib/CollectionsLibType.scala
@@ -16,7 +16,7 @@ sealed trait CollectionsLibType {
   def isOptionalType(tpe: Type): Boolean
   def isArrayType(tpe: Type): Boolean
 
-  def completionStageToFutureType(completionStageExpr: Expression): Expression
+  def completionStageTermToFutureTerm(completionStageExpr: Expression): Expression
   def liftFutureType(tpe: Type): Type
   def futureMap(on: Expression, resultParamName: String, mapBody: List[Statement]): Expression
   def futureSideEffect(on: Expression, resultParamName: String, resultBody: List[Statement], errorParamName: String, errorBody: List[Statement]): Expression
@@ -58,7 +58,7 @@ trait JavaStdLibCollections extends CollectionsLibType {
   override def isOptionalType(tpe: Type): Boolean = CollectionsLibType.isContainerOfType(tpe, "java.util", "Optional")
   override def isArrayType(tpe: Type): Boolean    = CollectionsLibType.isContainerOfType(tpe, "java.util", "List")
 
-  def completionStageToFutureType(completionStageExpr: Expression): Expression = completionStageExpr
+  override def completionStageTermToFutureTerm(completionStageExpr: Expression): Expression = completionStageExpr
 
   override def liftFutureType(tpe: Type): Type =
     StaticJavaParser.parseClassOrInterfaceType("java.util.concurrent.CompletionStage").setTypeArguments(tpe)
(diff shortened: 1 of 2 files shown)


sullis (approved):
lgtm

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/collectionslib/CollectionsLibType.scala
```diff
@@ -13,13 +13,17 @@ import scala.compat.java8.OptionConverters._
...
   def isArrayType(tpe: Type): Boolean
 
-  def completionStageToFutureType(completionStageExpr: Expression): Expression
+  def completionStageTermToFutureTerm(completionStageExpr: Expression): Expression
+  def futureTermToCompletionStageTerm(futureExpr: Expression): Expression
   def liftFutureType(tpe: Type): Type
   def futureMap(on: Expression, resultParamName: String, mapBody: List[Statement]): Expression
   def futureSideEffect(on: Expression, resultParamName: String, resultBody: List[Statement], errorParamName: String, errorBody: List[Statement]): Expression
+
+  def asMonadicInstance(expr: Expression): Expression
```
blast-hardcheese, line 26: What does this mean?

blast-hardcheese: Are reasonable implementations of this `identity _` for things that are already monads? Does it rely on a particular shape, typeclasses, or what?

kelnos: Typeclasses!  If only Java had those.

It's there to get around an annoying bit with Java's stdlib collections: they chose to not actually put `map()`, `flatMap()`, etc. on them, and instead put them on `Stream`, which you can convert a normal collection to by calling `stream()` on it.  Vavr's collections classes are super similar to Scala's, and are fine as is (the implementation there is indeed`identity`).

I'm finding overall that the issue with using different collections libraries is that they just often don't behave the same, which I suppose is to be expected.  The simpler stuff in the main interpreter, like `literalString` or `liftOptionalTerm`, or `emptyOptional` is fine when we're talking about building DTO primitives, but when we try to actually build non-trivial logic, those quickly fall apart.  It's lucky that `map()`, `flatMap()`, etc. are pretty universal, but then you run into things like:

| Thing | Stdlib | Vavr |
|---------|------|------|
| Do a side effect | Optional.isPresent(Consumer\<T>) | Option.forEach(Consumer\<T>) |
| Provide a default value | Optional.orElse(T) | Option.getOrElse(T) |
| Provide a default value using a lambda | Optional.orElseGet(Supplier\<T>) | Option.getOrElse(Supplier\<T>) |
| Provide an alternate optional thing | (doesn't exist) | Option.orElse(Option\<T>) |

These things are just about naming, so they're not too bad.  Something I fortunately didn't need (yet?) is that the aforementioned collection -> `Stream` stuff in the stdlib also requires you to "collect" things at the end if you want a regular collection back (stdlib streams are otherwise lazy and won't do anything).  While with Vavr (or Scala) you just do `xs.map(x -> doIt(x))`, with the stdlib you have to do `xs.stream().map(x -> doIt(x)).collect(Collectors.toList())`.  The one place where I needed the `stream()` nonsense, I was calling `.toArray()` on the result, which fortunately is the same with both the stdlib (and is also a terminal thing that will collect) and in Vavr.

Ultimately I feel like this is the least-bad approach.  While it pulls some of the code that's getting generated out of the generators (reducing readability by adding indirection), it also means that the same generators can be used for either collections lib, without needing to copy/paste or come up with a clean way to parameterize or compose things (which ultimately I think might even be _less_ readable).

Anyway, guess I rambled on a bit farther than your question maybe warranted, but also just wanted to dump my thoughts :)

kelnos:
gonna take a new approach!

PR #877, opened 2020-12-10
add SBT command alias:  publishM2

sullis:
"Similar to publishLocal, publishM2 task will publish to the user’s Maven local repository."

https://www.scala-sbt.org/1.x/docs/Publishing.html


blast-hardcheese (approved):
Good call, thank you!

PR #887, opened 2020-12-15
New collections abstraction

kelnos:
Decided to go in a different direction with collections stuff.  Removed the `CollectionsLibType` stuff and replaced it with a new `CollecitonsAbstraction` which defines things like monadic operations on a collections library, as well as lifting AST nodes of particular types into this abstraction.

There are some warts to it... an interesting idea for this was that you could encode some type information in guardrail itself to make things a bit safer.  So you could do something like:

```scala
val five = new IntegerLiteralExpr("5").lift[Int]
val optFive = five.liftOptionalTerm[Option[Int]]
val plusOne = new LambdaExpr(new Parameter(new UnknownType, "number"), new BinaryExpr(new NameExpr("number"), new IntegerLiteralExpr("1"), BinaryExpr.Operator.PLUS)).lift[Int => Int]
val maybeFivePlusOne = optFive.map(plusOne)
val unliftedExpr: Expression = maybeFivePlusOne.value
```
... and then the generated code would be, for java stdlib:
```java
Optional.ofNullable(5).map(number -> number + 1)
```
... or for java vavr:
```java
Option.of(5).map(number -> number + 1)
```

However, it turns out that it's rare that we actually have types like `Int` lying around.  Usually it's a type that's going to be generated by guardrail, so we can't use it as a type parameter in guardrail itself (best we could do would be `AnyRef`), or it's some primitive type, but we don't have information as to what type it is anymore anyway (so now we just have to use `Any`).

There's definitely value to some of it: knowing that something is `Option[Any]` vs. `Vector[Any]` vs. even `Future[Vector[Any]]` can be useful to keep generation safer.

Another wart is that javaparser obviously doesn't expose type information for types _contained_ in expressions.  So I could do
```scala
new IntegerLiteralExpr("5").lift[Int => String]
```
... and the machinery would allow that.

So there's probably some room for improvement here, but I do think we're going to hit a ceiling pretty quickly on how type-safe we can be.  But I still think this is a nice start, and is a decent base to build off of.

(Thanks of course go to @blast-hardcheese for the idea and initial implementation of it; this is definitely something I wouldn't've come up with on my own.)

**Contributing to Twilio**


modules/codegen/src/test/scala/aaajunk/JunkTest.scala
```diff
@@ -0,0 +1,52 @@
...
+import com.github.javaparser.ast.NodeList
+import com.github.javaparser.ast.`type`.UnknownType
+import com.github.javaparser.ast.body.Parameter
+import com.github.javaparser.ast.expr._
+import com.twilio.guardrail.languages.JavaLanguage
+import com.twilio.guardrail.terms.collections._
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers
+
+class JunkTest extends AnyFreeSpec with Matchers {
```
blast-hardcheese, line 13: Rename this into an entry level test for this strategy and this is great, thank you!

blast-hardcheese (approved):
Fantastic

modules/codegen/src/test/scala/aaajunk/JunkTest.scala
kelnos: Made it more than entry level and even found a typo bug :-P

PR #890, opened 2020-12-16
Fix stdlib collections stream/collect

kelnos:
Properly handles the need to call `stream()` when starting "monadic operations" on `java.util.List`, and then `collect(Collectors.toList())` when finishing.  I do this by creating a `TermHolder` subclass that is used _after_ `stream()` has been called.  This way multiple monadic calls will not require a round trip between `Stream` and `List`.

Also adds `.toArray()` and `.filter()` to the collections abstraction.

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/terms/collections/JavaStdLibCollections.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/terms/collections/JavaStdLibCollections.scala
@@ -2,14 +2,15 @@ package com.twilio.guardrail.terms.collections
 
 import com.github.javaparser.StaticJavaParser
 import com.github.javaparser.ast.Modifier.finalModifier
-import com.github.javaparser.ast.NodeList
-import com.github.javaparser.ast.`type`.{ PrimitiveType, Type, UnknownType }
+import com.github.javaparser.ast.`type`.{ ArrayType, Type, UnknownType }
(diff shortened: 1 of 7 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/terms/collections/JavaStdLibCollections.scala
```diff
@@ -69,44 +87,93 @@ trait JavaStdLibCollections extends CollectionsAbstraction[JavaLanguage] {
         )
       )
 
+    override def empty[A]: TermHolder[JavaLanguage, MethodCallExpr, Vector[A]] =
```
blast-hardcheese, line 90: Might be useful to use `List` when representing `List`? `Vector` is closer to `Array` as far as Liskov substitution goes.

blast-hardcheese (approved):
Minor comment, but looks good other than that. The addition of ClassTag and reflection into the represented types is regrettable, but it seems mandated by the nature of javaparser.

modules/codegen/src/main/scala/com/twilio/guardrail/terms/collections/JavaStdLibCollections.scala
kelnos: Yeah that's fine.  I was mainly going with your `liftVectorTerm` etc. naming, but can change it.

PR #899, opened 2020-12-22
update Java reserved words list

sullis:
new reserved words: permits, record, sealed, yield

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
@@ -219,11 +219,14 @@ object Java {
     "new",
     "null",
     "package",
+    "permits",
     "private",
     "protected",
     "public",
+    "record",
     "requires",
     "return",
+    "sealed",
     "short",
     "static",
     "strictfp",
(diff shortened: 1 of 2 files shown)


modules/codegen/src/main/scala/com/twilio/guardrail/generators/syntax/Java.scala
```diff
@@ -219,11 +219,14 @@ object Java {
     "new",
     "null",
     "package",
+    "permits",
     "private",
     "protected",
     "public",
+    "record",
     "requires",
```
blast-hardcheese, line 227: Wild, Java got case classes? Cool!

kelnos:
Are these actually reserved, or just reserved for certain positions in the token stream?  I would suspect that they've maintained backward compatibility here and we don't need to escape these.

blast-hardcheese:
@kelnos I believe you're right. On JDK14,

```java
public class Hello {
    public record Foo(String record, String yield) {}

    public static void main() {
        String record = "Hello";
    }
}
```
in `Hello.java` compiles fine with `javac --enable-preview --release 14 -Xlint:preview /tmp/Hello.java`.

Closing this one out for now, but if I missed something please feel free to reopen.

PR #903, opened 2020-12-24
Formatting race condition redux

blast-hardcheese:
1. Deduplicated `WriteTree` sprawl in Scala and Java generators
2. Made `WriteTree` contain `Future[Target[Array[Byte]]]` so the render can safely fail
3. Incorporated that failure into the main runner instead of just throwing an exception

**Contributing to Twilio**

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/CLI.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/CLI.scala
@@ -170,7 +170,7 @@ trait CLICommon {
           rs =>
             ReadSwagger
               .readSwagger(rs)
-              .map(_.map(WriteTree.unsafeWriteTree))
+              .flatMap(_.traverse(WriteTree.writeTree))
               .leftFlatMap(
                 value =>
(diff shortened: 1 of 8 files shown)


blast-hardcheese:
@kelnos I've modified the `prettyPrintSource` in JavaGenerator to better take advantage of the `Future` in `WriteTree`. Presumably it was unpleasant to take advantage of `Future` before these changes, but when I attempted to move these two lines:
```scala
          val _         = source.getChildNodes.asScala.headOption.fold(source.addOrphanComment _)(_.setComment)(GENERATED_CODE_COMMENT)
          val sourceStr = source.toString
          Future {
```
into `Future`, I ended up getting some kind of `ConcurrentModificationException` or something from inside javaparser. I (possibly erroneously) assumed that by the time `prettyPrintSource` was called, ownership of `cu` was fully transferred over. Do you know anything about this?

If not, I'm fine to merge this and push forward, as there currently still is a race condition in the Java formatter (I re-ran the CI action associated with your recent commit to master, flipping the failed build to a passing one)

blast-hardcheese:
Merging as I keep running into the race condition in the formatter

PR #912, opened 2021-01-01
update release instructions

sullis:


blast-hardcheese (changes requested):
This is technically not correct, the process is to click "Release" on the draft maintained by `release-drafter`, that'll create the tag and kick off publish. As I made the last minute changes to your previous actions, I can clarify, but you can get a preview by looking at the state of the current draft release under "Releases"

(also I need to go back and tag the merged PRs with their semantic labels, as configured in `release-drafter`)

blast-hardcheese (approved):

PR #920, opened 2021-01-03
fix javaparser deprecation warnings

sullis:
https://www.javadoc.io/doc/com.github.javaparser/javaparser-core/latest/com/github/javaparser/ast/expr/IntegerLiteralExpr.html


blast-hardcheese (approved):
This was unexpected, why are they going _backwards_ in terms of statically representing values in their AST?

PR #950, opened 2021-01-30
[OAS3.0] Fixed responses with different content types for http4s

### LINKED ISSUE #949: [OAS3.0] http4s doesn't support a response with different content types
Http4s generator can't generate correct server for following specification:
```
openapi: 3.0.2
info:
  title: 'mixed content types'
  version: 1.0.0
paths:
  /foo:
    post:
      operationId: doFoo
      responses:
        200:
          description: success
          content:
            text/plain:
              schema:
                type: string
        400:
          description: "Invalid"
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Error'
components:
  schemas:
    Error:
      type: object
      properties:
        name:
          type: string
```
All responses are encoded as JSON, but `200` should be encoded as `String`

According the documentation, each response may have different content type but in the current implementation if there is a response with content type: `application/json` then all responses will be encoded as JSON.

LukaszMarchewka:
OpenAPI 3.x allows to define a content type per a response, but in Swagger 2.x you have to define `produces` for an operation. Swagger automatically converts operation without `produces` to content type `application/json`. In the existing code base a response without `produces` is converted to `Response` with content type `*/*`. So I have created new content type `AnyContentType` which is next converted to json in `Http4sHelper`. I'm not sure if it is the best solution, I have also had two other options.

this PR fixes #949 

**Contributing to Twilio**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -135,7 +135,8 @@ val exampleCases: List[ExampleCase] = List(
   ExampleCase(sampleResource("binary.yaml"), "binary").frameworks("java" -> Set("dropwizard", "dropwizard-vavr"), "scala" -> Set("http4s")),
   ExampleCase(sampleResource("conflicting-names.yaml"), "conflictingNames"),
   ExampleCase(sampleResource("base64.yaml"), "base64").frameworks("scala" -> scalaFrameworks.toSet),
-  ExampleCase(sampleResource("server1.yaml"), "customExtraction").args("--custom-extraction").frameworks("scala" -> Set("akka-http", "http4s"))
+  ExampleCase(sampleResource("server1.yaml"), "customExtraction").args("--custom-extraction").frameworks("scala" -> Set("akka-http", "http4s")),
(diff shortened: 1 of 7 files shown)


blast-hardcheese:
"I am not a child of my parent" seems unrelated to this change, re-running the tests

blast-hardcheese:
Not sure what's going on with GitHub checks, but I've confirmed that the suite passes.

Could you elaborate on the other options, just for posterity?

I don't have any objections to the strategy you employed here, _though_ it may be a useful motivator once and for all to de-thread the `com.twilio.guardrail.shims` package's `consumes` and `produces` once and for all, since they were initially only written to ease the migration to OAI3.

All 2-to-3 code should (ideally) be part of the swagger-parser library itself, to reduce the chance for this kind of deviation from the spec.

LukaszMarchewka:
I'm going to create an issue and explain how we can clean `ContentType`s.

Can you merge my PR in the meantime?

blast-hardcheese (approved):
Thank you for your contribution!

PR #1010, opened 2021-03-02
List -> NonEmptyList conversions

blast-hardcheese:
In situations where we are guaranteed to need lists with at least one element, use `NonEmptyList` to communicate this.

An alternate approach would be to use `::` explicitly, but I think NEL is less error-prone.


blast-hardcheese:
@kelnos This'll break the guardrail API, consumers will need to shuffle interfaces around to accommodate the new `NonEmptyList` methods. The advantage of this, otoh, is it makes the package hierarchy explicit, you can't have an empty package in this world.

kelnos (approved):

PR #1025, opened 2021-03-16
Resource extends from CirceInstances to allow for overriding Circe default behavior

rbrondgeest:
Proposal to add CirceInstances as a trait extension in generated resources, 

For being able to override the default behavior of Circe I would like to access CirceInstance members from its trait extension in in generated Resources.

This would allow for overriding f.e. default error handler behavior like so:

```
val apiResoure = new MyGeneratedApiResource[Task] with CirceInstances {
      override def jsonDecodeError: (Json, NonEmptyList[DecodingFailure]) => DecodeFailure = {
        (json, failures) => MyCustomMoreDescriptiveError(failures)
      }
}

```


blast-hardcheese:
So, as I commented [here](https://github.com/guardrail-dev/guardrail/issues/1024#issuecomment-800338990),

> This makes sense, though namespace collisions with user-controlled methods in handlers is not ideal.
> 
> This would better be served by configurations passed to routes, since that's the part of the code that actually does the work (think of the handler as purely business logic, almost a configuration in itself, that gets passed to the resource)

What's interesting though is if this really is all that's required, how does it work? Coincidentally via a `._` import that's already present?

rbrondgeest:
Yes, the package object `org.http4s.circe` extends CirceInstances, these methods become available trough `import org.http4s.circe._` 

So based on your comment; If we should not extend the Resource, then you would have to provide the custom default error handler as a configuration parameter, and let the generator add this override in the Resource? 

That way either you have to inject the entire error handler as a configuration parameter, or let the generated code be depending on user written code. Not sure if that is workable?

blast-hardcheese:
Ah, I actually misunderstood this initially, I thought you were adding the `CirceInstances` to `Handler`, not `Resource`. I understand the point you're making now. I'll review more carefully and get back to you shortly -- this is definitely a problem that needs more consideration, error messages are quite bad (and often break the OpenAPI contract) currently.

rbrondgeest:
> Ah, I actually misunderstood this initially, I thought you were adding the `CirceInstances` to `Handler`, not `Resource`. I understand the point you're making now. I'll review more carefully and get back to you shortly -- this is definitely a problem that needs more consideration, error messages are quite bad (and often break the OpenAPI contract) currently.

Did you get any time to reconsider this?

blast-hardcheese:
> Did you get any time to reconsider this?

I didn't get a chance last week, but I'm taking a closer look now.

This actually makes a lot of sense, especially since the `Resource` class already extends `Http4sDsl`.

What I'm testing now is _removing_ the `org.http4s.circe._` import, to ensure that we have a SPOT for those implicits (preventing accidentally falling back to the defaults).

That seems to have worked, so pushed up a patch.

blast-hardcheese:
@rbrondgeest Hey, so I believe this is the correct direction. Do you anticipate any problems with the patch I pushed? If not, I think we can merge and release this.

Please pardon the delay as well, thanks for your patience.

rbrondgeest:
Looks good to me, thanks for your guidance!

A next step, after merging this, would be to expose some possibilities to users, as it could be a bit trivial to figure out how this makes things more extensive.

rbrondgeest:
@blast-hardcheese can we proceed? Would like your approval :)

blast-hardcheese:
@rbrondgeest Yep, I was trying to coordinate the release with #1033, but I have no further concerns about merging this as it stands. If you could open a subsequent PR at your convenience to add an integration test that exercises this behaviour, it would be appreciated, but I'm not going to move the goalpost here 😉

blast-hardcheese (approved):

blast-hardcheese:
0.63.0 has been released! Thanks for your patience!

rbrondgeest:
For reference, here is an example of how we are using this now:

```scala
object DetailedJsonParseError {
  final case class DetailedJsonParseError(failures: NonEmptyList[DecodingFailure]) extends MessageBodyFailure {
    override def cause: Option[Throwable] = None

    private def detailedMessages = {
      failures.map(f =>{
        CursorOp.opsToPath(f.history) + " -> " + f.message
      })
    }

    def message: String = s"Malformed message body"

    def toHttpResponse[F[_]](httpVersion: HttpVersion): Response[F] =
      Response(Status.BadRequest, httpVersion)
        .withEntity(s"The request JSON has incorrect structure: " + detailedMessages.toList.mkString(", "))(EntityEncoder.stringEncoder[F])
  }

  trait WithDetailedJsonParseError extends CirceInstances {
    override def jsonDecodeError: (Json, NonEmptyList[DecodingFailure]) => DecodeFailure = {
      (json, failures) => DetailedJsonParseError(failures)
    }
  }
}
```

Adding it to resources like this:

```scala
val shipmentLabels = new v1.gen.label.LabelResource[Task] with WithDetailedJsonParseError)

val externalRoutes = Router("/" -> shipmentLabels)
```

Giving an error to HTTP/REST clients that looks like this:
`The request JSON has incorrect structure: .shipment.shipmentDetails.weight.unitCode -> KRT not a member of UnitCode`

blast-hardcheese:
that's great! Improving error messages is still something I'm thinking about how best to do, since anything with compile-time guarantees will invariably be onerous on the consumer.

PR #1028, opened 2021-03-17
ci: add JDK 16

sullis:


.github/workflows/ci.yml
```diff
@@ -76,7 +76,7 @@ jobs:
     needs: [core]
     strategy:
       matrix:
-        java: [ '8', '15' ]
+        java: [ '8', '15', '16' ]
```
blast-hardcheese, line 79: When modifying these matrix builds, please make sure to keep the total up to date: https://github.com/guardrail-dev/guardrail/blob/master/codecov.yml#L4-L11

blast-hardcheese: This is a workaround for what seems to be a bug in codecov where it just fails the build and doesn't recover even when subsequent submissions come in.

blast-hardcheese: thank you

blast-hardcheese (approved):
Verified that 24 is correct by checking the total reported runs [here](https://codecov.io/gh/guardrail-dev/guardrail/commit/de9d2bb041e3146a5a3a6f5fd5f14d78c355e70f/build), thank you for keeping that up to date!

PR #1030, opened 2021-03-18
Non string enumerations

### LINKED ISSUE #1029: type generation on enum fails
Schema declaration will fail as the EnumContainer will reference MyEnumType, but no type exists. Either a type alias in the package object or some type of coproduct is necessary to get the code to compile. Behavior same on client and server for both http4s and akka-http

```yaml
# This is in the components/schemas section
MyEnumType:
  type: integer
  enum:
    - 0
    - 1
    - 2
    - 3
 EnumContainer:
   type: object
   properties:
     myEnum:
       $ref: "#/components/schemas/MyEnumType"
 ```
 
 Generates
 
 ```scala
 case class EnumContainer(myEnum: Option[MyEnumType] = None)
 ```
 
 But no `MyEnumType` was generated.
 
 Can be fixed with 
```scala
 package object definitions {
   type MyEnumType = Int
   ... rest of code-gen already here
}
```

blast-hardcheese:
Resolves #1029

Permits generation of `Int` and `Long` enumerations. Round-trip tests need to be written, though it's reassuring that initial tests compile.


modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -233,7 +248,7 @@ object JacksonGenerator {
                 new BlockStmt(
                   new NodeList(
                     new IfStmt(
-                      new MethodCallExpr("value.name.equals", new NameExpr("name")),
+                      new MethodCallExpr("value.name.toString().equals", new NameExpr("name")),
```
kelnos, line 251: I would really much prefer something like
```scala
tpe match {
  case pt: PrimitiveType => new BinaryExpr(/* use '==' to check */)
  case other => // use toString
}
```

kelnos (changes requested):

modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/JacksonGenerator.scala
blast-hardcheese: I was just trying to preserve the `fromString` function type, but the more I think about it, we really should be `MyEnum.fromValue(5l): Optional<MyEnum>` or something, having the methods be hard-coded to `String` seems like a mistake

kelnos: Unfortunately we have to provide `fromString()` as that's a magic method name that Jersey will look for if you use an enum type as a resource method parameter.

kelnos: (I'm cool with adding a safer version for our own use, though.)

blast-hardcheese: So, after doing some offline conversation, I think leaving this the way it is actually makes the most sense, at least for now. `.toString()` should be a no-op on `String`, and other types will need to be looked up with this magic function via Jersey, so we can't use `==` as you indicate in the original comment.

Please feel free to correct me if I misunderstood, but I'll consider this resolved for now.

kelnos (approved):
lgtm

Peppi-Ressler:
Hi, currently this works only for enum objects but does not work for anonymous non string enums. Is there plan to fix that? Thanks

blast-hardcheese:
@Peppi-Ressler No plan currently, would you be so kind as to open a related issue to this one with a reproducing test case?

PR #1032, opened 2021-03-18
add guardrail presentation:  2021-03-18

sullis:


blast-hardcheese:
I'm concerned with binary files in master, but I also think `gh-pages` gets wiped every release. LFS is cumbersome as well, is there somewhere else we can put these files? Maybe a dedicated repo for learning materials in the new org?

kelnos:
Yeah, I don't love binary files either... I'd be down for a dedicated repository.  Might be easier to make these sorts of things more discoverable with its own repo.  i'd also be in favor of just putting it in the gh_pages stuff, though (if @blast-hardcheese hasn't done it yet) it might make sense to move the site into its own repo entirely (tho not sure if that will make generation more complicated).

sullis:
Closing this PR.

PR #1033, opened 2021-03-19
Add user-settable Dropwizard request timeouts

kelnos:
Dropwizard (well, Jersey) allows setting a timeout on the `AsyncResponse` instance passed to every request method invocation.  I've implemented two things here (both for the Java and Scala DW backends):

First the resource method constructor now takes a `RequestTimeout` param (POJO/case class to encapsulate a `Duration` and a function that returns a timeout `Response`).  In each resource method we call `asyncResponse.setTimeout()` and `asyncResponse.setTimeoutHandler()`.  The old constructor (that does not take a timeout) is also maintained for source compatibility; in that case, no timeout is set.

The second bit is to allow the timeout to be overridden on a per-operation basis. Say you have a bunch of normal REST-y JSON GET/POST endpoints that should usually return in 200ms or whatever, but then you have one endpoint that downloads a file, so you want to set the timeout to 30 seconds.  The way I've done this is, for each operation method, it also emits a `get${operationName}RequestTimeout` method.  By default, it just returns the class's common request timeout, but users can subclass and override these to provide different timeouts.

I know we're not a huge fan of subclassing around here, but I'm not sure of a better way to implement this second feature.  For the Scala backend I suppose an alternate constructor/apply could expose a defaulted method param for each, so something like this:
```scala
object FooRoutes {
  def withTimeouts(
    handler: FooHandler,
    defaultRequestTimeout: RequestTimeout,
    getBlahTimeout: Option[RequestTimeout] = None,
    getBarTimeout: Option[RequestTimeout] = None,
    postBarTimeout: Option[RequestTimeout] = None
  ) = ???
}
```
... but that has two issues: 1) it's kinda ugly, and 2) it doesn't allow someone to override to set _no_ timeout (`Option[Option[RequestTimeout]]`, anyone? ugh).

But I guess subclassing in the traditional sense isn't really _required_ in Scala at least; someone could do:
```scala
environment.jersey.register(new FooRoutes(handler, timeout) {
  override val getBlahRequestTimeout: Option[RequestTimeout] = ???
})
```

On the Java side, there just really isn't a great way of doing this without subclassing and overriding methods, since Java doesn't have default arguments.

Of course, it's possible there is a better way that I haven't thought of!

Also, style/idiom question: for Scala, `java.time.Duration`, or `scala.concurrent.duration.Duration`?


blast-hardcheese:
Easy thing first

> Also, style/idiom question: for Scala, `java.time.Duration`, or `scala.concurrent.duration.Duration`?

j.t.Duration is preferred, as s.c.d.Duration existed before joda got merged into the stdlib, and in some documentation somewhere(?) it was advised to prefer joda/java.time over Scala's duration when working with application code.

I don't remember where I read this, but if we're looking at maximal interoperability, java.time is fine for the most part.

blast-hardcheese:
For the timeout method question, this has two major considerations:

1. Anything decided here will significantly impact the direction of how route-based customization works (akka-http `Directive` injection, for example). Thinking about how akka-http's `File` form-field parameter mapper works, it leaves an unimplemented method similar to what you've described here for timeouts, so there is prior art to suggest that your approach here is the right one.
2. While I actually like the configuration object you described, even if it had a builder or something, something unpleasant from the users' perspective would end up being the separation of configuration from implementation. This is one of the advantages of annotations, in my opinion, although they can certainly be taken too far.

In an alternate universe, we have a construction like

```scala
new MyHandler {
  val myOperation = new MyOperation(timeout, authRequirements) {
    def tempFileMapper = (_, _, _) => File.mkTemp()
    def apply(respond: MyOperationResponse.type)(foo: Foo, bar: Bar) = ???
  }
}
```
where each method exists within its own namespace.

1. This avoids namespace collision (What happens if one of my operations is `foo`, leading to `getFooTimeout`, and another is `getFooTimeout` (first off leading to `getGetFooTimeoutTimeout`, secondarily conflicting with the generated `getFooTimeout` method)
2. This permits route-level configuration to be passed to the method class constructor, structurally promoting it to a high-visibility position (with the possibility of reuse, as these are just values)

For better or worse, we do not exist in that universe, but I'm inclined to think we're running into some serious limitations of the one we're in.

blast-hardcheese:
Last note, on the question of `Option[Option[RequestTimeout]]`, there's no reason we can't generate a `DefaultTimeout`/`CustomTimeout(RequestTimeout)`/`NoTimeout` type in the `support` package.

kelnos:
> This avoids namespace collision (What happens if one of my operations is foo, leading to getFooTimeout, and another is getFooTimeout (first off leading to getGetFooTimeoutTimeout, secondarily conflicting with the generated getFooTimeout method)

Heh, I thought of this, but we're safe here: the timeout getters are always zero-arity methods, and the resource methods will always take at least 1 argument (the Jersey `AsyncResponse` instance), so while the methods might end up having the same name, that will always be ok.

> Last note, on the question of Option[Option[RequestTimeout]], there's no reason we can't generate a DefaultTimeout/CustomTimeout(RequestTimeout)/NoTimeout type in the support package.

Good call.  I like this, and it's even not _that_ awkward to do in Java.

> While I actually like the configuration object you described, even if it had a builder or something, something unpleasant from the users' perspective would end up being the separation of configuration from implementation.

Oh, maybe I wasn't clear on that.  The `FooRoutes.withTimeouts()` function would actually return the routes instance, not a config object.  Oh, or maybe you mean the fact that it puts configuration in the routes instance instead of the handler instance?  Yeah, I can see that as being undesirable.  Certainly this same functionality could be moved to the handler; what do you think?  (Then again, the handler methods _can_ be 0-arity, so the possible name collision returns.)

On the other hand, thinking about it more, I do in some ways _like_ the separation of config/impl.  Config goes on the routes, and impl goes in the handler.  🤷🏻‍♂ eh I could go either way on that.

I _really_ like your "alternate universe" idea, but yeah, alas, that would of course be a big breaking change.  On the other hand, I think this approach kinda violates the first part of the "make the common case easy and the uncommon cases possible" principle; having to provide an implementation of an object rather than just a method for each operation isn't a _huge_ deal, but it does introduce some friction.

blast-hardcheese:
> I _really_ like your "alternate universe" idea, but yeah, alas, that would of course be a big breaking change. On the other hand, I think this approach kinda violates the first part of the "make the common case easy and the uncommon cases possible" principle; having to provide an implementation of an object rather than just a method for each operation isn't a huge deal, but it does introduce some friction.

The current interface could just supply the objects for the more complex variant, migration would need to be frictionless though.

blast-hardcheese:
I think we can fully exclude `FooRoutes.withTimeouts` as a possibility if only because it makes configuration opt-in painful, as you indicated. Due to the lack of optional parameters in Java, the first time you need explicit timeouts, you end up adding tens or hundreds of these effective no-op parameters, even if you end up setting them all to `Default` it's still excessively painful.

The override is begrudgingly the simplest way to do it, with the advantage of not precluding the alternate universe strategy should we be pushed in that direction either from feature requests or attempted reduction in... perceived(?) complexity. I think the alternate universe approach will need to be tested out in a real-world scenario before we make a decision.

blast-hardcheese:
Closing this out due to stale. If someone wants to take this on to modernize it, it's likely not going to have too terribly much in the way of merge conflicts, but I don't have the time or inclination to do this now (especially with the DW2.x upgrade in progress in #1299)

PR #1052, opened 2021-04-02
Update org

blast-hardcheese:
Repackage sources from `com.twilio` -> `dev.guardrail` to match new GitHub org.

I _think_ before this gets merged it would be good to add a scalafix rewrite rule to scala-steward.

The rules:

- `import com.twilio.guardrail` -> `import _root_.dev.guardrail`, to avoid possible conflicts considering the new root (`dev` is somewhat of an awkward root package, so `_root_` for safety)
- `com.twilio.guardrail.FooBar` -> `_root_.dev.guardrail.FooBar` for the same reason, but to handle `Term.Select`'s

Opening this PR to announce intent


kelnos (approved):
yessssss

kelnos:
i do find the soft-need to prefix with `_root_` to be kinda icky, though i do agree that it's a good idea.

blast-hardcheese:
> i do find the soft-need to prefix with `_root_` to be kinda icky, though i do agree that it's a good idea.

Missing where that happened in this PR -- the only reason the `io` package is prefixed is because it conflicts with some libraries, but I'm not seeing it in the diff here

kelnos:
no i mean the need for _customers_ to prefix with `_root_` if they use guardrail as a library.

blast-hardcheese:
> no i mean the need for _customers_ to prefix with `_root_` if they use guardrail as a library.

Ah. Yes. Such is life with these new TLDs though. Is `dev` common enough as a package that this will be an issue?

kelnos:
it's probably not that big a deal, i guess.

PR #1067, opened 2021-04-13
HTTP4s - Add support for subtypes of application/json 

rbrondgeest:
Small change to allow for subtypes of application/json contenttype to be recognized for generating entityEncoder/decoder 

For example useful when types such as `application/problem+json` are used to indicate more specifically what kind of content an entity is represented by.


blast-hardcheese (approved):
Looking at [ietf: Additional Media Type Suffixes](https://trac.tools.ietf.org/html/rfc6839#section-3.1), this looks like the correct approach. I presume that everything else in the rfc regarding structure here would be done directly in the OpenAPI specification, so no further work is required to support this.

Thanks for fixing this!

PR #1075, opened 2021-04-21
Use new annotation for Unwrapping used as of Hibernate Validator 6.0.0

manuelbernhardt:
See https://hibernate.org/validator/documentation/migration-guide/ - `org.hibernate.validator.valuehandling.UnwrapValidatedValue` no longer exists and thus using a recent version of Dropwizard (2.x onwards) means that there's a clash between versions of the Hibernate Validator library

### DIFF
--- a/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
+++ b/modules/codegen/src/main/scala/com/twilio/guardrail/generators/Java/DropwizardServerGenerator.scala
@@ -224,7 +224,7 @@ object DropwizardServerGenerator {
         "javax.ws.rs.core.MediaType",
         "javax.ws.rs.core.Response",
         "org.glassfish.jersey.media.multipart.FormDataParam",
-        "org.hibernate.validator.valuehandling.UnwrapValidatedValue",
+        "javax.validation.valueextraction.Unwrapping",
         "org.slf4j.Logger",
         "org.slf4j.LoggerFactory"
       ).traverse(safeParseRawImport)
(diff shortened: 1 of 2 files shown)


kelnos:
Does this annotation exist in the versions of hibernate validator that older DW versions depend on?

manuelbernhardt:
@kelnos indeed they do not. So I'm wondering about the following: should I
- upgrade dropwizard and its dependencies in the dev projects
- only bump the version for dropwizard java
- introduce a new dropwizard2 module / artifact (not quite sure how best to achieve this)

?

kelnos:
I think there are some other issues that will make DW2 not work with guardrail, though I don't have a list offhand.

I think my preference would be to renamed the current module/framework to `dropwizard1`, and add an alias with the current name, that prints a deprecation warning.  Then we can add `dropwizard2` that supports DW2.  Obviously this should not just be a copy-paste of the existing generators, but a proper refactoring, since there will only be a few small differences between them.  In addition, we'd want to duplicate the sample module so that both DW1 and DW2 are covered by the tests during build.

Are you interested in taking that on?  It's something I really want to get done, but realistically won't have the bandwidth to work on for a while.  I can definitely offer answers to any questions you have as you dive in.

hjizettle:
This is preventing us now.  DW currently pulls in `hibernate-validator-6.1.17.Final` which does not have this class.

blast-hardcheese:
@hjizettle To be clear, dropwizard 1.x is what pulls in `hibernate-validator-6.1.17.Final`, or have you been using DW2.x?

hjizettle:
Indeed we are using DW2 @blast-hardcheese

blast-hardcheese:
Incorporated this change into #1299, thank you

PR #1077, opened 2021-04-22
Removing last two warnings

blast-hardcheese:


kelnos (approved):
we're warning-free now?  woohoo!

blast-hardcheese:
@kelnos Yep! Good stuff 😄

PR #1078, opened 2021-04-23
Only permit generating a body parameter for non-form content types

### LINKED ISSUE #313: JSON body param with embedded schema generates FormData

blast-hardcheese:
Resolve #313 
Extending from #424, I didn't realize that @sbrunk already wrote exactly this before, but I hadn't encountered it myself, so I'm using the metadata from that commit to recognize the effort put forward at that time.

### DIFF
--- a/modules/codegen/src/test/scala/core/issues/Issue313.scala
+++ b/modules/codegen/src/test/scala/core/issues/Issue313.scala
@@ -1,0 +1,93 @@
+package core.issues
+
+import cats.data.NonEmptyList
+import com.twilio.guardrail.generators.Scala.AkkaHttp
+import com.twilio.guardrail.generators.syntax.Scala.companionForStaticDefns
+import com.twilio.guardrail.{ Client, Clients, Context }
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+import support.SwaggerSpecRunner
+
+import scala.meta._
+
+class Issue313 extends AnyFunSuite with Matchers with SwaggerSpecRunner {
+  val swagger: String = s"""
+    |openapi: 3.0.2
+    |info:
+    |  title: Whatever
+    |  version: 1.0.0
+    |servers:
+    |  - url: http://localhost:1234
+    |paths:
+    |  /users/{name}:
+    |    patch:
+    |      operationId: changeUser
+    |      parameters:
+    |      - name: name
+    |        in: path
+    |        required: true
+    |        schema:
+    |          type: string
+    |      requestBody:
+    |        required: true
+    |        content:
+    |          multipart/form-data:
+    |            schema:
+    |              type: object
+    |              required:
+    |                - id
+    |              properties:
+    |                id:
+    |                  type: integer
+    |                  format: int64
+    |      responses:
+    |        '204': {}
+    |""".stripMargin
+
+  test("Test in body generation") {
+    val (
+      _,
+      Clients(Client(tags, className, imports, staticDefns, NonEmptyList(Right(cls), _), _) :: _, Nil),
+      _
+    )       = runSwaggerSpec(swagger)(Context.empty, AkkaHttp)
+    val cmp = companionForStaticDefns(staticDefns)
+
+    val client = q"""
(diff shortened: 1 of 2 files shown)


modules/codegen/src/test/scala/core/issues/Issue313.scala
```diff
@@ -0,0 +1,92 @@
...
+    |        required: true
+    |        content:
+    |          multipart/form-data:
+    |            schema:
+    |              required:
+    |                - id
+    |              properties:
+    |                id:
+    |                  type: integer
+    |                  format: int64
```
kelnos, line 41: I thought this bug only triggered when `type: object` is explicitly specified. At least my workaround was always to omit that...

blast-hardcheese: I ran into this again with the Twilio OpenAPI specs -- it should be possible to generate clients and servers directly from the specifications, unmodified.

kelnos: Right, but I mean that this example spec here always worked fine for me; that is, even before this fix, there was no extra body parameter present.  To trigger the bad behavior I always had to add `type: object`.  I think there was actually a sample where I commented it out with the hopes that we'd fix it someday.  Right, yeah, here: https://github.com/guardrail-dev/guardrail/blob/master/modules/sample/src/main/resources/multipart-form-data.yaml#L17-L22

(It's also possible that I hackily "fixed" this in the Java server & client generators by ignoring `bodyParams` if `formParams` was non-empty.)

blast-hardcheese: Yeah -- Removing `type: object` is the workaround for this issue, but manually modifying generated OpenAPI specifications is untenable.

With this patch, that comment is resolved, since it prevents the `ObjectSchema` from being processed in the case of form. I tried generating AsyncHttpClient, fwiw, and the useless `body` was emitted for that as well, so if there is a workaround hidden in the Java code, it's not in the client generator.

PR #1114, opened 2021-05-22
[WIP] Add Http4s documentation to microsite

efossier:
First pass at adding docs for `http4s` client and server generation to the guardrail docs microsite.

There is a fair amount of common content between this (copy pasted) and the other modules, for example "What is guardrail" etc. Before merging this I would like to investigate whether we could consolidate these into a common directory so we can update them once for all languages/frameworks. 

This seems to be possible by doing relative links like so:
Before:
```markdown
Table of Contents
=================

1. [What is guardrail](what-is-guardrail)
    1. [Single Point of Truth](what-is-guardrail#single-point-of-truth)
```
After:
```markdown
Table of Contents
=================

1. [What is guardrail](../../common/what-is-guardrail.md)
    1. [Single Point of Truth](../../common/what-is-guardrail.md#single-point-of-truth)
```

However the above will most likely cause problems with the backwards/forwards links on the individual pages, since `what-is-guardrail.md` in each framework links to the next page which may be specific to the framework.

Additionally I was not able to build the docs locally by running:
```shell
$ sbt
> mdoc
```
This throws many errors related to rendering the code snippets and also the links, i'm not sure how the linking works between pages now tbh.

Finally, I would like some suggestions for any other useful things we should add to the new Http4s docs.


kelnos (approved):
Yeah, we really do need to revamp the docs to move the generic 'what is guardrail' stuff out into a common location.  @blast-hardcheese are you ok with just merging this in the meantime?  This is great work and should be really useful, better to get it into our docs rather than waiting on a docs reorg, I think.

blast-hardcheese:
Certainly, though I _think_ we ended up hitting something of a roadblock, then I was suddenly overcome by events and forgot to circle back. I've got a changeset that needs to be applied before this gets merged, I believe the generated mdoc hilariously includes a stacktrace, if I remember correctly.

blast-hardcheese:
Alright. It took a bit longer than anticipated due to a detour over in @ghostdogpr's excellent Caliban code generator for GraphQL, but this looks good to me now.

@efossier and I talked about the possibility of a docs restructure to deduplicate the commonalities, as well as to streamline the "getting started" stuff, but I don't think that should block this from going out.

blast-hardcheese (approved):

modules/microsite/src/main/scala/DocsHelpers.scala
```diff
@@ -74,6 +74,68 @@ object DocsHelpers {
...
+          Some(server.handlerDefinition.toString),
+          Some(""),
+          Some(q"""
+            class ${oname}[..${tparms}](...${resourceParams}) extends ..${xtends} {
+              ..${routeStats};
+              `...`
+            }
+          """.toString)
+        )
+      case (Http4s, GeneratingClients) if false =>
```
blast-hardcheese, line 104: lol, explicitly disabled this to test `Unknown docs sample` but then forgot to uncomment it. Fixing in #1151

PR #1120, opened 2021-05-26
add json derived content types to code generator

andrealexandre:
add json derived content types to code generator


kelnos (changes requested):
Two issues here:

1. This only works in one direction, for servers.  Now clients will also match on this, but if the spec says `application/problem+json`, generated clients will still send `application/json`.
2.  On the server side, this will allow non-conforming requests through.  If a spec says `application/json`, and a client sends `application/problem+json`, the server will still accept it, when it should not.

The right fix here for the client side is to turn `ApplicationJson` into a `case class` that captures the actual value from the spec, and uses it rather than hard-coding `application/json` (this will need to be updated in all client generators to avoid generating buggy clients when this feature goes out).

On the server side it's a little more of a pain, because those unmarshallers are static boilerplate rather than being generated based on the actual types specified.  That will require a bit more surgery to fix.

PR #1140, opened 2021-06-16
Pin various eclipse java formatter versions

kelnos:
Some of these packages have started being compiled against Java 11, which means users building their projects with a JDK8 distribution will fail to build (even though the generated code could of course run on Java 8).  These packages are somewhat odd in that many specify an allowed version range, rather than a specific version, of their dependencies.  So all of our existing releases through summer 2019 have likely stopped working over the last month or so when the toolchain used is JDK8.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,6 +41,31 @@ val javaxElVersion         = "3.0.0"
 val vavrVersion            = "0.10.3"
 val dropwizardVavrVersion  = "1.3.0-4"
 
+// TAKE CARE WHEN UPDATING THESE
+val eclipseFormatterDependencies = Seq(
+  "org.eclipse.jdt" % "org.eclipse.jdt.core" % "3.24.0",
+  // These version pins are necessary because a bunch of transitive dependencies
+  // are specified via an allowed version range rather than being pinned to a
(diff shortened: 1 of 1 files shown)


blast-hardcheese (approved):

blast-hardcheese:
Thanks for looking into this -- Presumably others are running into this issue, is there a github issue against Eclipse that we can link to this PR if this gets resolved upstream?

kelnos:
Not that I know of.

kelnos:
Looks like they're not on GH: https://bugs.eclipse.org/

I don't see anything related with a quick search (tho the search is awful), and I'm not even sure where to file a bug there.  Maybe under the JDT?  The dependencies in question?  Hm.

blast-hardcheese:
Yeah, or worst case switch formatters or not format the code that gets generated. This is a great fix to get us back up and running, but without shading the deps or otherwise insulating us from future changes I anticipate this turning into a maintenance chore until the world gets off jdk8

kelnos:
For now we can get by just not updating the formatter at all, or being sure to verify that new versions don't require one of the Java11-compiled dependencies.  I'm fine with the former solution.

PR #1168, opened 2021-07-19
Ensure deser failure when primitive type is missing in Jackson DTOs

kelnos:
If a primitive type is specified but no value is present in the JSON string, Jackson will pass its zero value to the constructor.  Using
boxed types in the constructor allows Jackson to pass `null` to the constructor, which we can catch.

### DIFF
--- a/modules/codegen/src/main/scala/dev/guardrail/generators/Java/JacksonGenerator.scala
+++ b/modules/codegen/src/main/scala/dev/guardrail/generators/Java/JacksonGenerator.scala
@@ -126,7 +126,6 @@ object JacksonGenerator {
                     new AssignExpr(
                       new FieldAccessExpr(new ThisExpr, term.parameterName),
                       term.fieldType match {
-                        case _: PrimitiveType => new NameExpr(term.parameterName)
                         case ft if ft.isOptionalType =>
                           new ConditionalExpr(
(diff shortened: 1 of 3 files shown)


kelnos:
NB: the scala version of jackson generation also has this bug, but since scala doesn't have primitive types, and jackson-module-scala treats scala's non-primitive primitives by filling in zero values when they're missing, I'm not really sure how to fix this without some major refactoring.

modules/codegen/src/main/scala/dev/guardrail/generators/Java/JacksonGenerator.scala
```diff
@@ -126,7 +126,6 @@ object JacksonGenerator {
                     new AssignExpr(
                       new FieldAccessExpr(new ThisExpr, term.parameterName),
                       term.fieldType match {
-                        case _: PrimitiveType => new NameExpr(term.parameterName)
```
kelnos, line 129: The idea is to remove this because this case will never happen anymore; we _always_ want to wrap these in `requireNonNull()` even if the underlying (field) type is a primitive, the constructor param will always be boxed.  In reality, though, we don't actually need to use `requireNonNull()` here, because attempting to assign a `null` boxed value to its corresponding primitive type will throw an NPE.  But `requireNonNull()` is useful to show intent, and to get a better error message.

PR #1171, opened 2021-07-21
Dynamically set scala matrix versions from build.sbt versions

### LINKED ISSUE #1130: Find a way to keep the scala versions in the GH Actions CI workflow in sync with build.sbt
In order to parallelize the CI builds, we run each scala cross version in a separate job.  To do this, we need to specify the specific versions on the sbt command line in the CI workflow. Unfortunately this means that when scala-steward submits a PR to bump the scala version in build.sbt, the CI workflow will lag behind if we don't remember to update it.

Often this isn't _terrible_, but this time it caused a build failure because the kind-projector and wartremover compiler plugins needed to be bumped as well as the scala version (the versions of the former we were using were not built for the new scala compiler version).  The CI build did not catch this issue because it still built using the old scala version.

For reference: https://github.com/guardrail-dev/guardrail/pull/1129

blast-hardcheese:
Resolves #1130

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
 
 jobs:
   core:
+    name: Run core tests
     runs-on: ubuntu-20.04
     strategy:
       matrix:
@@ -18,6 +19,8 @@ jobs:
         scala: [
             { version: '2.12.14', bincompat: '2.12' }
           ]
+    outputs:
+      scala_versions: ${{ steps.set-scala-versions.outputs.scala_versions }}
     steps:
     - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -39,6 +39,11 @@ jobs:
       with:
         file: ./modules/codegen/target/scala-${{ matrix.scala.bincompat }}/scoverage-report/scoverage.xml
         fail_ci_if_error: true
+        steps:
+        - id: set-scala-versions
+          run: echo "::set-output name=scala_versions::$(sbt 'print codegen/crossScalaVersions' | grep '^\* ' | sed 's/^\* //')"
```
kelnos, line 44: `grep | sed` can be replaced with `awk '/^\* /{ print $2 }'`if you are so inclined

kelnos (approved):

.github/workflows/ci.yml
blast-hardcheese: Ended up needing to do it in Scala anyway, but good call

blast-hardcheese:
It took a bit to convince sbt to emit something usable here, so this shouldn't be merged unless the various scala versions are tested.

PR #1197, opened 2021-08-23
Module decomposition

blast-hardcheese:
Addresses the core repo decomposition described in #1195 

Some structural changes to support this:

- `codegen` sbt project has been renamed `guardrail`
- `xyzSample` projects have been renamed `sample-xyz`

The following projects exist:
```
[info]     guardrail
[info]     guardrail-core
[info]     guardrail-java-async-http
[info]     guardrail-java-dropwizard
[info]     guardrail-java-spring-boot
[info]     guardrail-java-support
[info]     guardrail-scala-akka-http
[info]     guardrail-scala-dropwizard
[info]     guardrail-scala-endpoints
[info]     guardrail-scala-http4s
[info]     guardrail-scala-support
```

all of which are both depended on by the `guardrail` project.

I believe this can be released as-is, as a minor version bump, as the majority of the packages haven't changed (except for moving some members out of `SwaggerUtil`.

There are three files left in the main `guardrail` project:

```
  > modules/codegen/src/main/scala/dev/guardrail/generators/ScalaModule.scala
  > modules/codegen/src/main/scala/dev/guardrail/generators/JavaModule.scala
  > modules/codegen/src/main/scala/dev/guardrail/CLI.scala
```

- `ScalaModule` and `JavaModule` both knit together all the various different modules and provide the infrastructure required to select what frameworks/modules you want to the build tooling
- `CLI` is the main CLI runner

PR #1198, opened 2021-08-25
Debug body serialization

arainko:
Currently when the deserialization of a JSON body fails we get a generic message (`The request body was malformed.`) with a `422` status code, which proved to cause some misunderstandings when trying to integrate our backend with a frontend especially when the input JSON document is on the larger side.

This PR tries to somewhat address that by exposing serialization errors in the response (currently only for `http4s`).

Is this the right way? I know we can achieve a somewhat similar result by using the `mapRoute` mechanism, but that requires additional boilerplate.

If it is the right way, then given that I introduced a new argument (`--debug-body`) all plugin projects should also be updated (or at the very least the `Scala` ones) to accommodate the change, right?


blast-hardcheese:
So, this is certainly useful, and already supported in akka-http without a flag:
```
guardrail-sample-sbt-akkahttp (master) dstewart$ curl localhost:8080/v1/pets -X POST -H 'Content-Type: application/json' --data '{}'
The request content was malformed:
Attempt to decode value on failed cursor: DownField(name)
```

I think that comes from [here](https://github.com/guardrail-dev/guardrail/blob/master/modules/scala-akka-http/src/main/scala/dev/guardrail/generators/Scala/AkkaHttpServerGenerator.scala#L593),
```scala
                          Unmarshaller { implicit executionContext => part =>
                            ${unmarshaller}
                              .apply(part)
                              .map(${Term.Select(partsTerm, containerName.toTerm)}.apply)
                              .recoverWith({
                                case ex =>
                                  Future.failed(RejectionError(MalformedFormFieldRejection(part.name, ex.getMessage, Some(ex))))
                              })
```

I think it would be OK to support this in http4s as well, without a flag for explicitly enabling this feature.

(btw, despite the conflicts incurred from merging #1197, I rebased this on master with no merge conflicts, no worries there.)

arainko:
Thanks for the feedback @blast-hardcheese!

So the course of action is: rebase, make the serialization error response a default one, delete the `--debug-body` flag.

I'll get to it tomorrow :smile:

arainko:
@blast-hardcheese if you could take another look at it :pray:. 

I've also updated `Issue148` tests, which referenced the old generic error responses.

modules/sample-http4s/src/test/scala/core/Http4s/Http4sDebugBodyTest.scala
```diff
@@ -0,0 +1,74 @@
...
+    val actualErrorMessages =
+      Stream
+        .resource(request)
+        .flatMap(_.bodyText)
+        .compile
+        .toList
+        .unsafeRunSync()
+
+    val expectedErrorMessage =
+      "The request body was malformed. Underlying message: Attempt to decode value on failed cursor. Problematic fields: something1"
```
blast-hardcheese, line 57: Since you have functionality for multiple errors reported, would you mind adding a third test for when both `something1` and `something2` were invalid? If that's desired to be supported (I think so), it would be good to have a test for it.

Otherwise, just surfacing the raw exception from circe, to ensure we get the most information from the underlying exception.

arainko: I think that'd require swapping `jsonOf` to `accumulatingJsonOf` when generating decoders to make it work and I actually did at first but couldn't extract all the errors at once for some reason, it was always just the first one.

I'll try once again and report back 👍

blast-hardcheese: No worries, thanks for taking a look.

If you do get it working, I think concatenating each `toString` of each of the validation errors, which would likely give more context for each error, one per line.

arainko:
Alright I'm back with more info, turns out `accumulatingJsonOf` is tied to method implementations in `CirceInstances` (despite the docs saying it'll always return a cause of `org.http4s.DecodingFailures`), namely `jsonDecodeError` and if you override it inside the `Resource` class, like so:

```scala
override protected def jsonDecodeError: (Json, NonEmptyList[DecodingFailure]) => DecodeFailure = (json, failures) => {
    val failureStr = failures.mkString_("", ", ", "")
    InvalidMessageBodyFailure(
      s"Could not decode JSON: ${json}, errors: $failureStr",
      cause = Some(DecodingFailures(failures))
    )
  }
```
you SHOULD get the desired behavior... But you don't :shrug:, despite my attempts the errors were not accumulating I kept on only getting the first one (eg. missing multiple field in a JSON object, mistyping the values etc). With a schema like this:

```yaml
  Entry:
      type: object
      properties:
        name:
          type: string
        name2:
          type: string
        icon_url:
          type: string
        desc:
          type: string
      required:
        - name
        - name2
```

and a request with a body like this: 

```json
{
  "nameWrong": "woohoo",
  "name3": "you aint gonna work, are you"
}
```

I've got these errors: 

```scala
NonEmptyList(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(name))))
```
The decode logic looks like this:

```scala
accumulatingJsonOf[F, API.Entry]
          .decode(req, strict = false)
          .foldF(
            err =>
              err match {
                case InvalidMessageBodyFailure(msg, Some(DecodingFailures(failures))) =>
                  Response[F](
                    status = org.http4s.Status.UnprocessableEntity,
                    body = stringEncoder.toEntity(failures.toString).body
                 ).pure[F]
                case err => err.toHttpResponse[F](req.httpVersion).pure[F]
              },
            body => ...)
```

It's probably me doing something silly, but I'm at loss for now :crying_cat_face: 

With that said, I'll settle for the first-error-only version for now and try and come back with accumulating errors in a different PR once I get some more time for contributions :+1:

arainko:
Done, also fixed codegen tests that I missed the last time.

@blast-hardcheese if you could approve the CI check and take a look one last time, it'd be lovely :smile:

blast-hardcheese (approved):
Great stuff, thank you for your detailed work here!

PR #1200, opened 2021-08-25
Bring compatibility with http4s 0.22

sideeffffect:
closes https://github.com/guardrail-dev/guardrail/issues/1199
blocked on https://github.com/http4s/http4s/pull/5109


sideeffffect:
@blast-hardcheese can you please run the CI again, I can't do it :(

Regarding

```scala
scala> org.http4s.dsl.io.Path.apply("this/is/a/test", "").toString().init.tail
res0: String = this%2Fis%2Fa%2Ftest
```
vs
```
@ org.http4s.dsl.io.Path.Segment.apply("this/is/a/test").toString
res2: String = "this%2Fis%2Fa%2Ftest"
```
Aren't the `"` around the string an issue? Or is it better if it contains them?

Regarding the support for multiple http4s version -- I like the plan, but I'm not sure this is the right time to do that. http4s 0.21 is End Of Life anyway, so it's not worth doing it for the sake of 0.21.

The currently alive versions of http4s are 0.22 and 0.23, but I would hope that they are source compatible (the reason for 0.23 is AFAIK support for cats-effect 3 which is fundamental redesign over 2, but I would hope that http4s stays the same :smiley: ).

sideeffffect:
I've just checked, and the current changes work not only with 0.22, but also 0.23 _and_ 1.0.0-M24.

sideeffffect:
The tests are now fixed. Can you please run the CI? :pray:

sideeffffect:
I've fixed the tests again. Please run the CI :pray:

sideeffffect:
OK, hopefully now everything passes :crossed_fingers: Please give the CI one more kick :smile:

blast-hardcheese:
`runtimeScalaSuite` at the SBT console will exercise these classes, btw

sideeffffect:
OK, `runtimeScalaSuite` is green now. Let's try once again :smile_cat:

blast-hardcheese:
Well, I wasn't expecting to get a `StackOverflowException` from scalac in this PR, that's for sure

sideeffffect:
:sob:

sideeffffect:
Oh well... Scala 2.13 has been doing this since like 2.13.3 or 2.13.4 (I suspect since they re-implemented exhaustiveness checking for pattern matching :shrug:, but maybe it's something totally different). But what's worse, I can't reproduce it locally with `runtimeScalaSuite`.

Could you perhaps run it again? :pleading_face:

blast-hardcheese:
So the problem was `-Xss` being too small in whatever `sbt` wrapper GitHub Actions sets by default. TIL we're not actually customizing which sbt launcher we're using. I'll open a PR to swap the `setup-java` with `setup-scala` which will resolve this by at least giving us a stable `sbt` version.

blast-hardcheese (approved):
See #1218, unsure whether that was caused by this bump or new warnings from 2.13.x

blast-hardcheese:
🎉

PR #1210, opened 2021-08-28
Using slf4j-simple for development cli command

blast-hardcheese:
slf4j implementations are supplied by sbt and maven, so just hardcode slf4j-simple when running `cli` at the sbt prompt


build.sbt
```diff
@@ -405,6 +405,7 @@ lazy val core = commonModule("core")
     libraryDependencies ++= Seq(
       "com.github.javaparser"       % "javaparser-symbol-solver-core" % javaparserVersion,
       "io.swagger.parser.v3"        % "swagger-parser"                % "2.0.27",
+      "org.slf4j"                   % "slf4j-simple"                  % "1.7.32",
```
kelnos, line 408: you should generally never depend on a concrete slf4j logger impl in a library, because that will cause problems for consumers of the library who may want to use logback or whatever.

Issue #1195, opened 2021-08-22
Proposal to modularize the guardrail repo


blast-hardcheese:
Problem
---

Currently, guardrail has no explicit library dependencies brought into
consuming projects, permitting users to supply their own implementations
that satisfy source compatibility. This is a core design goal of guardrail,
and one that I aim to maintain.

One challenge, however, is that any change to any source generators for any
language necessitates a version bump for the whole project, leading to
unnecessary upgrade PRs and work for users, for explicitly no benefit.

Explicitly, a user who only uses Scala source generators should not have to
bump guardrail if only the Java generators have changed, and vice versa.

Proposed solution
---

### Core guardrail repo

The `guardrail` project will be further modularized, instead of emitting

    dev.guardrail::guardrail:x.y.z

we will instead release

    dev.guardrail::guardrail-core:a.b.c
    dev.guardrail::guardrail-${language}-${library}:x.y.z

where each individual module has its own `version.sbt`,
distinct from `guardrail-core`.

As it stands today, this will end up with the following:

    dev.guardrail::guardrail-core:a.b.c
    dev.guardrail::guardrail-scala-akka-http:d.e.f
    dev.guardrail::guardrail-scala-http4s:g.h.i
    dev.guardrail::guardrail-scala-dropwizard:j.k.l
    dev.guardrail::guardrail-java-dropwizard:m.n.o
    dev.guardrail::guardrail-java-spring-boot:p.q.r

### Build tools (sbt)

For `sbt-guardrail`, this will have two ramifications.

First, the `modules` configuration will move into sbt-guardrail.

Instead of:

```sbt
// project/guardrail.sbt
addSbtPlugin("dev.guardrail" % "sbt-guardrail" % "a.b.c")

// build.sbt
Compile / guardrailTasks += ScalaClient(file("server.yaml"), pkg="example.client", framework="http4s")
```

we would instead have:

```sbt
// project/guardrail.sbt
addSbtPlugin("dev.guardrail" % "sbt-guardrail" % "a.b.c")
libraryDependencies += "dev.guardrail" %% "guardrail-scala-http4s" % "d.e.f"

// build.sbt
Compile / guardrailTasks += ScalaClient(file("server.yaml"), pkg="example.client").modules(_.http4s)
```

where a proof-of-concept of how the modules might be implemented being

```scala
object GuardrailModules
def modules[L <: LA, F[_]](holders: (GuardrailModules.type => Holder[L, F])*): Types.Args = ...
```

[...]

PR #1213, opened 2021-08-29
Reflowing package structure

blast-hardcheese:
Draft, but here's progress so far. Continuing #1195 

More work needs to be done in `modules/core/src/main` to either break up disparate classes defined in the same file or to identify commonalities and define stable packages around them.

Finally, before this can be merged, a migration guide should be created to help maintain internal forks.

I'd like a tool to generate something like tag clouds but for JVM libraries, where distance between nodes is some factor of the number of references between those two classes scaled by how many other classes refer to one or the other.
The idea would be visually determining packages based on existing structure


blast-hardcheese:
Merging this to prevent future merge conflicts, as this is building towards 1.0.0

PR #1219, opened 2021-09-02
Swap setup-java for setup-scala

blast-hardcheese:
No idea how we even have `sbt` on the path. Use `olafurpg/setup-scala` to at least get a managed version from a trusted source.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        java: [ '15' ]
+        java: [ '1.15' ]
         scala: [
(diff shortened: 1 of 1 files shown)


kelnos:
yeah, that's weird.  i can't imagine the ubuntu 20.4 image has sbt installed by default, that'd be bizarre.

sullis:
> yeah, that's weird. i can't imagine the ubuntu 20.4 image has sbt installed by default, that'd be bizarre.


GitHub Actions pre-installs popular development tools into the Ubuntu Virtual Environment

https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md


```
Sbt 1.5.5
```

PR #1230, opened 2021-09-09
Multi release drafter

blast-hardcheese:
Modularizing release drafter, based on a fork of `release-drafter`. The intent is to open this as a PR once I've tested out the flow in the guardrail repo, but this should give us the same sort of release automation that we currently have, but for the different modules.


blast-hardcheese:
This might need a bit of tweaking, but [my test repo](https://github.com/blast-hardcheese/multi-release-drafter-test/releases) _seems_ to suggest this'll just work.

Gonna use my fork of `release-drafter` until the changes are merged upstream

PR #1236, opened 2021-09-14
http4s can return fs2.Stream[F, Byte] in OpenAPI 3.x

sideeffffect:


sideeffffect:
This is the error
```
[error] /home/ondra/Projects/guardrail/modules/sample-http4s/target/generated/binary3/client/http4s/Client.scala:46:36: type mismatch;
[error]  found   : binary3.client.http4s.OperationIdResponse.BadRequest.type
[error]  required: binary3.client.http4s.OperationIdResponse[F]
[error]         F.pure(OperationIdResponse.BadRequest): F[OperationIdResponse[F]]
[error]                                    ^
[error] /home/ondra/Projects/guardrail/modules/sample-http4s/target/generated/binary3/server/http4s/Routes.scala:51:38: pattern type is incompatible with expected type;
[error]  found   : binary3.server.http4s.OperationIdResponse.BadRequest.type
[error]  required: binary3.server.http4s.OperationIdResponse[F]
[error] Note: if you intended to match against the class, try `case BadRequest()`
[error]             case OperationIdResponse.BadRequest =>
[error]                                      ^
```
And these probably the places where it goes wrong
https://github.com/guardrail-dev/guardrail/blob/master/modules/scala-http4s/src/main/scala/dev/guardrail/generators/Scala/Http4sServerGenerator.scala#L651
https://github.com/guardrail-dev/guardrail/blob/master/modules/scala-http4s/src/main/scala/dev/guardrail/generators/Scala/Http4sClientGenerator.scala#L254
/cc @blast-hardcheese

blast-hardcheese:
@sideeffffect So this is a bit of a bigger problem. How would you expect bytestreams to be consumed? Buffer in memory? Pause the stream until the user function is called? What about if there are more than one file upload parameter?

sideeffffect:
@blast-hardcheese I'm not sure I understand why this should be a problem. It should be possible for the server to return to the client a `Stream[F, Byte]` regardless where that stream comes from. It could be for example from a large file, that doesn't even have to fit into memory.
Of course, this makes sense only in the case when the Stream is the _only_ (or last?) thing that the server sends to the client. Do you see what I mean?

blast-hardcheese:
> Of course, this makes sense only in the case when the Stream is the _only_ (or last?) thing that the server sends to the client. Do you see what I mean?

That's exactly my point -- what should happen in the case of two file parameters, or if the client sends the parameters out of order? Fail the request with a 400? I'm happy to proceed here, just looking for a strategy that is stable. Ideally, `x-scala-type` should not be required in this case, you should just get a `Stream[F, Byte]` natively.

blast-hardcheese:
Well, I'm happy to take a look once I'm done with the repackaging project. That's taking the majority of my free time currently, but I'm almost done. That's one of the last things blocking the 1.0 release, now that the modular versioning is working.

blast-hardcheese:
Also, this would also simplify the way file handling works in akka-http. Currently I just ask the user for a file handle that can be written to, then stream to the file, then hand the handle back to the user.

blast-hardcheese:
Circling back.

So, I see two opportunities for proceeding here:

1. If there is exactly one file at the very end of the payload, you get a raw `Stream[F, Byte]`, no extra handling. If the parameters are emitted in the wrong order, we have no choice but to emit a 400.

2. We can expose a helper method, `${routeName}HandleBytes${parameterName}` or something (named explicitly verbosely to avoid inadvertent name collisions), to take each `Stream[F, Byte]` and emit some `F[???]` once that stream has been consumed.

We could emit a `F[Stream[F, Byte]]`, which would lend itself to a default implementation of `F.pure(x)` for those functions, offered as a suggestion to the developer in the documentation (and guided by the types), instead of providing an unsafe definition by default and expecting the developer `override` if they want safety.


---

Presumably for the client, it would be easy enough to just sort the parameters so the streams are at the end, as we have control there.

blast-hardcheese:
@sideeffffect Do you know if there is any sort of default implementation semantics for http4s that either of these approaches would conflict with?

blast-hardcheese:
Also, completely missing the fact that this is just a "missing parenthesis" bug... hopefully this resolves at least the current issue, though the broader one should be resolved as well.

blast-hardcheese (approved):
Going to merge and release this for now, as clearly this is an omission in the http4s `generic` generator. The reduction of whatever type to `isGeneric` is undesirably fragile, though rethinking that flow is outside of the scope of fixing this immediate bug.

sideeffffect:
Thanks Devon!
:heart:

PR #1265, opened 2021-09-24
ci: add JDK 17

sullis:


kelnos (approved):
I think we should also remove 14 & 16 since those aren't really supported anymore?

sullis:
> I think we should also remove 14 & 16 since those aren't really supported anymore?

Done.

blast-hardcheese (approved):

blast-hardcheese:
Thanks!

PR #1293, opened 2021-10-20
Updating README with explicit module versions

blast-hardcheese:
Now that there are multiple modules, a single version is only a part of the picture. Restructure README to surface useful version numbers.


blast-hardcheese:
Skipping tests because it's just README

PR #1296, opened 2021-10-25
Issue-572 akka-http response headers

Peppi-Ressler:
implementation of https://github.com/guardrail-dev/guardrail/issues/572

**Contributing to Twilio**

- [x]  I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/.github/release-drafter-guardrail.yml
+++ b/.github/release-drafter-guardrail.yml
@@ -2,7 +2,7 @@ name-template: 'guardrail-v$RESOLVED_VERSION'
 tag-template: 'guardrail-v$RESOLVED_VERSION'
 tag-prefix: guardrail-v
 include-paths:
-  - "modules/codegen/"
+  - "modules/codegen/src/main/"
   - "build.sbt"
   - "project/src/main/scala/Build.scala"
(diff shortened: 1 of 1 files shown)


Veske:
```
/bar:
    get:
      operationId: getBar
      responses:
        '204':
          headers:
            X-Required-Header:
              type: string
              required: true
            X-Optional-Header:
              type: string
```

According to: https://swagger.io/docs/specification/describing-responses/
```
headers:
            X-RateLimit-Limit:
              schema:
                type: integer
              description: Request limit per hour.
            X-RateLimit-Remaining:
              schema:
                type: integer
              description: The number of requests left for the time window.
            X-RateLimit-Reset:
              schema:
                type: string
                format: date-time
              description: The UTC date/time at which the current rate limit window resets.
```

it seems like the spec expects the type to be nested under schema. Why was the implementation here made against the spec?

blast-hardcheese:
We are using the official OpenAPI parser, walking their domain objects, to ensure we're as close to the spec as we can be. Perhaps this is a bug upstream?

Alternately, we may be relying on a deprecated property and not realizing it, I suppose.

PR #1299, opened 2021-10-26
WIP: Move to Dropwizard 2.0 series

sdreynolds:
### Why:
1.3 branch was last released in 2020-12 making it pretty stale. Moving
off of this branch keeps us with the releases.

### How:
This should remove all references to dropwizard 1.3.X releases and
should add the specific dependencies required for this
project. Specifically, with Dropwizard 2.X, Guardrail needs to include

1. `hibernate-core`
2. `hibernate-validator`

as they are no longer transitively pulled into the project via
Dropwizard dependencies.

### Need Help:
I don't understand how the classpath for the tests is
constructed. This results in the classpath missing the `hibernate`
dependencies during test runs.


blast-hardcheese:
Alrighty, I've been sorta hands-off on this because I really don't know anything about this, but I was able to stumble through at least enough of the deprecation notes to get it to _attempt_ the tests.

I had no idea what whatever `Unwrapping` is, so I commented it out. Presumably we'll end up with runtime errors, but it gets us closer to iterating

martijnhoekstra:
A problem here is that dropwizard-scala is discontinued and has no dropwizard 2.x support

PR #1314, opened 2021-11-07
Explicit dependencies

blast-hardcheese:
So the experiment with using MiMa to ensure ABI compatibility across modules was unsuccessful, and the build must be stable, so for the time being just go directly to explicit dependencies and rely on scala-steward to bump.


blast-hardcheese:
Closing in favor of #1318. This may still be useful in the future, if that idea doesn't work, but starting there.

PR #1316, opened 2021-11-09
Move http4s response defs to local scope

zeal18:
The change repeats the same solution as in akka-http generator. The main problem is solves is to make it possible to generate both a server and a client to the same package. The best solution would be to reuse the same response definitions on both sides but it requires a wider change


blast-hardcheese:
Initial pass here looks great, I'll need a few more days to get back to this. Thank you for your contribution so far!

blast-hardcheese (approved):
This is great. I was initially concerned that `import ...Resource._` would bring in helpers or other static methods into scope, but considering this is a completely new object, that's perfect.

Thanks for shoring up this usability issue!

PR #1323, opened 2021-11-21
Add http4s 0.23.6 support

zeal18:
The change fixes regression tests filtering by framework: `scalaFrameworks` and `javaFrameworks` contained `projectName` before but `ExampleCase` required framework name instead. When `scalaFrameworks.toSet` was used, frameworks which have different name and projectName were filtered out. Considering that all `scalaFrameworks.toSet` usages were replaced by list of frameworks to prevent code generation where it's not required.

The http4s update change follows the suggestion from [the comment](https://github.com/guardrail-dev/guardrail/pull/1267#issuecomment-939404470). Also it copies http4s sample and tests for 0.22 to control regressions.

### DIFF
--- a/modules/codegen/src/main/scala/dev/guardrail/generators/scala/ScalaModule.scala
+++ b/modules/codegen/src/main/scala/dev/guardrail/generators/scala/ScalaModule.scala
@@ -8,7 +8,7 @@ import dev.guardrail.generators.scala.akkaHttp.{ AkkaHttpClientGenerator, AkkaHt
 import dev.guardrail.generators.scala.circe.CirceProtocolGenerator
 import dev.guardrail.generators.scala.dropwizard.{ DropwizardClientGenerator, DropwizardGenerator, DropwizardServerGenerator }
 import dev.guardrail.generators.scala.endpoints.{ EndpointsClientGenerator, EndpointsGenerator, EndpointsServerGenerator }
-import dev.guardrail.generators.scala.http4s.{ Http4sClientGenerator, Http4sGenerator, Http4sServerGenerator }
+import dev.guardrail.generators.scala.http4s.{ Http4sClientGenerator, Http4sGenerator, Http4sServerGenerator, Http4sVersion }
 import dev.guardrail.generators.scala.jackson.JacksonProtocolGenerator
 import dev.guardrail.terms.client.ClientTerms
 import dev.guardrail.terms.server.ServerTerms
@@ -70,13 +70,23 @@ object ScalaModule extends AbstractModule[ScalaLanguage] {
     EndpointsGenerator()
   )
 
+  def http4sV0_22(implicit Cl: CollectionsLibTerms[ScalaLanguage, Target]): (
+      ClientTerms[ScalaLanguage, Target],
+      ServerTerms[ScalaLanguage, Target],
+      FrameworkTerms[ScalaLanguage, Target]
+  ) = (
+    Http4sClientGenerator(),
+    Http4sServerGenerator(Http4sVersion.V0_22),
+    Http4sGenerator()
+  )
+
   def http4s(implicit Cl: CollectionsLibTerms[ScalaLanguage, Target]): (
       ClientTerms[ScalaLanguage, Target],
       ServerTerms[ScalaLanguage, Target],
       FrameworkTerms[ScalaLanguage, Target]
   ) = (
     Http4sClientGenerator(),
-    Http4sServerGenerator(),
+    Http4sServerGenerator(Http4sVersion.V0_23),
     Http4sGenerator()
   )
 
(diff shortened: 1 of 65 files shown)


blast-hardcheese:
Reviewing this now. Presumably if this builds and tests pass locally but not in CI here, it's because the build is attempting to use the released versions of inner modules. I'm still iterating on how to ensure we maintain bincompat moving forward after the modularization, so the failure to build here likely isn't a reflection of an actual problem.

blast-hardcheese:
> that all `scalaFrameworks.toSet` usages were replaced by list of frameworks to prevent code generation where it's not required.

The ability to compile at all is part of the regression tests, even if there are no runtime tests utilizing those -- it's to ensure that source compatibility doesn't drift with the underlying libraries

zeal18:
> > that all `scalaFrameworks.toSet` usages were replaced by list of frameworks to prevent code generation where it's not required.
> 
> The ability to compile at all is part of the regression tests, even if there are no runtime tests utilizing those -- it's to ensure that source compatibility doesn't drift with the underlying libraries

good point, returned usage of the `scalaFrameworks`

> Reviewing this now. Presumably if this builds and tests pass locally but not in CI here, it's because the build is attempting to use the released versions of inner modules. I'm still iterating on how to ensure we maintain bincompat moving forward after the modularization, so the failure to build here likely isn't a reflection of an actual problem.

how can I address the problem with the CI, maybe, split the PR?

blast-hardcheese:
> how can I address the problem with the CI, maybe, split the PR?

I think during CI, if the `major` or `minor` labels are applied to the PR, it should disable the ABI checking

blast-hardcheese:
In the meantime, I've run `runtimeScalaSuite` locally and it seems to fail with the same errors here, so those will need to be resolved before this can be merged. The errors from the most recent build look correct

zeal18:
... and it looks like I don't have permissions to set labels 😅

modules/codegen/src/main/scala/dev/guardrail/generators/scala/ScalaModule.scala
```diff
@@ -70,13 +70,23 @@ object ScalaModule extends AbstractModule[ScalaLanguage] {
...
       ServerTerms[ScalaLanguage, Target],
       FrameworkTerms[ScalaLanguage, Target]
   ) = (
     Http4sClientGenerator(),
-    Http4sServerGenerator(),
+    Http4sServerGenerator(Http4sVersion.V0_22),
+    Http4sGenerator()
+  )
+
+  def http4sV0_23(implicit Cl: CollectionsLibTerms[ScalaLanguage, Target]): (
```
blast-hardcheese, line 83:
```suggestion
  def http4s(implicit Cl: CollectionsLibTerms[ScalaLanguage, Target]): (
```

modules/codegen/src/main/scala/dev/guardrail/generators/scala/ScalaModule.scala
```diff
@@ -104,7 +114,9 @@ object ScalaModule extends AbstractModule[ScalaLanguage] {
         "framework",
         ("akka-http", catchClassNotFound(akkaHttp(modelGeneratorType), MissingDependency("guardrail-scala-akka-http"))),
         ("akka-http-v10.1", catchClassNotFound(akkaHttpV10_1(modelGeneratorType), MissingDependency("guardrail-scala-akka-http"))),
-        ("http4s", catchClassNotFound(http4s, MissingDependency("guardrail-scala-http4s"))),
+        ("http4s", catchClassNotFound(http4sV0_23, MissingDependency("guardrail-scala-http4s"))),
```
blast-hardcheese, line 117:
```suggestion
        ("http4s", catchClassNotFound(http4s, MissingDependency("guardrail-scala-http4s"))),
```

modules/codegen/src/main/scala/dev/guardrail/generators/scala/ScalaModule.scala
```diff
@@ -104,7 +114,9 @@ object ScalaModule extends AbstractModule[ScalaLanguage] {
...
+        ("http4s-v0.23", catchClassNotFound(http4sV0_23, MissingDependency("guardrail-scala-http4s"))),
```
blast-hardcheese, line 118:
```suggestion
        ("http4s-v0.23", catchClassNotFound(http4s, MissingDependency("guardrail-scala-http4s"))),
```

blast-hardcheese (approved):
One suggestion, though I can't do multi-change suggestions.

By keeping `def http4s` it keeps backwards ABI compat, as well as making the latest semantically the default.

zeal18:
Hi, @blast-hardcheese! I'm surprised a bit by the coverage report, could you assist finding a reason of such coverage falling?

modules/codegen/src/main/scala/dev/guardrail/generators/scala/ScalaModule.scala
zeal18: done

modules/codegen/src/main/scala/dev/guardrail/generators/scala/ScalaModule.scala
zeal18: done

modules/codegen/src/main/scala/dev/guardrail/generators/scala/ScalaModule.scala
zeal18: done

blast-hardcheese:
> Hi, @blast-hardcheese! I'm surprised a bit by the coverage report, could you assist finding a reason of such coverage falling?

codecov highlights the fact that for in `ScalaModule` we went from having `"http4s"` to `"http4s", "http4s-v0.22", "http4s-v0.23"` -- I'm not worried about such a small reduction in code coverage, don't worry 😁

Issue #545, opened 2020-02-14
Unable to access Circe's `dropNullValues` parameter


stjohnb:
I would like null values not to appear in my server's response json when they are an optional field containing a `None`. 

Circe enables this via the `dropNullValues` field in the [Printer](https://circe.github.io/circe/api/io/circe/Printer.html) class but I cannot see anyway to access this via guardrail

kelnos:
This isn't currently supported. We're aware that our handling of required/nullable isn't correct (https://github.com/twilio/guardrail/issues/315) but don't have an ETA on a fix.

stjohnb:
Ok, thanks for the info

PR #1339, opened 2021-12-11
Add possibility to provide custom circe printer for http4s

zeal18:
Also requested in #545

### DIFF
--- a/modules/sample-http4s-v0_22/src/test/scala/core/Http4s/Http4sCustomCircePrinterTest.scala
+++ b/modules/sample-http4s-v0_22/src/test/scala/core/Http4s/Http4sCustomCircePrinterTest.scala
@@ -1,0 +1,98 @@
+package core.Http4s
+
+import _root_.customCircePrinter.client.{ http4sV022 => cdefs }
+import _root_.customCircePrinter.server.http4sV022.users.{ UsersHandler, UsersResource }
+import _root_.customCircePrinter.server.http4sV022.users.UsersResource.{ GetUserResponse, PostUserResponse }
+import cats.effect.IO
+import customCircePrinter.client.http4sV022.users.UsersClient
+import customCircePrinter.server.http4sV022.definitions.{ User, UserAddress }
+import org.http4s.{ HttpRoutes, Request, Uri }
+import org.http4s.client.Client
+import org.http4s.client.Client
+import org.http4s.implicits._
+import org.http4s.dsl.io._
+import org.http4s.syntax.StringSyntax
+import org.scalatest.EitherValues
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+import io.circe.{ Json, Printer }
+
+class Http4sCustomCircePrinterTest extends AnyFunSuite with Matchers with EitherValues with StringSyntax {
+
+  test("server: prints nulls with default Printer") {
+    val server: HttpRoutes[IO] = new UsersResource[IO]().routes(new UsersHandler[IO] {
+      override def getUser(respond: GetUserResponse.type)(id: String): IO[GetUserResponse] =
+        IO.pure(GetUserResponse.Ok(User("id", UserAddress(None, Some("line 2"), None))))
+      override def postUser(respond: PostUserResponse.type)(body: User): IO[PostUserResponse] =
+        IO.pure(PostUserResponse.Ok(""))
+    })
+
+    val client = Client.fromHttpApp(server.orNotFound)
+
+    val retrieved: String =
+      client
+        .run(Request[IO](uri = Uri.unsafeFromString(s"/user/id")))
+        .use { _.bodyText.compile.string }
+        .attempt
+        .unsafeRunSync()
+        .value
+
+    retrieved shouldEqual """{"id":"id","address":{"line1":null,"line2":"line 2","line3":null}}"""
+  }
+
+  test("server: doesn't print nulls with configured Printer") {
+    val printer = Printer.noSpaces.copy(dropNullValues = true)
+
+    val server: HttpRoutes[IO] = new UsersResource[IO](printer = printer).routes(new UsersHandler[IO] {
+      override def getUser(respond: GetUserResponse.type)(id: String): IO[GetUserResponse] =
+        IO.pure(GetUserResponse.Ok(User("id", UserAddress(None, Some("line 2"), None))))
+      override def postUser(respond: PostUserResponse.type)(body: User): IO[PostUserResponse] =
+        IO.pure(PostUserResponse.Ok(""))
+    })
+
+    val client = Client.fromHttpApp(server.orNotFound)
+
+    val retrieved: String =
+      client
+        .run(Request[IO](uri = Uri.unsafeFromString(s"/user/id")))
+        .use { _.bodyText.compile.string }
+        .attempt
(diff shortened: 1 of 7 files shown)


blast-hardcheese:
What need are you trying to solve with a custom printer? I see you've got `dropNulls`, that's already provided by `propertyRequirement` -- is there some other benefit?

If you're in the root guardrail sbt, you can run:

```
cli --defaults --specPath issue1339.yaml --outputPath /tmp/http4s/foop/modules/sample/target/generated --module circe --module http4s --client --packageName issues.issue1339.client.http4s --server --packageName issues.issue1339.server.http4s
```

and compare the state between adding ` --optional-encode-as optional --optional-decode-as optional`:
```
cli --defaults --specPath issue1339.yaml --outputPath /tmp/http4s/foop/modules/sample/target/generated --module circe --module http4s --optional-encode-as optional --optional-decode-as optional --client --packageName issues.issue1339.client.http4s --server --packageName issues.issue1339.server.http4s
```

... with `issue1339.yaml` as:
```yaml
swagger: "2.0"
host: localhost:1234
schemes:
  - http
definitions:
  Baz:
    type: string
    enum:
      - beep
      - borp
      - blap
  Structure:
    type: object
    properties:
      foo:
        type: string
      bar:
        type: string
paths:
  "/foo":
    get:
      operationId: doFoo
      produces:
        - application/json
      parameters:
        - name: bar
          in: query
          type: array
          items:
            $ref: "#/definitions/Baz"
        - name: baz
          in: body
          schema:
            $ref: "#/definitions/Structure"
      responses:
        200:
          description: OK
```

I got this change between the two invocations:

```diff
diff --git a/modules/sample/target/generated/issues/issue1339/client/http4s/definitions/Structure.scala b/modules/sample/target/generated/issues/issue1339/client/http4s/definitions/Structure.scala
index 1f3a0f0..1d20cdf 100644
--- a/modules/sample/target/generated/issues/issue1339/client/http4s/definitions/Structure.scala
+++ b/modules/sample/target/generated/issues/issue1339/client/http4s/definitions/Structure.scala
@@ -8,11 +8,17 @@ import io.circe._
 import io.circe.syntax._
 import cats.implicits._
 import _root_.issues.issue1339.client.http4s.Implicits._
 case class Structure(
-  foo: Option[String] = None,
+  foo: _root_.issues.issue1339.client.http4s.support.Presence[String] = _root_.issues.issue1339.client.http4s.support.Presence.Absent,
-  bar: Option[String] = None
+  bar: _root_.issues.issue1339.client.http4s.support.Presence[String] = _root_.issues.issue1339.client.http4s.support.Presence.Absent
 )
 object Structure {
   implicit val encodeStructure: _root_.io.circe.Encoder.AsObject[Structure] = {
     val readOnlyKeys = _root_.scala.Predef.Set[_root_.scala.Predef.String]()
     _root_.io.circe.Encoder.AsObject.instance[Structure](a =>
       _root_.io.circe.JsonObject.fromIterable(
-        _root_.scala.Vector(("foo", a.foo.asJson), ("bar", a.bar.asJson))
+        _root_.scala.Vector() ++
+        a.foo.fold(ifAbsent = None, ifPresent = value => Some("foo" -> value.asJson)) ++
+        a.bar.fold(ifAbsent = None, ifPresent = value => Some("bar" -> value.asJson))
       )
     ).mapJsonObject(_.filterKeys(key => !(readOnlyKeys contains key)))
   implicit val decodeStructure: _root_.io.circe.Decoder[Structure] = new _root_.io.circe.Decoder[Structure] {
     final def apply(c: _root_.io.circe.HCursor): _root_.io.circe.Decoder.Result[Structure] = for {
-      v0 <- c.downField("foo").as[Option[String]]
+      v0 <- ((c: _root_.io.circe.HCursor) =>
+        c.value
+          .asObject
+          .filter(!_.contains("foo"))
+          .fold(c.downField("foo").as[String].map(x => _root_.issues.issue1339.client.http4s.support.Presence.present(x))) { _ => _root_.scala.Right(_root_.issues.issue1339.client.http4s.support.Presence.absent) }
+      )(c)
-      v1 <- c.downField("bar").as[Option[String]]
+      v1 <- ((c: _root_.io.circe.HCursor) =>
+        c.value
+          .asObject
+          .filter(!_.contains("bar"))
+          .fold(c.downField("bar").as[String].map(x => _root_.issues.issue1339.client.http4s.support.Presence.present(x))) { _ => _root_.scala.Right(_root_.issues.issue1339.client.http4s.support.Presence.absent) }
+      )(c)
     } yield Structure(v0, v1)
   }
 }
```

zeal18:
Hi @blast-hardcheese, thank you for your reply! 
I have the following requirements:
* specification file is 3rdparty and shouldn't be changed
* the server is 3rdparty too (generated with guardrail or not, I don't know) and it fails with 400 when {"key": null} is supplied

By default circe generates nulls and I want to disable it by providing a custom `Printer` (default solution for circe). Suggested solution introduces tristate models and I would like to avoid such models complication

blast-hardcheese:
Makes sense; I believe there's an "empty is absent" option for the nullable encoder flag, though I can't look it up until later. I think it should be there though, if not we should add it.

The guiding principle here is that if something is already part of the OAI specification, we should prefer that to bespoke extensions, as it will improve the quality of generated code in the long run.

blast-hardcheese:
@zeal18 With the same example as before, it looks like what you want is:

```
cli --defaults --specPath issue1339.yaml --outputPath /tmp/http4s/foop/modules/sample/target/generated --module circe --module http4s --optional-encode-as optional --optional-decode-as legacy --client --packageName issues.issue1339.client.http4s --server --packageName issues.issue1339.server.http4s
```

generates:
```diff
   implicit val encodeStructure: _root_.io.circe.Encoder.AsObject[Structure] = {
     val readOnlyKeys = _root_.scala.Predef.Set[_root_.scala.Predef.String]()
     _root_.io.circe.Encoder.AsObject.instance[Structure](a =>
       _root_.io.circe.JsonObject.fromIterable(
         _root_.scala.Vector() ++
-        a.foo.fold(ifAbsent = None, ifPresent = value => Some("foo" -> value.asJson)) ++
-        a.bar.fold(ifAbsent = None, ifPresent = value => Some("bar" -> value.asJson))
+        a.foo.map(value => ("foo", value.asJson)) ++
+        a.bar.map(value => ("bar", value.asJson))
       )
     ).mapJsonObject(_.filterKeys(key => !(readOnlyKeys contains key)))
   }
```

which is fine, though
```
[info] --optional-encode-as optional was used, which does not match value of --optional-decode-as legacy. This will result in the use of `Option[T]` as opposed to regular `Presence[T]`.
[info] --optional-encode-as optional was used, which does not match value of --optional-decode-as legacy. This will result in the use of `Option[T]` as opposed to regular `Presence[T]`.
```
is logged, which is misleading.

blast-hardcheese:
```scala
  ScalaClient(file("....yaml"), pkg="com.example.clients", 
              encodeOptionalAs = codingOptional,
              decodeOptionalAs = codingOptionalLegacy),
```
if you're using the sbt plugin

zeal18:
Thank you for the explanation! It works as expected, I think the PR could be closed

blast-hardcheese:
I think removing the warning is probably good, as well as clarifying why "legacy" is legacy, as well as ensuring there's a config value where the tri-state can be avoided.

Thank you for confirming!

PR #1342, opened 2021-12-15
Add basic authentication support for http4s

zeal18:
This is an idea, how authentication could be implemented. It's enabled automatically based on presents of global `securitySchemes` and at least one `security` requirement for a handler's route. Check `sample-http4s/authentication` for an example of generated code and `Http4sAuthenticationTest` for a possible usage example


blast-hardcheese:
Please pardon the initial silence here -- I didn't have the time this weekend, I'll try to get to it sometime this week

blast-hardcheese (dismissed):
If we are to attempt to actually implement authentication in a general way, I think it is important that we follow the spec, as avoiding generated API churn is an explicit goal of this project.

While you are correct, supplying an additional type parameter is likely the correct direction to go, the similarities between `authenticationMiddleware` and `mapRequest` is so great, it is actually possible to implement this feature [without a change to guardrail at all](https://gist.github.com/blast-hardcheese/ce10c703381e2d7397c3e6091397628b) by changing the supplied `F` from `IO` to `Kleisli[IO, Option[AuthContext], *]`.

As I commented elsewhere in this review, I think the primary concerns that need to be addressed are:
- Instead of enabling/disabling authentication via `Boolean`, we need to explicitly thread through which kind of authentication is supplied, per route, with failover
- If at all possible, source compatibility with http4s' [existing authentication middleware](https://http4s.org/v0.23/auth/) would be very useful, as it would permit code reuse between existing and new implementations, as well as compatibility with already-published http4s middleware providers.
  - If this also means switching from `HttpRoutes.of`... to `AuthedRoutes.of` and composing the two implementations together via `<+>`, that may help keep reimplementation down, and utilize some of the infrastructure already provided to us by http4s.

Thank you for your initial effort here, it's great to see interest for first-class authentication support in guardrail, at Twilio this was implemented in a bespoke way for our Dropwizard generators, targeting internal Twilio authentication provider libraries. What are your thoughts to my above observations?

zeal18:
> If we are to attempt to actually implement authentication in a general way, I think it is important that we follow the spec, as avoiding generated API churn is an explicit goal of this project.
> 
> While you are correct, supplying an additional type parameter is likely the correct direction to go, the similarities between `authenticationMiddleware` and `mapRequest` is so great, it is actually possible to implement this feature [without a change to guardrail at all](https://gist.github.com/blast-hardcheese/ce10c703381e2d7397c3e6091397628b) by changing the supplied `F` from `IO` to `Kleisli[IO, Option[AuthContext], *]`.
> 
> As I commented elsewhere in this review, I think the primary concerns that need to be addressed are:
> 
> * Instead of enabling/disabling authentication via `Boolean`, we need to explicitly thread through which kind of authentication is supplied, per route, with failover
> * If at all possible, source compatibility with http4s' [existing authentication middleware](https://http4s.org/v0.23/auth/) would be very useful, as it would permit code reuse between existing and new implementations, as well as compatibility with already-published http4s middleware providers.
>   
>   * If this also means switching from `HttpRoutes.of`... to `AuthedRoutes.of` and composing the two implementations together via `<+>`, that may help keep reimplementation down, and utilize some of the infrastructure already provided to us by http4s.
> 
> Thank you for your initial effort here, it's great to see interest for first-class authentication support in guardrail, at Twilio this was implemented in a bespoke way for our Dropwizard generators, targeting internal Twilio authentication provider libraries. What are your thoughts to my above observations?

Thank you for the review, I‘ll come back after the holidays

modules/scala-http4s/src/main/scala/dev/guardrail/generators/scala/http4s/Http4sServerGenerator.scala
```diff
@@ -55,6 +55,7 @@ class Http4sServerGenerator private (version: Http4sVersion)(implicit Cl: Collec
   private def this(Cl: CollectionsLibTerms[ScalaLanguage, Target]) = this(Http4sVersion.V0_23)(Cl)
 
   val customExtractionTypeName: Type.Name = Type.Name("E")
+  val authContextTypeName: Type.Name      = Type.Name("AuthContextT")
```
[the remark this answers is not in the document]
zeal18: it's definitely not a monad transformer, just an old OOP suffix notation popped up in my mind, will rename

modules/core/src/main/scala/dev/guardrail/generators/ServerGenerator.scala
```diff
@@ -50,6 +50,10 @@ object ServerGenerator {
           for {
             resourceName <- formatTypeName(className.lastOption.getOrElse(""), Some("Resource"))
             handlerName  <- formatTypeName(className.lastOption.getOrElse(""), Some("Handler"))
+            authenticationEnabled = securitySchemes.nonEmpty && routes.exists {
```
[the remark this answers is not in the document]
zeal18: I think, it's not necessary to have `securitySchemes` down in the routes class, because there is no guarantee that all of them or any of them are used in routes. A route in the OAS contains security schemes names it requires, this information should be enough (in case we don't care about the schemes types and give a user the responsibility to implement them)

zeal18:
> If we are to attempt to actually implement authentication in a general way, I think it is important that we follow the spec, as avoiding generated API churn is an explicit goal of this project.
> 
> While you are correct, supplying an additional type parameter is likely the correct direction to go, the similarities between `authenticationMiddleware` and `mapRequest` is so great, it is actually possible to implement this feature [without a change to guardrail at all](https://gist.github.com/blast-hardcheese/ce10c703381e2d7397c3e6091397628b) by changing the supplied `F` from `IO` to `Kleisli[IO, Option[AuthContext], *]`.

yes, it is similar, but it's not the same. `mapRequest` applies to all routes and to distinguish them a user needs to use `String` argument which a source of bugs. Using `Option[AuthContext]` doesn't guarantee a user has `Some` on routes, which were authenticated and has the context, this could lead having `.get` in user code (leaks authentication info to user code, but we should keep it on our side). Moreover, `mapRequest` doesn't have auth kinds and scopes information. 
> 
> As I commented elsewhere in this review, I think the primary concerns that need to be addressed are:
> 
> * Instead of enabling/disabling authentication via `Boolean`, we need to explicitly thread through which kind of authentication is supplied, per route, with failover

I agree, but we don't need to use `securitySchemes` for that, because any OAS route contains a list of auth kinds with scopes.

> * If at all possible, source compatibility with http4s' [existing authentication middleware](https://http4s.org/v0.23/auth/) would be very useful, as it would permit code reuse between existing and new implementations, as well as compatibility with already-published http4s middleware providers.

I think it's possible to reuse, but it's better not to, because the http4s auth works on completely another level and doesn't have an information about auth kinds and scopes. Probably it could be implemented grouping routes but routes could have several kinds of auth at the same time. It would be hard to group such routes:
* route A with auth kinds [basic]
* route B with auth kinds [basic, oauth]
* route C with auth kinds [oauth]
we could have different groups for routes A and C to authenticate them, but it's very hacky to wrap route B correctly reusing middlewares for basic and oauth, the only way I see is to implement another middleware which combines them special way.

I've got an idea while writing this comment, I'll try to experiment with it. ☺️

>   * If this also means switching from `HttpRoutes.of`... to `AuthedRoutes.of` and composing the two implementations together via `<+>`, that may help keep reimplementation down, and utilize some of the infrastructure already provided to us by http4s.

I wouldn't move it to a user code, maybe we can do it internally if it helps

> Thank you for your initial effort here, it's great to see interest for first-class authentication support in guardrail, at Twilio this was implemented in a bespoke way for our Dropwizard generators, targeting internal Twilio authentication provider libraries. What are your thoughts to my above observations?

Thank you for the review, I'll try to push it forward whenever I have time

blast-hardcheese:
> Using `Option[AuthContext]` doesn't guarantee a user has `Some` on routes, which were authenticated and has the context, this could lead having `.get` in user code

When I've done this in the past, I've done something along the lines of

```scala
ctx.fold(respond.NotAuthorized) { auth =>
  respond.Ok(auth.username)
}
```
though I appreciate the observation that this isn't always obvious to the user. It does raise a consideration though, which is that if the route only specifies `200`/`404`, presumably auth failures would introduce an otherwise unspecified response code with an unspecified schema. Perhaps a `UserError` if we intend to fail with an error code that isn't mentioned in the specification (though that doesn't solve the `body` problem, I think it's a step in the right direction).

> I agree, but we don't need to use `securitySchemes` for that, because any OAS route contains a list of auth kinds with scopes.

Good -- So long as we agree that `Boolean` is suboptimal, I don't have an opinion on how it is implemented, so long as it's implemented in a flexible and clear way. Looking at the spec, it seemed as though there was at best a loose correlation between the defined schemes at the root of the specification and the usage of those schemes in individual routes.

One of the objections I have to `Boolean` is that it removes our ability to verify and offer detailed error messages if we see something that doesn't make sense, to guide the user towards a resolution to their issue. If we agree, and you take that into account in your implementation, then I'm comfortable moving forward with it 😁

> I wouldn't move it to a user code, maybe we can do it internally if it helps

I agree with this -- any careless modification could change fully functional production code to only partially implemented code without any compiler errors, which would be a huge detriment to users. Thanks for calling it out!

zeal18:
Hi again! I've found some time to get back. 
I've experimented a bit with http4s's middlewares without success. It's made for another use case and doesn't fit into OAS definition of security handling. The main problem is that OAS allows to define pretty complex authentication checks like:
```
security:    # (A AND B) OR (C AND D)
  - A
    B
  - C
    D
```
Another potential problem here is caching, for example in case `(A AND B) OR (A AND C)` we probably want to cache `A` result and, if `B` fails, reuse it to continue evaluating directly to `C`. Also http4s's middlewares hardcode response types and don't expose their decision to user code which makes inconsistency with the scheme (as you pointed on your comment).
So, in the latest implementation all decisions should be made on the used code side: it decides what to do with security schemes and scopes, implements caching and whatever optimisations make sense (like always evaluate `basic` before `oauth`). As a result it returns an Option (maybe we should replace it by Either to be able to provide more details about an error) which is passed directly to the handler and should be proceed on the used code side to chose correct response type and generate valid response body.

blast-hardcheese (commented):
This looks really good. I did a few stylistic tweaks:

- avoiding `unsafe*` methods
- preferring `A => A` function composition instead of maintaining structurally similar branches
- reducing `case (Some(_), Some(_))` into a single `Option[(A, B)]`

The one sticking point for me is the ergonomics around `authMiddleware` itself. Looking at a fairly common case of being able to answer a challenge via one or more `A => IO[B]` functions, as a user I'd expect to be able to (straightforwardly):

- Attempt to execute auth challenges in parallel
- Collect and enforce the "and" and "or" constraints

I'm sure I've gotten lost somewhere here, but I tried to do a mock implementation that would match the type signature. The following is what I ended up with:

```scala
  def lookupAuthedScopes(req: Request[IO], scopes: List[String]): IO[List[String]] =
    IO.pure(
      req.headers.get(CIString("dummy-scopes"))
        .map(_.head.value.split(",").intersect(scopes))
        .toList
        .flatten
    )

  def authMiddleware(schemes: NonEmptyList[NonEmptyMap[Resource.AuthSchemes, List[String]]], req: Request[IO]): IO[Option[List[String]]] = {
    println("Schemes:")
    schemes.toList.foreach(_.toSortedMap.foreach({ case (scheme, scopes) => println(s"  ${scheme.name} -> ${scopes.mkString(", ")}") }))
    schemes.collectFirstSomeM[IO, List[String]](_.toSortedMap.toList.flatTraverse({
      case (Resource.AuthSchemes.Basic1, scopes) =>
        lookupAuthedScopes(req, scopes)
      case (Resource.AuthSchemes.Basic2, scopes) =>
        lookupAuthedScopes(req, scopes)
    }).map(Option(_).filter(_.nonEmpty)))
  }
```

This additionally relies on the implementation agreeing with what the specification says in `#!/components/securitySchemes`, everything's kinda open-ended in that regard.

Do you have any guidance of how you were intending these functions be implemented? The tests are extremely narrow and immediately move away from the `NonEmptyList`/`NonEmptyMap` types, in favor of direct equality against known test values, which are not sufficient inspiration for how a user may adapt this to their environment.

blast-hardcheese:
Playing around a little bit more, in my example I've discovered that

```scala
schemes.collectFirstSomeM(...)
```
and
```scala
lookupAuthedScopes(...)
```

_almost_ mesh with the type of `AuthMiddleware` from http4s, which would be a good entrypoint for reuse of existing libraries, or for users trying to migrate existing codebases.

If `Kleisli[OptionT[F, *], Request[F], A]` can be escorted carefully through `NonEmptyList[NonEmptyMap[Scheme, Scopes]]`, there may be a good amount of code reuse and interoperation with the ecosystem there.

zeal18:
> This looks really good. I did a few stylistic tweaks:
> 
> * avoiding `unsafe*` methods
> * preferring `A => A` function composition instead of maintaining structurally similar branches
> * reducing `case (Some(_), Some(_))` into a single `Option[(A, B)]`
> 
> The one sticking point for me is the ergonomics around `authMiddleware` itself. Looking at a fairly common case of being able to answer a challenge via one or more `A => IO[B]` functions, as a user I'd expect to be able to (straightforwardly):
> 
> * Attempt to execute auth challenges in parallel
> * Collect and enforce the "and" and "or" constraints

Current implementation moves responsibility of processing "and" and "or" to the user side which increases complexity and reduces ergonomics, but from the other hand it gives user more freedom to optimise the check. Do you think it would be better to move parallelisation and running logic to the guardrail side and only request from the user a bunch of authentication functions (`(List[String], Request[IO]) => IO[Option[B]]`)?

> I'm sure I've gotten lost somewhere here, but I tried to do a mock implementation that would match the type signature. The following is what I ended up with:
> 
> ```scala
>   def lookupAuthedScopes(req: Request[IO], scopes: List[String]): IO[List[String]] =
>     IO.pure(
>       req.headers.get(CIString("dummy-scopes"))
>         .map(_.head.value.split(",").intersect(scopes))
>         .toList
>         .flatten
>     )
> 
>   def authMiddleware(schemes: NonEmptyList[NonEmptyMap[Resource.AuthSchemes, List[String]]], req: Request[IO]): IO[Option[List[String]]] = {
>     println("Schemes:")
>     schemes.toList.foreach(_.toSortedMap.foreach({ case (scheme, scopes) => println(s"  ${scheme.name} -> ${scopes.mkString(", ")}") }))
>     schemes.collectFirstSomeM[IO, List[String]](_.toSortedMap.toList.flatTraverse({
>       case (Resource.AuthSchemes.Basic1, scopes) =>
>         lookupAuthedScopes(req, scopes)
>       case (Resource.AuthSchemes.Basic2, scopes) =>
>         lookupAuthedScopes(req, scopes)
>     }).map(Option(_).filter(_.nonEmpty)))
>   }
> ```
> 
> This additionally relies on the implementation agreeing with what the specification says in `#!/components/securitySchemes`, everything's kinda open-ended in that regard.
> 
> Do you have any guidance of how you were intending these functions be implemented? The tests are extremely narrow and immediately move away from the `NonEmptyList`/`NonEmptyMap` types, in favor of direct equality against known test values, which are not sufficient inspiration for how a user may adapt this to their environment.

I've reimplemented your example to how I see it could work in a real case:

```scala
    final case class User(id: Long, name: String, scopes: List[String])
    def getUserFromDB(id: Long, password: String): IO[Option[User]] =
      IO.pure(Some(User(id, "John Smith", List("admin"))))

    def base64Decode(in: String): Option[String] = ???

    def extractCredentials(req: Request[IO]): Option[(Long, String)] =
      req.headers
        .get[Authorization]
        .collect { case Authorization(Credentials.Token(AuthScheme.Basic, token)) => token }
        .flatMap(base64Decode)
        .map(_.split(":").toList)
        .collect { case id :: password :: Nil => (id.toLong, password) } // unsafe throw

    def authMiddleware(schemes: NonEmptyList[NonEmptyMap[Resource.AuthSchemes, List[String]]], req: Request[IO]): IO[Option[User]] = {
      println("Schemes:")
      schemes.toList.foreach(_.toSortedMap.foreach({ case (scheme, scopes) => println(s"  ${scheme.name} -> ${scopes.mkString(", ")}") }))
      schemes
        .collectFirstSomeM[IO, User](
          _.toSortedMap.toList
            .map {
              case (Resource.AuthSchemes.Basic1, scopes) =>
                extractCredentials(req).flatTraverse { case (id, pass) => getUserFromDB(id, pass).map(_.filter(_.scopes.toSet == scopes.toSet)) }
              case (Resource.AuthSchemes.Basic2, scopes) =>
                extractCredentials(req).flatTraverse { case (id, pass) => getUserFromDB(id, pass).map(_.filter(_.scopes.toSet == scopes.toSet)) }
            }
            .reduce[IO[Option[User]]] {
              case (l, r) =>
                l.flatMap {
                  case Some(_)  => r
                  case None       => IO.pure(None)
                }
            }
        )
    }
```
Some notes after writing it:
* it doesn't support parallelisation
* it doesn't support caching: it will invoke `getUserFromDB` twice
* it doesn't check that in the first check user id is the same as in the second (but in a real case different checks could use different user sources with different id's)
* I think `scopes` should be `Set` instead of `List`
* I think we should return `Either[AuthError, A]` instead of `Option[A]` to be able to distinguish different errors on the handler (in the user's code) and map them to different responses. It would be also helpful for logging and debugging. If it's redundant for a user, it's always easy to convert an `Either` to an `Option` on the user side.

While I wrote this I've realised that the implementation looks a bit complicated, especially when a real use case has only one authentication type. I would suggest to introduce a config parameter like "custom authentication" which would enable current 'complicated' implementation for a complicated case (with "and"s and "or"s, parallelisation requirement and etc.). But replace the current implementation by simplified (from user point of view) one: the code you wrote in the `authMiddleware` we can execute on the guardrail side and require from a user only a list of functions `(List[String], Request[IO]) => IO[Option[B]]` depending how many authentication types are defined in a spec.

Even in the simplified case the http4s `AuthMiddleware` wouldn't be applicable because it cuts out the user code returning hardcoded error responses directly from it, which makes it harder to customise bodies and could create an inconsistency with a spec.

blast-hardcheese:
> Current implementation moves responsibility of processing "and" and "or" to the user side which increases complexity and reduces ergonomics, but from the other hand it gives user more freedom to optimise the check. Do you think it would be better [...]

> to move parallelisation and running logic to the guardrail side [...]

Enforcing parallelized calls against some user code seems like it would be easier for the user to implement, which is good, at the expense of some (possibly nonexistent?) case where a user _does not want_ that parallelism.

> and only request from the user a bunch of authentication functions ((List[String], Request[IO]) => IO[Option[B]])?

I wonder if a better strategy wouldn't be to offer some `guardrail-auth` library that conforms to the expected type you've specified here, with adapters to known-working implementations of various authentication schemes (I really like [bare-bones-digest](https://github.com/al-broco/bare-bones-digest), as an example. Zero dependency digest auth implementation).

---

As for ways to improve, I'm interested in your thought about what the user experience of `Either[E, A]` instead of `Option[A]` for the auth context would look like, I think it's also useful to communicate `No Auth` vs `Failed Auth` states.

I was experimenting with whether I liked the idea of reducing the `NEL[NEM[Scheme, Scopes]] => F[A]` to just `(Scheme, Scopes) => F[A]`, and while it does make the code simpler (see the example at the end of this post), it would make `guardrail-auth` much more cumbersome to manage for the user (choosing different objects or methods that would otherwise be nearly identical but for the number of elements in a list in a specification file).

Thank you for the dialogue around what end-user code would look like, your implementation clears up a fair amount.

---

Idea for what the singleton scheme code may look like, though to reiterate, I don't believe this is worth the cost of an explosion of complexity in the to-be-designed `guardrail-auth` library.

```scala
    final case class User(id: Long, name: String, scopes: Set[String])
    def getUserFromDB(id: Long, password: String): IO[Option[User]] =
      IO.pure(Some(User(id, "John Smith", Set("admin"))))

    def base64Decode(in: String): Option[String] = ???

    def extractCredentials(req: Request[IO]): Option[(Long, String)] =
      req.headers
        .get[Authorization]
        .collect { case Authorization(Credentials.Token(AuthScheme.Basic, token)) => token }
        .flatMap(base64Decode)
        .map(_.split(":").toList)
        .collect { case id :: password :: Nil => (id.toLong, password) } // unsafe throw

    val authMiddleware: ((Resource.AuthSchemes, Set[String]), Request[IO]) => IO[Option[User]] = {
      case ((Resource.AuthSchemes.Basic, scopes), req) =>
        extractCredentials(req)
          .flatTraverse { case (id, pass) =>
            getUserFromDB(id, pass)
              .map(_.filter(u => scopes.diff(u.scopes).isEmpty)) // If the user contains at least the scopes requested
          }
    }
```

zeal18:
I like the latest code example, I think it has enough simplicity and we should consider it as the main solution. For users who have special cases we can provide `NEL[NEM[Scheme, Scopes]] => F[A]` option through config. Also I'm thinking about a possibility to completely switch off the authentication on the guardrail side (at least to be compatible with the old implementations and don't force users to migrate). It's a good idea to provide a `guardrail-auth` library, but I think it should be an addition not a mandatory part.
The next steps I'll apply:
* replace `Option[A]` result by `Either[E, A]`
* introduce a config parameter with `disable`, `simple` and `custom` options considering the current implementation (with NEL and NEM) as a `custom`
* implement `simple` one based on the latest code example

blast-hardcheese:
I just merged #1359, which will likely resolve the code coverage issue you're running into in this PR. I'm fine ignoring it for now, since we've been dancing around our coverage target for a while, pushed into the red by the http4s 0.22 and 0.23 tests being split in two until we drop 0.22 support. Feel free to rebase off master if you want to see green, there should be no merge conflicts.

modules/scala-http4s/src/main/scala/dev/guardrail/generators/scala/http4s/Http4sServerGenerator.scala
```diff
@@ -206,18 +206,16 @@ class Http4sServerGenerator private (version: Http4sVersion)(implicit Cl: Collec
...
-              }).reduce[F[Either[$authErrorTypeName, $authContextTypeName]]]({
-                case (l, r) =>
-                  l.flatMap({
-                    case Right(_) =>
-                      r
-                    case l: Left[$authErrorTypeName, $authContextTypeName] =>
-                      Applicative[F].pure(l)
-                  })
-              })
+              val nel = el.toNel
```
blast-hardcheese, line 209: One caveat of the `NonEmptyMap` is that it seems to preclude the ["Optional Oauth2 Security"](https://swagger.io/specification/#security-requirement-object) example, reproduced here:

```yaml
security:
  - {}
  - petstore_auth:
    - write:pets
    - read:pets
```

This looks like an oversight from the first foray into authentication, so if you'd like to consider that case out of scope here I'm happy to take that on, if you're running out of time. I appreciate the number of iterations you've taken to this problem so far, just trying to make sure we merge something that isn't going to need more work immediately.

zeal18: that's true, I've wrote about the bug above. I would fix it before merging the change because it's not very useful with the problem

zeal18: ah, I've checked the link and it's not the bug I mentioned, it's another use case related with optional requirements, anyway, I think it needs to be fixed)

zeal18:
Hi again, I'm still here)
It seems I've finalised the whole change:
* `Option` result has been replaced by `Either`, but only for `simple` authentication. For `custom` authentication we just do nothing 😜 because we move the responsibility to a user and the user could define any model he/she wants through `AuthContext`, we just pass it from an authentication middleware to a handler.
* `x-security-optional` type has been changed to boolean (it wasn't used anyway) to use it as a workaround for the ["Optional Oauth2 Security"](https://swagger.io/specification/#security-requirement-object). The underlying swagger library we use doesn't support it and the `x-security-optional` comes to help
* also fixed a possibility to disable global security requirements for particular paths

modules/scala-http4s/src/main/scala/dev/guardrail/generators/scala/http4s/Http4sServerGenerator.scala
zeal18: as I wrote in the previous comment, the swagger library we use under the hood doesn't support this approach to define optional authentication. It seems there is the `x-security-optional` to cover the case, but it was a list of names which is a bit overwhelming and unclear. To make it a bit more clear, I've changed its type to Boolean and it could be defined just for a route to make its security requirements optional.
For the `custom` authentication we provide its value to the `authenticationMiddleware` and let the user to decide what to do with that: produce some special case of its `AuthContext` or just pass it through inside of the `AuthContext` to the handler and deal with that there.
For the `simple` authentication we replace `Either` by an `Option` for the handler but the `authenticationMiddleware`  signature wasn't changed

blast-hardcheese:
> The underlying swagger library we use doesn't support it

Before moving forward with the review, I did a small experiment with:

```diff
diff --git a/modules/sample/src/main/resources/issues/issue1218.yaml b/modules/sample/src/main/resources/issues/issue1218.yaml
index 08a65de4..05751565 100644
--- a/modules/sample/src/main/resources/issues/issue1218.yaml
+++ b/modules/sample/src/main/resources/issues/issue1218.yaml
@@ -11,6 +11,10 @@ paths:
       responses:
         '204':
           description: No response
+      security:
+        - basic:
+          - "bar:basic"
+        - {}
       parameters:
         - name: refvec
           in: query
@@ -125,6 +129,10 @@ paths:
               type: integer
               format: int64
 components:
+  securitySchemes:
+    basic:
+      type: http
+      scheme: basic
   schemas:
     refvec:
       type: array
```

combined with

```diff
diff --git a/modules/core/src/main/scala/dev/guardrail/Common.scala b/modules/core/src/main/scala/dev/guardrail/Common.scala
index 47b5c90b..6445a953 100644
--- a/modules/core/src/main/scala/dev/guardrail/Common.scala
+++ b/modules/core/src/main/scala/dev/guardrail/Common.scala
@@ -72,6 +72,15 @@ object Common {
         .filter(_ != "/")

       paths = swagger.downField("paths", _.getPaths)
+      () = {
+        import scala.jdk.CollectionConverters._
+        swagger.unwrapTracker.getPaths.asScala.values.toList.flatMap(_.readOperations().asScala.toList).flatMap(_.getSecurity.asScala.toList).foreach(println)
+      }
       globalSecurityRequirements = NonEmptyList
         .fromList(swagger.downField("security", _.getSecurity).unwrapTracker)
         .flatMap(SecurityRequirements(_, SecurityOptional(swagger), SecurityRequirements.Global))
```

which, while not null safe at all and quite unreadable, it does show the expected states:

```
[info] class SecurityRequirement {
[info]     {basic=[bar:basic]}
[info] }
[info] class SecurityRequirement {
[info]     {}
[info] }
```

What is not supported? Is this some interaction with the automatic type-conversion in `Tracker`? If so, I can help -- that structure is still somewhat difficult to work with.

Since the case of `- {}` is actually in the documentation for the OpenAPI specification itself, I'd really like to get that case working, instead of adding a workaround that extends the specification.

Additionally, if `x-security-optional` is currently not used today (perhaps it was used for something internally at Twilio in the past, unsure), we can likely drop it, simplifying the user experience.

zeal18:
oh, I definitely misunderstood some concepts, thanks for noticing. I replaced the hack with `x-security-optional` to normal optional security requirements definition and deleted the `x-security-optional` flag support. Other changes are left the same

modules/scala-http4s/src/main/scala/dev/guardrail/generators/scala/http4s/Http4sServerGenerator.scala
```diff
@@ -739,13 +739,11 @@ class Http4sServerGenerator private (version: Http4sVersion)(implicit Cl: Collec
...
-            securityRequirements.map(
+            securityRequirements.map({
               sr =>
-                if (sr.optional)
-                  (LanguageParameter.fromParam(param"authContext: Option[$authContextTypeName]"), sr)
-                else
-                  (LanguageParameter.fromParam(param"authContext: Either[${Term.Name(resourceName)}.$authErrorTypeName, $authContextTypeName]"), sr)
-            )
+                val inner = if (sr.optional) t"Option[$authContextTypeName]" else authContextTypeName
+                (LanguageParameter.fromParam(param"authContext: Either[${Term.Name(resourceName)}.$authErrorTypeName, $inner]"), sr)
```
zeal18: I would change the type to `Either[Forbidden, Option[AuthContext]]` to eliminate the inconsistent state

blast-hardcheese:
Just force pushed to fix the last comment. Provided the tests pass, I'll release the branch back to you

zeal18:
Hi! It's been a while... Finally I've found some time to implement the 'native' authentication for http4s. It ignores OpenAPI spec security definitions and create `AuthedRoutes` to be provided to the `AuthMiddleware`.

blast-hardcheese:
I've tested `native` and `simple` in a test project, this looks excellent.

Thanks so much for seeing this through to the end! I'm convinced on functionality, I'll do a few more passes to feel comfortable I didn't miss anything and merge.

modules/scala-http4s/src/main/scala/dev/guardrail/generators/scala/http4s/Http4sServerGenerator.scala
```diff
@@ -608,7 +762,62 @@ class Http4sServerGenerator private (version: Http4sVersion)(implicit Cl: Collec
...
+      val authContext: Option[(LanguageParameter[ScalaLanguage], Term => Term)] = authImplementation match {
+        case Disable => None
+        case Native  => Some((LanguageParameter.fromParam(param"authContext: $authContextTypeName"), identity _))
+        case Custom =>
+          securityRequirements.map({ sr =>
+            val arg                           = LanguageParameter.fromParam(param"authContext: $authContextTypeName")
+            val securityRequirements          = renderCustomSecurityRequirements(sr)
+            val authContextParam              = param"${arg.paramName}"
+            val authTransformer: Term => Term = inner => q"""
+                authenticationMiddleware($securityRequirements, ${sr.optional}, req).flatMap { $authContextParam =>
```
blast-hardcheese, line 774: This is the last sticking point for me. `${sr.optional}` being passed in as a `Boolean` into `authenticationMiddleware` is really confusing, as a user. If authentication is optional, shouldn't there just be a `recover` on whatever `$authContextTypeName` is returned? `Some(Unauthorized)` => `None`?

blast-hardcheese: I took a stab at resolving this. There's something nagging at me about having both `SecurityExposure` and `AuthImplementation` passed in as separate parameters to a bunch of parameters. I've not changed any of the underlying logic, though reducing the number of `Boolean`'s exposed to users is better. I'll come back to this soon (I'd like to get this merged and on its way), but offering here in case anyone is interested in taking a look.

blast-hardcheese: Haven't found a great way to resolve this, so I'm going to merge as is.

blast-hardcheese:
The last concern I had was whether these parameters made sense across Scala/Java, and for all the different supported frameworks. I'm alright with proceeding.

blast-hardcheese (approved):

blast-hardcheese:
Thanks for your patience, @zeal18, life has been getting in the way of getting 100% behind where we ended up by the end here.

zeal18:
> The last concern I had was whether these parameters made sense across Scala/Java, and for all the different supported frameworks. I'm alright with proceeding.

I thought about it, especially about the `native`. I think it become clearer over the time.

I've took a look at some of your improvements but had no chance to check them all, it ok for me by the way) It's nice that it's merged, that means I'll check how the idea works in our project very soon 🤤

> Thanks for your patience, @zeal18, life has been getting in the way of getting 100% behind where we ended up by the end here.

thanks, it was nice to work on it!

PR #1368, opened 2022-01-05
Split out Java frameworks

blast-hardcheese:
Recently Java runtime tests have been just interrupted partway through,
so splitting up the frameworks in order to limit each individual runtime

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,9 +57,14 @@ jobs:
     needs: [core]
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


blast-hardcheese:
@sullis I removed some JDK versions in order to limit runtime, but do you have an alternate recommended set of representative JDKs to test?

blast-hardcheese:
@sullis Merging to gets CI to start passing. If there's a reason to re-add the other JDKs then that can be considered, but as of now I don't think it's necessary

Issue #738, opened 2020-07-24
vendor extension inside "items"


nickfun:
Hey guardrail is a super great project and I love it but I'm here to report a bug:

Generating clients for akka-http. Using the maven plugin 0.59.0.

In my spec file I choose a `x-jvm-type` nested inside some request paramters on a Post. The generated client ends up taking `String` instead of the type I specified.

Example snippet:

```yaml
  /v1/Services/{ServerlessServiceSid}/Builds:
    post:
      tags:
        - "Builds"
      x-jvm-package: "Serverless"
      summary: "Create a build of the service"
      operationId: "createBuild"
      parameters:
        - $ref: "#/components/parameters/ServerlessServiceSid"
      requestBody:
        required: true
        content:
          application/x-www-form-urlencoded:
            schema:
              required:
                - AssetVersions
                - FunctionVersions
                - Dependencies
              properties:
                AssetVersions:
                  type: array
                  items:
                    x-jvm-type: ServerlessFileVersionSid         # !
                    example: ZN00000000000000000000000000000000  # !
                    type: string                                 # !
                FunctionVersions:
                  type: array
                  items:
                    x-jvm-type: ServerlessFileVersionSid         # !
                    example: ZN00000000000000000000000000000000  # !
                    type: string                                 # !
                Dependencies:
                  type: string
      responses:
        201:
          description: "Created"
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/ServerlessBuildResource"
        400:
          $ref: '#/components/responses/BadRequest'
        404:
          $ref: '#/components/responses/NotFound'
        500:
          $ref: '#/components/responses/InternalServerError'
```

The generated client has this:

```
def createBuild(serverlessServiceSid: ServerlessServiceSid, assetVersions: Iterable[String], functionVersions: Iterable[String], dependencies: String, headers: List[HttpHeader] = Nil): EitherT[Future, ServiceClientError, CreateBuildResponse]
```

I expected the two `Iterable[String]` to be `Vector[ServerlessFileVersionSid]`

Have attempted a workaround of moving the `#!` lines above to a `$ref` but generated client didn't change.

kelnos:
Just some further info so I don't forget, because I got confused about this:

* This _does_ work for array properties in an object specified in `components/schemas` (or `definitions` for Swagger v2).
* This does _not_ work for array _parameters_ specified on a request.
* This does _not_ work for array properties in an object specified in-line in a `requestBody`.

[...]

Issue #190, opened 2019-02-25
Array<Long> is not supported in query parameters


plapinsc:
The following swagger definition:
```
    /WTF:
    get:
      operationId: "wtf"
      produces:
      - "application/json"
      parameters:
      - name: "id"
        in: "query"
        required: false
        type: "array"
        items:
          type: "integer"
          format: "int64"
        collectionFormat: "multi"
      responses:
        200:
          description: "successful operation"
```

generates a method with `Iterable[String]` input parameters instead of `Iterable[Long]`: 
`def wtf(respond: Resource.wtfResponse.type)(id: Option[Iterable[String]] = None): scala.concurrent.Future[Resource.wtfResponse]`

the expected result is:
`def wtf(respond: Resource.wtfResponse.type)(id: Option[Iterable[Long]] = None): scala.concurrent.Future[Resource.wtfResponse]`

To quickly mitigate the issue, I wanted to implement it with String and parse long by myself, but:
https://github.com/twilio/guardrail/issues/184

plapinsc:
Quick workaround: `x-scala-type: "Iterable[Long]"` seems to be working...

rtfpessoa:
@blast-hardcheese if I wanted to fix this where should I start looking?

blast-hardcheese:
@plapinsc Which version of guardrail was this happening on?

@rtfpessoa Depending on the answer, likely [`propMeta`](https://github.com/twilio/guardrail/blob/d6f2dbb74d37cd8374b0c9304bde8f4e4a6bc272/modules/codegen/src/main/scala/com/twilio/guardrail/SwaggerUtil.scala#L259-L270).

That being said, this got upended in 0.44.0 due to the switch to OpenAPI 3.x; to save yourself some effort, I'd check to make sure this is still broken in master (or with what's in #192, as that'll be reviewed and merged at some point this week).

The easiest way to test this would be to just modify `modules/sample/src/main/resources/server1.yaml` and run `runtimeSuite` in sbt, examining the generated result in `modules/sample-*`

plapinsc:
Our guardrail version is: `0.43.0`

rtfpessoa:
At least a handful of versions before and including `0.44.0` have this problem.
Will try to check master.

rtfpessoa:
@blast-hardcheese seems like this still happens in `0.45.1` will try to look at your suggestion when I get some time.

blast-hardcheese:
[...]

PR #1407, opened 2022-02-20
Replace strings with schema type resolvers

blast-hardcheese:
Continuing the work started in #1359, push the resolvers down and adding more structure (`ReifiedRawType` hierarchy for influencing generators)

- Pushing `Schema[_]` further down instead of just the `(rawType: Option[String], rawFormat: Option[String])` pair
- Using `Schema[_]` for `getDefault`
- Adding `(Schema[_], Components) => Schema[_]` for dereferencing `$ref`


blast-hardcheese:
The one caveat to this approach is that it (finally?) unifies the Scala server generators (akka-http and http4s) under `Vector[_]` instead of `Iterable[_]` for all route types.

This is a pretty significant breaking change, but hopefully a desirable one from the perspective of both consistency and speed, as well as by using a distinct collection instead of an interface, we now get typeclass instances from cats.

Due to the fairly idiosyncratic `respond` parameter, it should actually be possible to write a scalafix rule that takes care of at least some of these changes, but due to the fact that bugs like #190 have existed for some time, this won't be able to catch all breaking changes.

blast-hardcheese:
Also resolves #738

blast-hardcheese:
This is blocked by actually writing the scalafix rule to replace `Iterator[A]` with `Vector[A]` in route parameters, at the very least.

blast-hardcheese:
Alright, scalafix rule was submitted: https://github.com/scala-steward-org/scala-steward/pull/2612

PR #1412, opened 2022-02-27
Removing endpoints stubs

blast-hardcheese:
Endpoints has been stuck in somewhat of a limbo state for quite a while, due to both needing to overhaul the ScalaJS integration, as well as the increasing pain of maintaining source compatibility with endpoints4s for nontrivially complex specifications.

As a result, I move to finish dropping support for it. If there are some who continue to use it on an old version of guardrail and have been hoping for it to come back, please speak up here and we can come to some sort of way forward.


blast-hardcheese:
Merging this to unblock #1414

PR #1497, opened 2022-05-28
Replacing default language mapping with spi

blast-hardcheese:
- Replacing centrally-managed `GeneratorMappings` with SPI
- Deleted legacy `guardrail` project
- Updated `CONTRIBUTING.md`


blast-hardcheese:
I incorrectly removed the default language fallback here. I should have done a partition based on leading `--` instead of assuming the first would be a language

PR #1501, opened 2022-05-30
Improving module loader error messages

blast-hardcheese:
- Track individual module component extractors
- Accumulate consumed and missed modules
- Roll that stuff up into useful errors

This is a pretty big change over how the loaders previously worked, the ergonomics aren't great currently right now, but tracking in parallel (applicative-style) gives us this reporting capability.


blast-hardcheese:
Alright. This should resolve the usability regression brought on in #1414, where `--module scala-language` and `--module scala-stdlib` were required to be added despite being the only possible options.

modules/core/src/main/scala/dev/guardrail/generators/spi/FrameworkLoader.scala
```diff
@@ -1,34 +1,61 @@
 package dev.guardrail.generators.spi
 
+import cats.implicits._
+import cats.data.NonEmptyList
+import dev.guardrail.Target
```
blast-hardcheese, line 5:
```suggestion
```

PR #1503, opened 2022-06-05
StringSchema incorrectly treated as a model

### LINKED ISSUE #1502: could not find implicit value for parameter ev
Awesome library, thank you so much! I am encountering a couple of issues using the latest version.

addSbtPlugin("dev.guardrail" % "sbt-guardrail" % "0.71.0")

Generated client code for Akka HTTP does not compile for the following: 

```
openapi: 3.0.2
servers:
  - url: "http://localhost:1234"
paths:
  /foo:
    post:
      operationId: doFoo
      responses:
        '204':
          description: No response
      parameters:
        - name: bar
          in: query
          required: false
          schema:
            $ref: '#/components/schemas/Bar'
components:
  schemas:
    Bar:
      type: string
```

gives error 

```
Client.scala:45:84: could not find implicit value for parameter ev: com.example.test1.Implicits.AddArg[Option[io.circe.Json]]
[error]     makeRequest(HttpMethods.POST, host + basePath + "/foo" + "?" + Formatter.addArg("bar", bar), allHeaders, HttpEntity.Empty, HttpProtocols.`HTTP/1.1`).flatMap(req => EitherT(httpClient(req).flatMap(resp => resp.status match {
[error]                                                                                    ^
```

Similar for 

```
openapi: 3.0.2
servers:
  - url: "http://localhost:1234"
paths:
  /foo:
    post:
      operationId: doFoo
      responses:
        '204':
          description: No response
      parameters:
        - name: bar
          in: query
          required: false
          schema:
            $ref: '#/components/schemas/Bar'
components:
  schemas:
    Bar:
      type: object
      properties:
        baz: 
          type: string
```

gives error

```
Client.scala:46:84: could not find implicit value for parameter ev: com.example.test2.Implicits.AddArg[Option[com.example.test2.definitions.Bar]]
[error]     makeRequest(HttpMethods.POST, host + basePath + "/foo" + "?" + Formatter.addArg("bar", bar), allHeaders, HttpEntity.Empty, HttpProtocols.`HTTP/1.1`).flatMap(req => EitherT(httpClient(req).flatMap(resp => resp.status match {
[error]
```

blast-hardcheese:
Resolves #1502 

Following @djm1329's investigation, `StringSchema` was using `modelTypeAlias`, which wasn't designed to handle primitive type aliases. This may have been a hold-over from before the object model was standardized in `swagger-parser`, or perhaps it's just been a bug forever and #1407 exposed it.

This'll be a `bugfix`, not a `minor`, but due to `early-semver` seeming to be incompatible with `dependsOn`, the only way to get the tests to run successfully currently is to use `minor` to disable the `early-semver` check.

### DIFF
--- a/modules/core/src/main/scala/dev/guardrail/generators/ProtocolGenerator.scala
+++ b/modules/core/src/main/scala/dev/guardrail/generators/ProtocolGenerator.scala
@@ -736,7 +736,9 @@ object ProtocolGenerator {
                 defaultPropertyRequirement,
                 components
               )
-              alias <- modelTypeAlias(clsName, x, components)
+              customTypeName <- SwaggerUtil.customTypeName(x)
+              (declType, _)  <- SwaggerUtil.determineTypeName[L, F](x, Tracker.cloneHistory(x, customTypeName), components)
+              alias          <- typeAlias[L, F](formattedClsName, declType)
             } yield enum.orElse(model).getOrElse(alias)
           )
           .orRefine { case x: IntegerSchema => x }(x =>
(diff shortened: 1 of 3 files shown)


djm1329 (approved):
lgtm

PR #1539, opened 2022-07-15
circe-refined array size validation

stanislav-chetvertkov:


modules/scala-support/src/test/scala/tests/circe/ArrayValidationTest.scala
```diff
@@ -0,0 +1,88 @@
...
+          supportPackage = NonEmptyList.one("foop"),
+          defaultPropertyRequirement = PropertyRequirement.OptionalLegacy
+        )
+        .value
+
+      val expected =
+        q"""case class ValidatedCollections(boundedSizeArray: Option[Vector[Int] Refined
+           _root_.eu.timepit.refined.collection.Size[_root_.eu.timepit.refined.numeric.Interval.Closed[_root_.shapeless.Witness.`1`.T, _root_.shapeless.Witness.`10`.T]]] = None)"""
+
+      println(cls)
```
blast-hardcheese, line 82:
```suggestion
```

blast-hardcheese (approved):
Looks good, the reason the "patch" check is failing is because the coverage of the new code is low. If you can add the other two cases that should be sufficient

modules/scala-support/src/main/scala/dev/guardrail/generators/scala/circe/CirceRefinedProtocolGenerator.scala
```diff
@@ -28,6 +28,40 @@ object CirceRefinedProtocolGenerator {
   def applyValidations(className: String, tpe: Type, prop: Tracker[Schema[_]]): Target[Type] = {
     import scala.meta._
     tpe match {
+      case Type.Apply(t"Vector", inner) =>
+        val containerElementSchema: Tracker[Option[Schema[_]]] = prop.downField("items", _.getItems)
+        val innerType: Target[Type] = containerElementSchema.fold(
+          Target.fromOption(inner.headOption, RuntimeFailure(s"Unexpected state - empty 'items': ${containerElementSchema.showHistory}"))
+        ) { tracker =>
+          Target
```
stanislav-chetvertkov, line 36: there is too much nesting and the error propagation doesn't look right, how can one improve on this?

blast-hardcheese (approved):
This looks quite good, I wonder about the error messages though

modules/sample-akkaHttp/src/test/scala/generators/Circe/refined/ValidationTest.scala
```diff
@@ -1,14 +1,86 @@
...
+
+    "should fail if collection elements rejected by the predicate (string)" in {
+      ValidatedCollections2
+        .decodeValidatedCollections2(
+          parse("""{ "bounded_size_array" : ["pet", "pet", "carpet", "pet shop", "does not match"] }""").value.hcursor
+        )
+        .left
+        .value
+        .toString
+        .contains("""DecodingFailure(Predicate failed: "does not match"""") shouldBe true
```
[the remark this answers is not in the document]
stanislav-chetvertkov: It comes from `"io.circe"  %% "circe-refined"` - some error messages could be improved but imo it is on a good enough level as is

PR #1564, opened 2022-08-22
use extends to avoid duplicate config

jetersen:
Now the main release-drafter.yml does not need to managed by the script and you can simply modify it once.


.github/release-drafter-cli.yml
```diff
@@ -1,37 +1,7 @@
+_extends: guardrail:.github/release-drafter.yml
```
jetersen, line 1: if you want we can be specific that it should be `
guardrail-dev/guardrail:.github/release-drafter.yml`

blast-hardcheese:
I appreciate the help! I remember this being mentioned in the original discussion, but I never ended up revisiting the initial hacky implementation. This is much better!

jetersen:
@blast-hardcheese your welcome! :)

PR #1565, opened 2022-08-23
Remove regenerate release drafter script

blast-hardcheese:
Continuation from #1564


jetersen:
Oh, I though it was actually useful to keep it if you were to introduce a new module, but I guess that does not happen that often 😅

blast-hardcheese:
Yeah, plus I'm quite happy to maintain a handful of lines for a new file, the problem was really keeping them all in sync. Extending the config is a much more sensible solution.

PR #1595, opened 2022-09-25
Adding proper support for empty path segments

### LINKED ISSUE #1594: Empty path segments in URLs fail code generation
Splitting out of #1592,

> Also discovered another error(?):
> 
>     servers:
>         - url: /api/1-dev/
> 
> fails with
> 
>     Error:Failed to parse URL: endOfInput (unparsed: /datasets/) (.paths./datasets/)
>     [error] (Compile / guardrail) dev.guardrail.sbt.CodegenFailedException

The reason for this is that the parser doesn't know what to do with empty path segments, so `/api/1-dev/` combined with `/datasets/` ends up as `/api/1-dev//datasets/`.

blast-hardcheese:
Fixes #1594 

From the [whatwg spec](https://url.spec.whatwg.org/#urls), `https://example.org//` should be permitted.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -44,6 +44,23 @@ runExample := Def.inputTaskDyn {
   )
 }.evaluated
 
+lazy val runIssue: InputKey[Unit] = inputKey[Unit]("Run generators associated with a particular regression spec")
+runIssue := Def.inputTaskDyn {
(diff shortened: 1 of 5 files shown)


blast-hardcheese:
issues here because atto doesn't support full backtracking

Issue #1038, opened 2021-03-24
Add support for zio-http


blast-hardcheese:
Given the recent announcement of [zio-http](https://blog.dream11engineering.com/introducing-zio-http-2b66bb036fdc), adding [zio-http](https://github.com/dream11/zio-http) servers and clients to guardrail would remove the need for `zio-interop-cats` + http4s when generating for a zio codebase.

Peppi-Ressler:
Hi any chance to getting this done soon @blast-hardcheese ? :)

blast-hardcheese:
I've not started implementation at this time, but I'm happy to give guidance for those interested in pursuing this.

I opened this issue mainly as an indicator of intent after reading the blog posts with performance benchmark numbers, it was particularly amazing to see.

As I do not have any active ZIO projects to speak of after changing companies, I don't think it would be a service to the ecosystem if I were to not reflect popular idioms in whatever implementation I'd come up with.

If there's time and capacity, I'm happy to chat here, in [gitter](https://gitter.im/guardrail-dev/guardrail?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge), or via a Zoom chat or equivalent (timezones permitting).

Thanks for pinging me though, good to know there's active interest around this feature!

zeal18:
Hi guys! I've started working on this, but cannot promise it would be fast

blast-hardcheese:
> Hi guys! I've started working on this, but cannot promise it would be fast

Thank you for giving this a shot! Please let me know if there's a way I can assist, I'm also usually pretty responsive over Matrix (room linked from README)

blast-hardcheese:
> Would it make sense to extract it into a separate repository using the SPI approach to keep the core repository more focused?

Either way is fine for now, the module system is stable but there are no examples of doing this yet (especially when it comes to the test suites, unfortunately, I'll open an issue to fix that). Whatever helps make this easier!

Peppi-Ressler:
Hi any progress here? ZIO-http is gaining user base and this would be honestly quite huge for this project. @zeal18 has done quite some work there, how much remains? Would be possible to finish it by someone else or directly by @blast-hardcheese ? Thank you

blast-hardcheese:
@Peppi-Ressler The biggest barrier to entry is knowing what the end state for generated code would look like.

If someone can supply an example repo of a high quality zio-http application, including stuff like file uploads, json encoding and decoding, and service composition.

[...]

PR #1615, opened 2022-10-28
[WIP] zio-http support

zeal18:
So far I've defined the module, but it won't compile because it appeared that the zio-http is not published yet (local build requires building other dependencies locally as well and couldn't be verified in the CI). The plan is the following:
* wait until zio-http finalises its api and publishes at least the first version (during this period I can start experimenting locally)
* implement routes/endpoints/http-app generation using existing circe models generator
* add zio-json or/and zio-schema models generation support

Issue:  #1038

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -38,6 +38,11 @@ scala-http4s:
 - modules/scala-http4s/src/main/**/*
 - project/src/main/scala/modules/scalaHttp4s.scala
 
+scala-zio-http:
+- modules/scala-zio-http/*
+- modules/scala-zio-http/**/*
+- project/src/main/scala/modules/scalaZioHttp.scala
+
 scala-support:
 - modules/scala-support/src/main/*
 - modules/scala-support/src/main/**/*
--- a/.github/release-drafter-scala-zio-http.yml
+++ b/.github/release-drafter-scala-zio-http.yml
@@ -1,0 +1,7 @@
+_extends: guardrail:.github/release-drafter.yml
+name-template: 'scala-zio-http-v$RESOLVED_VERSION'
+tag-template: 'scala-zio-http-v$RESOLVED_VERSION'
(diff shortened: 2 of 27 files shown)


blast-hardcheese:
Understood about the dependency.

I've added the `minor` label to this PR which bypasses the infrastructure that favors stable-only module dependencies inside guardrail.

That should help the CI get further.

zeal18:
Hi @blast-hardcheese!
I found some time to continue working on this feature, but I'm struggling with the framework structure: it's too coupled. I think it's time to revise it.

The problem: the framework is too coupled and forces plugins to follow its ways. It defines generators interfaces and implements a common logic using them. The common logic prevents plugins to evolve independently, forces to them some common dependencies (for example cats and circe libraries) and complicates generating more "natural" for a particular plugin code.

Suggested solution: get rid of the common logic. Of cause the OpenAPI parsing logic should stay shared, but plugins should become as independent as possible, ideally, they could become even a separate applications

WDYT?

blast-hardcheese:
This is what I was attempting to do with the modules and SPI; I agree with the difficulties you're experiencing, the explosion of build tooling requirements is concerning though. I will try to respond with a more thorough consideration within the next week, I've been pulling a lot of long nights recently.

blast-hardcheese:
@zeal18 How do you want to proceed here, and do you still have interest in tending to this PR? Otherwise I can take a pass on getting the branch rebased, time permitting. Your choice.

zeal18:
> @zeal18 How do you want to proceed here, and do you still have interest in tending to this PR? Otherwise I can take a pass on getting the branch rebased, time permitting. Your choice.

Hi @blast-hardcheese! yes, I'm still considering continuing working on this, but I have a lack of time at the moment. I'll try to keep it moving starting from rebasing

blast-hardcheese:
@zeal18 if there's anything I can do to help please let me know.

PR #1686, opened 2023-02-05
Added support for HTTP HEAD method for akka-http and http4s

abrighton:
Also added minimal test cases to show that the "Unknown method" error no longer occurs.


blast-hardcheese:
I don't know why you're getting these test failures, they look unrelated. I'll take a look

blast-hardcheese (approved):
:ok_hand: Thank you!

PR #1785, opened 2023-07-30
Degeneralize server and client generators

zeal18:
The change degeneralizes the common logic for server/client generation making it possible to have different solutions for different frameworks and simplifying customizations


blast-hardcheese:
Pardon the delay, I'm road tripping through Italy for the next two weeks.

This will require review when I get back to a laptop, but I think you're right to do this on first glance.

Thank you for your effort here. I think the combinations of Config and which modules are enabled should be sufficient to maintain flexibility without the generalized methods, and this would give greater stability to the ABI.

The biggest downside is the inability to override a single method, which was the intent of the `copy` function on generators. That being said, any nontrivial use of that feature creates extraordinarily tight coupling, so maybe it's good to go.

Perhaps AST transformers in critical locations during generation would preserve customization somehow, but we can cross that bridge when we get there.

blast-hardcheese:
One of the features that I have struggled to maintain is the ability for the user to override a single method by using `.copy(...)` on a module and packaging that as their own module.

This could still be maintained by making all methods `protected` instead of `private`, as was recently proposed to resolve a usability issue in the http4s server generator, permitting forks of modules to override methods individually.

As much as I wish it were not necessary, I'm still sympathetic to this change, so with that change I think this can be merged. Hopefully it will increase maintainability enough that the reduction of symmetry between all generators will be worth it.

blast-hardcheese:
Release notes and migration guide to follow.

PR #1797, opened 2023-08-21
Make resource vals protected

gclaramunt:
Making encoders protected allows to overwrite them a provide a customized version (particularly to use streamJsonArrayEncoderOf instead of jsonEcnoderOf for Streams)
```
new SomeResource[F] {
 override  protected[this] val someOkEncoder = streamJsonArrayEncoderOf[F, fs2.Streamr[Entity]]

}
```

(not sure of the right collaboration protocol, let me know if something else is needed :) )


blast-hardcheese:
Ah, interesting. This looks straightforward, thank you!

I'll review and release most likely tomorrow.

blast-hardcheese (approved):

blast-hardcheese:
Thank you! If you modify your `project/guardrail.sbt` to include

```scala
libraryDependencies += "dev.guardrail" % "guardrail-scala-http4s" %% "0.76.1"
```
that'll pull in your changes. I'm moving towards opt-in modules to reduce the amount of code that needs to be downloaded, as well as to permit tools like scala-steward to automatically pull down bugfix versions of individual modules without having to run the whole release train for every small change.

PR #1854, opened 2023-10-25
OpenAPI v3.1.0 spec files do not generate valid data classes

ptrdom:
~~If spec would contain paths that use said schemas, then models would be generated.~~ https://github.com/guardrail-dev/guardrail/pull/1854#issuecomment-1780667775

### DIFF
--- a/modules/core/src/main/scala/dev/guardrail/core/extract/VendorExtension.scala
+++ b/modules/core/src/main/scala/dev/guardrail/core/extract/VendorExtension.scala
@@ -64,8 +64,8 @@ object VendorExtension {
     implicit val defaultVendorExtensibleMapProperty: VendorExtensible[MapSchema] =
       build[MapSchema](m => key => Option(m.getExtensions).map(_.get(key)))
 
-    implicit val defaultVendorExtensibleObjectProperty: VendorExtensible[ObjectSchema] =
-      build[ObjectSchema](m => key => Option(m.getExtensions).map(_.get(key)))
+    implicit val defaultVendorExtensibleObjectProperty: VendorExtensible[Schema[Object]] =
+      build[Schema[Object]](m => key => Option(m.getExtensions).map(_.get(key)))
 
     implicit val defaultVendorExtensibleStringProperty: VendorExtensible[StringSchema] =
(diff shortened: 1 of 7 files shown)


ptrdom:
Also, code works fine with OpenAPI `v3.0.0`. even though `paths` only became optional in `v3.1.0`.

blast-hardcheese:
So I poked around a bit and discovered that (hilariously?) [`JsonSchema extends Schema<Object>`](https://github.com/swagger-api/swagger-core/blob/74d826dee89e8af0a87e664cd4e5485141877047/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/JsonSchema.java#L9), but is functionally distinct from `ObjectSchema extends Schema<Object>`.

All the matching we do on `ObjectSchema` is invalid in the `JsonSchema` case, which is why you get `RandomType` instead of something coherent.

ptrdom:
Seems like this is not related to `schemas` being not references by `paths`, but a `v3.1.0` issue in general. Adjusted the PR title to reflect that.

blast-hardcheese:
OK. A little more coffee helped clear up what likely needs to happen in the interim, and some initial poking around to see if I could convince the parser to emit anything _other_ than `JsonSchema` suggests to me that we're dealing with some future-proofing.

Considering that, we can fall back to the supertype and just match on the various implementation branches and be done with things for now.

blast-hardcheese:
I'm offering this here as what I intend to merge, though the test still needs to be reflowed into `src/test/scala/core/issues/Issue1854.scala` with the original test restored.

ptrdom:
Nice! So is this solution okay to merge?

blast-hardcheese:
Thanks again for the report!

ptrdom:
Ah, too bad there are no snapshots for guardrail, hoping for a release soon!

blast-hardcheese:
Yeah, I've not had time to figure out how to get snapshots working with the release. If you know something about that, PRs are welcome! This will come in the 1.0 release coming soon.

ptrdom:
I have some thoughts about snapshot releases, will create an issue.

PR #1900, opened 2023-12-22
make code in sbt.md compile

mberndt123:


blast-hardcheese (approved):
Good catch, thank you!

blast-hardcheese:
I think `plugins/sbt.md` used to just be in the repo, but now that we're using docsify https://guardrail.dev/#/plugins/sbt.md is actually rendered. Rebuilding the website, but TODO to actually link those pages in with the rest of the website instead of just pushing the user off to the plugin README.md's

PR #1979, opened 2024-05-01
Zio http poc

stanislav-chetvertkov:


sideeffffect:
Implementation in zio-http itself https://github.com/zio/zio-http/pull/2562
Could be helpful, so I'm linking it here.

Issue #2050, opened 2024-11-10
Code generation fails when components are named identically to one of their enum properties


Jonnty:
I've been getting errors when generating code for a YAML OpenAPI 3 spec ([this one](https://developer.data.nationalhighways.co.uk/api-details#api=road-and-lane-closures&operation=RoadClosures), incidentally). They appear to be happen when one of the enum properties of a component is the same as the component itself. The cause seems to be that attempts to reference the component class are taken to be references to the identically named property object within the class when such an object happens to exist. (This doesn't happen for standard, non-enum strings as no child object is generated.)

The conflict is not case-sensitive, as generated class/object names are automatically camel cased.

I think the suggested fix for https://github.com/guardrail-dev/guardrail/issues/1147 - fully qualifying component class names within generated code - would probably fix this as well.

I've narrowed this down into a minimal failing case - I get the following error when attempting to generate code for the specification below it by running `sbt compile`  on a project using sbt-guardrail 1.0.0-M1.

**Error**
```
[info] compiling 1 Scala source to C:\...\untitled\target\scala-2.13\classes ...
[error] C:\...\untitled\target\scala-2.13\src_managed\main\nhclosures\definitions\Test.scala:13:129: value test is not a member of nhclosures.definitions.Test.Test
[error]     _root_.io.circe.Encoder.AsObject.instance[Test](a => _root_.io.circe.JsonObject.fromIterable(_root_.scala.Vector(("test", a.test.asJson))))
[error]                                                                                                                                 ^
[error] C:\...\untitled\target\scala-2.13\src_managed\main\nhclosures\definitions\Test.scala:15:219: type Test is not a member of object nhclosures.definitions.Test.Test
[error]   implicit val decodeTest: _root_.io.circe.Decoder[Test] = new _root_.io.circe.Decoder[Test] { final def apply(c: _root_.io.circe.HCursor): _root_.io.circe.Decoder.Result[Test] = for (v0 <- c.downField("test").as[Test.Test]) yield Test(v0) }
[error]                                                                                                                                                                                                                           ^
[error] C:\...\untitled\target\scala-2.13\src_managed\main\nhclosures\definitions\Test.scala:15:236: nhclosures.definitions.Test.Test.type does not take parameters
[...]

PR #2051, opened 2024-11-11
Circe - fully qualify references within component classes

Jonnty:
Previously, when a property was named the same as its component, references to the component class and property class conflicted, causing errors. Fully qualifying the references to each should prevent this from happening.

A regression test covering a minimal failing case is included.

Fixes issue #2050


Jonnty:
Just realised I didn't include the updated TypesTest expectations - added that now.

